PR #1069, opened 2013-08-30
fixes #1068 pagination included when not neeeded

### LINKED ISSUE #1068: Pagination shown when no need in notmyidea example theme
I started a theme by looking how notmyidea example was made. I had an issue when the number of article was lower than the DEFAULT_PAGINATION setting, the pagination.html was included when it should not. Showing 1/1 is not helpful.

Example: DEFAULT_PAGINATION set to 5 and a blog with only 4 articles.

JSteunou:
It fixes the bug in my case, but maybe there are some edge case I'm not aware off.

### DIFF
--- a/pelican/themes/notmyidea/templates/index.html
+++ b/pelican/themes/notmyidea/templates/index.html
@@ -43,8 +43,9 @@
         {% endif %}
         {% if loop.last %}
             </ol><!-- /#posts-list -->
-            {% if loop.last and (articles_page.has_previous()
-            or not articles_page.has_previous() and loop.length > 1) %}
+            {% if loop.last and
+                ( (loop.length == DEFAULT_PAGINATION  and articles_page.has_next()) or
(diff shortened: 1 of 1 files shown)


ehashman:
Confirmed that this is still an issue on pelican 3.6.0. I can look into taking over this PR and updating the tests if that would be helpful?

justinmayer:
@ehashman: Interesting timing... I ran into this same problem as well earlier this week. Thank you for offering to take over and update the tests, Elana. That would indeed be very helpful!  (^_^)

justinmayer:
Hi @ehashman. Would you be willing to help with this and update the tests? (^_^)

ehashman:
@justinmayer: just took a look and I already fixed this (more extensively and with tests) in #1770. So I think this can be closed.

leotrs:
@justinmayer this can be closed as per @ehashman's last comment.

justinmayer:
Many thanks, Leo.

PR #1138, opened 2013-10-26
the beginning of zh_CN docs & a bug fix in en docs

linmx0130:
It's the beginning of zh_CN docs.
I am a student from China. I will have a busy time next 10 days and restart the translating work on Nov. 11th. Could you please accept my translation?


leotrs:
What is the state of this? It would be nice to have a translation. If no one is working on this any more, we should probably close the PR.

justinmayer:
If anyone would like to pick up and update this Chinese translation endeavor, we would certainly welcome your contributions.

PR #1149, opened 2013-11-06
Fixes #1143, spaces in filename links

### LINKED ISSUE #1143: {filename} does not work properly with files containing spaces
It appears when using {filename} spaces are removed resulting in an error.

For instance <{filename}docs/Space In Name.pdf> results in the following error:

```
WARNING: Unable to find docs/SpaceInName.pdf, skipping url replacement
```

FloorLamp:
If %20 in path and path doesn't exist, replaces %20 with spaces.

### DIFF
--- a/pelican/contents.py
+++ b/pelican/contents.py
@@ -213,6 +213,10 @@ class Content(object):
                         os.path.join(self.relative_dir, path)
                     )
 
+                # spaces in filename
+                if path not in self._context['filenames'] and '%20' in path:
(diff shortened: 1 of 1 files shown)


tooh:
i'm runnng v4.1.1 and experience still the problem of spaces in filenames.
Is this fix released ?

justinmayer:
Given that the fix in question was merged six years ago, presumably it is indeed in 4.1.1.

Perhaps review the “How to Get Help” section of the docs and follow the steps listed there?

PR #1175, opened 2013-12-01
Prioritize the latest version of the doc for search engines.

2xyo:
Make search engines only see the latest version of the documentation.

This code is picked from the [FAQ of readthedocs.org](https://read-the-docs.readthedocs.org/en/latest/faq.html#can-i-make-search-engines-only-see-one-version-of-my-docs).


justinmayer:
I've just changed the [relevant setting](https://readthedocs.org/dashboard/pelican/versions/) so that requests for http://docs.getpelican.com/ will re-direct to http://docs.getpelican.com/en/stable/, so I think we can finally put this issue to rest.

PR #1197, opened 2013-12-25
Add custom configuration setting for summary ending text

kalefranz:
Pelican currently forces auto-created summaries to end with the text `_...` where `_` here is a space.  This pull request makes that the default (so nothing breaks), but adds an optional configuration parameter.  An example use would be `...` instead of `_...`.


justinmayer:
Would appreciate some help from @getpelican/reviewers. Would you please have a look at this pull request and submit a review?

pelican/utils.py
```diff
@@ -360,7 +360,7 @@ def path_to_url(path):
         return '/'.join(split_all(path))
 
 
-def truncate_html_words(s, num, end_text='...'):
+def truncate_html_words(s, num, end_text=' ...'):
```
leotrs: Shouldn't the default value of `end_text` be `self.settings['SUMMARY_END_TEXT']`? I personally don't like hardcoding the default value again if we already have a `self.settings` for it.

galaxy4public: I would argue that we should not bind an utility function to a specific use-case or setting and that having the hardcoded default in the utility function is fine.  I am currently working on a PR that implements this functionality, please see #2710 .

galaxy4public:
I submitted a PR that introduces the same functionality, includes tests, and does not conflict with the current `HEAD`.  Despite that I was working on my version independently (did not know about this PR until I got mine working), the changes are almost exactly the same.  Except for the configuration variable name.

Since both PRs are addressing the same issue, yet mine is not in conflict with the source tree, I would suggest to close this PR and move further the discussion to #2711 .

justinmayer:
This functionality is now present in master via the aforementioned new PR.

PR #1204, opened 2013-12-27
Ignore empty files silently. Fixes #1203.

### LINKED ISSUE #1203: Empty files throw exceptions / warning
I'm using 'touch slug' in my content directory to take notes on what I still want to write about - these show (unnecessary) warnings which clutter other output.

berlincount:


justinmayer:
I think @kylef is right. If someone would like to help implement the approach he suggested, by all means please do so.

PR #1207, opened 2013-12-28
Add support for unicode slugs

martinpaljak:
I want to use Pelican in a fully localized environment, where even the slugs would make use of unicode characters like õäöü.

This patch helps (there is an additional option to slugify, which defaults to OFF at the moment (as there is no direct access to settings?).

I felt that this approach is better than using the current configurable substitution option.

### DIFF
--- a/pelican/utils.py
+++ b/pelican/utils.py
@@ -227,7 +227,7 @@ def pelican_open(filename):
     yield content
 
 
-def slugify(value, substitutions=()):
+def slugify(value, substitutions=(), use_unicode=True):
     """
     Normalizes string, converts to lowercase, removes non-alpha characters,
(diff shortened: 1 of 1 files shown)


justinmayer:
Hi Martin. Very sorry for the long delay in reviewing this pull request. It would appear that the Travis CI build failed. Would you be willing to rebase on current master, ensure all tests pass, and push a squashed commit to this branch so we can review it?

justinmayer:
@avaris: Could you take a look and offer your take on whether there is sufficient value here for inclusion in Pelican?

avaris:
We could provide a setting for not ASCII-fying the slugs, so it's probably worth having. But this needs docs, setting and tests. I can work on it.

Issue #1198, opened 2013-12-25
Links to localized article URL-s with month names are in wrong language


martinpaljak:
On disk, everything is correct:
(conf)
DEFAULT_LANG = 'et'
LOCALE = "et_EE.UTF-8"

DATE_FORMATS = {
    'et': ('et_EE.UTF-8','%A, %d %B %Y'),
}

MONTH_ARCHIVE_URL = '{date:%Y}/{date:%B}/'
MONTH_ARCHIVE_SAVE_AS = '{date:%Y}/{date:%B}/index.html'

and I get on disk:
veeb/2013/detsember/index.html

Yet the generated HTML (not my idea theme) includes:

a href="./2013/December/teine-post/" rel="bookmark"

Have I missed something?

PR #1208, opened 2013-12-28
Fix for getpelican/pelican#1198

martinpaljak:
When rendering templates, the locale was set to C in getpelican/pelican@ddcccfeaa952d2e1e24ceac94e5d66c73b57c01b

If one used a locale that made use of unicode characters (like et_EE.UTF-8)
the files on disk would be in correct locale while links would be to C.

Works with Python3, gives ascii decoding errors with Python2


justinmayer:
I believe the referenced issue has been resolved. If you find that not to be the case, please comment here and I will re-open.

PR #1233, opened 2014-01-14
Adds support for themes functions

magnunleno:
With this changes the themes can have (not obligatory) a functions.py
file bundled with the templates. The purpose of this file is to hold
helper functions to be injected into the jinja2 templates. To control
which functions will be injected is specified the **export** list. A
sample was inserted in pelican/themes/simple/templates/functions.py.


lgiordani:
This could be also helpful to initialize theme variables. For example a theme that supports content downloaded from a CDN may set the content version in a Python file and use the variables in its templates.

ingwinlu:
themes already can set default values. so no profiting there

lgiordani:
How? Sorry but I was not aware of that feature, otherwise I wouldn't have proposed it.

justinmayer:
@ingwinlu: Care to elaborate on your comment above?

Issue #980, opened 2013-07-29
Typogrify to ignore MathJAX Latex code


sahands:
I'd like to use Typogrify but it does not work well together with MathJAX. I use $'s to encompass MathJAX Latex code. Is there any way to get either typogrify to ignore such pieces of code, or to stop Pelican from feeding such parts to Typogrify to begin with? I am using reST to generate the content. 

Thank you.

PR #1250, opened 2014-02-01
Added TYPOGRIFY_FILTERS as discussed in #980

sahands:
TYPOGRIFY_FILTERS allows users to specify which typogrify filters to are be applied. This was discussed as a solution to #980, to prevent typogrify from corrupting MathJax output.


jonathan-s:
@justinmayer This pull request can be closed as it is a duplicate of #1270 which is slightly more updated.

justinmayer:
Thanks, Jonathan.

PR #1253, opened 2014-02-06
Add support for ISSO (http://posativ.org/isso/) comment system.

oerdnj:
Hi,

this commit adds support for self-hosted ISSO comment system to have an alternative to the Disqus. The code is quite simple (it adds ISSO_SITEURL config variable) and it already runs on deb.sury.org.

Thanks for considering this pull request.

O.


silverhook:
Any news on this?

justinmayer:
I don't believe this belongs in core, and as far as I am aware, nobody has stepped forward to implement ISSO support via a plugin. I suggest using: https://github.com/getpelican/pelican-plugins/tree/master/pelican_comment_system.

PR #1270, opened 2014-02-21
Typogrify filters setting added. Setting TYPOGRIFY_FILTERS.

sahands:
Added TYPOGRIFY_SETTINGS setting to partially help with #980.


justinmayer:
I believe the issue that prompted this pull request has been resolved. If that is not the case, please comment here so this pull request can be resurrected.

PR #1282, opened 2014-03-09
Workaround for s3cmd mimetype detection

FedericoCeratto:
Mime-type detection is not always reliable
https://github.com/s3tools/s3cmd/issues/198


sevagh:
I'm actually still experiencing this issue. Not sure how much the scene has changed since this ticket was open (it's been more than 2 years...). I need to run s3cmd from a Python 2 venv, and Pelican from Python 3. I do "make html" with Pelican, and "s3cmd sync output/ --acl-public --no-mime-magic --guess-mime-type s3://sevag.io/" with s3cmd, and I'm getting incorrect MIME types in my S3 bucket:

The stylesheet http://sevag.io/theme/css/bootstrap.min.css was not loaded because its MIME type, "text/plain", is not "text/css".

Is there a better way to upload to S3 nowadays?

sevagh:
I used awscli and everything is OK now:

    aws s3 sync output/ s3://sevag.io

PR #1327, opened 2014-04-24
Add Multi-theme support

aftersomemath:
Adds multi-theme support using the new THEMES setting.
You can specify all the themes that you will be using in python
dictionary form. The keys to the themes can then be used by other themes
to inherit.

This is still a work in progress, I have a couple things left to do including testing everything completely, a minor thing to prevent default settings from getting overwritten, and adding documentation. 

I am just making a PR so that I can close the old one #1297. So much has changed that I feel the old PR needed to be closed completely.

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -653,6 +653,12 @@ Setting name (default value)                        What does it do?
                                                     the paths defined in this settings, they will be
                                                     progressively overwritten.
 `CSS_FILE` (``'main.css'``)                         Specify the CSS file you want to load.
+`THEMES`                                            Extra themes that can be inherited from. They can also
+                                                    inherit from each other. Use a dictionary to make a list
+                                                    of all the available list. The key in the dictionary will
+                                                    also be the prefix you use to inherit from the theme.
+
+                                                    Example: ``THEMES = {'!foo': foo, '!foobar':foobar}``
 ================================================    =====================================================
 
 
--- a/docs/themes.rst
+++ b/docs/themes.rst
@@ -298,15 +298,22 @@ Here is a complete list of the feed variables::
 Inheritance
 ===========
 
-Since version 3.0, Pelican supports inheritance from the ``simple`` theme, so
-you can re-use the ``simple`` theme templates in your own themes.
+Since version 3.4, Pelican supports inheritance from the ``simple`` theme, as well
+as any themes specified in the ``THEMES`` setting. You can re-use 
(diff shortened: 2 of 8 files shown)


honzajavorek:
I wish to have this feature. Let me know if there's something I can help with.

mitchtbaum:
iirc, this patch needs refactoring to read ~`PARENT_THEME` from a metadata file within each theme, instead of in the templates themselves (Sphinx Docs and Nikola work this way).  Then, Pelican could assemble its theme combo by reading each template from each theme in the `themes` dir and navigating within the theme directory to resolve missing templates referenced in each theme, based on its parent (similar to how simple theme currently works, only with multiple levels).

honzajavorek:
@mitchtbaum Thanks for explanations and encouraging. I go skiing for several days, but then I'd like to solve this my issue https://github.com/honzajavorek/danube-delta/issues/4, so I'll really try to get to this.

Elsewhere you mentioned @smartass101's work in the [add_multi_theme_support](https://github.com/getpelican/pelican/commits/add_multi_theme_support) branch. With the one in this PR it's already two unfinished branches. What is the relation between them? The PR was first, was not finished, @smartass101 started to work on it in his separate branch, but also wasn't able to finish it?

I could try to resurrect this. When I look at all the comments and the code which was already written, I'll might have more questions, but now it's up to me to make the first move I guess. Thanks again for support.

justinmayer:
Hi @honzajavorek. Any updates regarding your takeover of the work in this pull request?

honzajavorek:
@justinmayer Unfortunately no, I temporarily overcame the limitation by having `site.css` with some specific CSS and colors linked after the main theme style. The file lives alongside content of each blog and overrides styles by CSS cascade. Example: https://github.com/pyvec/blog.python.cz/ It has two downsides:

1. Changes in `site.css` are not being picked up by reloader, so when overriding my default style, I have to manually restart the pelican server and that's annoying.
2. It just doesn't feel right to have CSS files in the same folder as content.

However, having this half-working solution brought down my motivation to work on built-in support and I ended up dedicating time to different projects, such as PyCon CZ.

boxydog:
Since this PR was created, someone merged a different inheritance scheme (THEME_TEMPLATE_OVERRIDES) in #2072, described in #2021.

I don't understand THEME_TEMPLATE_OVERRIDES and THEMES well enough to know how they should co-exist.

PR #1342, opened 2014-05-10
pelican should also parse the summary for {filename}

smartass101:
Pelican does not parse the summary of an article for replacing the local links with the {filename} syntax.

if the article start with :

```rst

Title
-----

:summary: I want to link a previous_ article

.. _article: {filename}my_previous_post.rst

Lorem ipsum dolor…

```

Then the interlink will not be processed and interpreted. A broken link will be inserted if I try to display this summary on the articles page.

Correction : in contents.py call the _update_content method when loading the summary :

```python
    def _get_summary(self):
        """Returns the summary of an article.

        This is based on the summary metadata if set, otherwise truncate the
        content.
        """
        if hasattr(self, '_summary'):
            return self._update_content(self._summary, self._context.get('localsiteurl', ''))
            …

```

### DIFF
--- a/pelican/contents.py
+++ b/pelican/contents.py
@@ -278,7 +278,8 @@ class Content(object):
         content.
         """
         if hasattr(self, '_summary'):
-            return self._summary
+            return self._update_content(self._summary,
+                                        self._context.get('localsiteurl', ''))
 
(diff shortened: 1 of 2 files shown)


ingwinlu:
since summary is properly memoized now this can be rebased and merged

PR #1352, opened 2014-05-18
Add regular expressions support into STATIC_PATHS.

gw0:
Documentation is describing that it is possible that other posts and images may be sitting alongside the current post, but for images and other non-article or non-page content this is not entirely true. By specifying `STATIC_PATHS` it is only possible to specify whole paths that will get copied, not files stored alongside any article or page.

Therefore I quickly implemented regular expressions support for STATIC_PATHS. As can be seen from the code previous behavior remains the same (if specified whole path or file exists it is used), but in case you specify a regexp or something that doesn't exists, the whole directory tree is traversed searching for matching paths.

Example usage:

```
content/article1/index.md  -- with: ![pic1]({filename}img/pic1.jpg)
content/article1/img/pic1.jpg
content/article1/img/pic2.jpg

STATIC_PATHS = [
    '.*/img$',
]
```

Although my initial opinion is that directory/file inclusion and exclusion related code needs some refactoring.

### DIFF
--- a/pelican/generators.py
+++ b/pelican/generators.py
@@ -9,6 +9,7 @@ import logging
 import shutil
 import fnmatch
 import calendar
+import re
 
 from codecs import open
 from collections import defaultdict
@@ -126,6 +127,7 @@ class Generator(object):
             root = os.path.join(self.path, path)
 
             if os.path.isdir(root):
+                # walk specified directory
                 for dirpath, dirs, temp_files in os.walk(root, followlinks=True):
                     for e in exclude:
                         if e in dirs:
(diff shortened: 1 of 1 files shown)


gw0:
Just to let everyone know: Attach is pretty good, but I ended up switching completely to tags and abusing categories to create the folder structure I want. By manually controling the folder structure issues such as attaching same images from different articles makes sense again.

egberts:
So, we didn't get this feature-static-path into merged into the Pelican master?

justinmayer:
@egberts: If someone wants to (1) describe multiple use cases where such a feature would be useful, (2) implement an appropriate solution (e.g., built on top of this PR or otherwise), and (3) add documentation and tests, then we would certainly consider merging such a contribution.

PR #1363, opened 2014-05-30
improve fabfile so we can do fab <command> in any sub-folder.

SkyLothar:
improve fabfile so we can do fab <command> in any sub-folder.
fix import order and code style to meet pep8

use alias=preview instead of define another function


justinmayer:
Given that Pelican now includes a template for Invoke tasks — and no longer includes a template for Fabric — I think it's time to retire this pull request. If anyone wants to adapt it for Invoke, please feel free to do so in the context of a new pull request.

PR #1126, opened 2013-10-17
Fix {category} and {tag} placeholders when called from subdirs

agateau:
{category} placeholder was failing for me when called from a page in a subdir. This change fixes it by applying the to {category} and {tag} the same operations applied to {filename}.

PR #1377, opened 2014-06-15
Prepend siteurl (or relative_url) to {tag} and {category} links

danielrichman:
I _think_, though am not sure, that the [unit tests](https://github.com/danielrichman/pelican/compare/cateogry_tag_siteurls?expand=1#diff-a6e75eddaf0b1d483ecdfcac11dad813L215) for `{tag}`/`|tag|`/etc. links are testing incorrect behaviour.

This patch puts site-url in front of intra-site links to tags or categories.

### DIFF
--- a/pelican/contents.py
+++ b/pelican/contents.py
@@ -244,9 +244,10 @@ class Content(object):
                                     "Other resources were not found"
                                     " and their urls not replaced"))
             elif what == 'category':
-                origin = Category(path, self.settings).url
+                origin = '/'.join((siteurl,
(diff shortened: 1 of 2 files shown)


justinmayer:
Related to #1126?

justinmayer:
Many thanks for your contribution, Daniel, and apologies for the long delay in reviewing. I don't think the relevant code exists any longer, so for now I think it's best to close this PR. If you think related follow-up work is needed, please submit an issue so we can discuss.

PR #1384, opened 2014-06-28
allows pelican-themes symlink command to use relative paths not beginnin...

jacoblyles:
...g with a dot

pelican-themes -s <dir-name> creates a broken symlink when the <dir-name> is a relative path - i.e. vendor/themes/clear-theme. Afterwards the theme does not show up with pelican-themes -l. Notice that the -i or install command does allow for relative paths.

This change makes sure that the symlink command makes a correct symlink. A test file has also been added for pelican-themes.

### DIFF
--- a/pelican/tests/TestTheme/theme/a.html
+++ b/pelican/tests/TestTheme/theme/a.html
@@ -1,0 +1,1 @@
+this is a test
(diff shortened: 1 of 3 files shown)


leotrs:
@jacoblyles is the command still broken? I have had problems with `-s` before too.

mUtterberg:
@kylef @justinmayer has this been reviewed in the past few years? I'm having an issue with symlink & this seems like a possible fix.

PR #1401, opened 2014-07-13
Support last page pattern in PAGINATION_PATTERNS

arty-name:
I want to be able to specify the URL pattern for the last page using `-1` as identifier like this: `(-1, '{base_name}/last/', '{base_name}/last/index.html')`. Currently, only special handling of the first page is possible.

Here’s the background for this feature request.

I want the page numbers to go in the different direction, so that page 1 would always contain the same set of articles: the oldest ones. The same applies to other pages as well: their content will not change over time. The original version of the PR was calling it **fixed pagination** or **static pagination**. Among other things this is good for SEO. I implemented that in my current weblog: https://blog.arty.name/. The current default on pelican and most websites on the internet is the opposite: the items on a page could be pushed to another page over time, and it is unfortunate.

Having `ARTICLE_ORDER_BY = 'date'` and `NEWEST_FIRST_ARCHIVES = True` I could use `object_list[::-1]` in `index.html` and `tag.html` of my theme to almost reach my goal. With this PR merged I can do the following: 

```python
PAGINATION_PATTERNS = (
    (1, '{base_name}/page-1/', '{base_name}/page-1/index.html'),
    (2, '{base_name}/page-{number}/', '{base_name}/page-{number}/index.html'),
    (-1, '{base_name}/', '{base_name}/index.html'),
)
```

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -1067,6 +1067,11 @@ as follows::
   )
 
 
+If you want a pattern to apply to the last page in the list, use ``-1``
+as the ``minimum_page`` value::
+
+    (-1, '{base_name}/last/', '{base_name}/last/index.html'),
+
 Translations
 ============
 
--- a/pelican/paginator.py
+++ b/pelican/paginator.py
@@ -118,8 +118,13 @@ class Page:
 
         # find the last matching pagination rule
         for p in self.settings['PAGINATION_PATTERNS']:
-            if p.min_page <= self.number:
-                rule = p
+            if p.min_page == -1:
+                if not self.has_next():
+                    rule = p
+                    break
+            else:
+                if p.min_page <= self.number:
+                    rule = p
 
         if not rule:
             return ''
--- a/pelican/tests/test_paginator.py
+++ b/pelican/tests/test_paginator.py
@@ -76,3 +76,29 @@ class TestPage(unittest.TestCase):
         page2 = paginator.page(2)
         self.assertEqual(page2.save_as, 'blog/2/index.html')
         self.assertEqual(page2.url, '//blog.my.site/2/')
+
+    def test_custom_pagination_pattern_last_page(self):
+        from pelican.paginator import PaginationRule
+        settings = get_settings()
+        settings['PAGINATION_PATTERNS'] = [PaginationRule(*r) for r in [
+            (1, '/{url}1/', '{base_name}/1/index.html'),
+            (2, '/{url}{number}/', '{base_name}/{number}/index.html'),
+            (-1, '/{url}', '{base_name}/index.html'),
+        ]]
+
+        self.page_kwargs['metadata']['author'] = Author('Blogger', settings)
+        object_list = [Article(**self.page_kwargs),
+                       Article(**self.page_kwargs),
+                       Article(**self.page_kwargs)]
+        paginator = Paginator('blog/index.html', '//blog.my.site/',
+                              object_list, settings, 1)
+        # The URL *has to* stay absolute (with // in the front), so verify that
+        page1 = paginator.page(1)
+        self.assertEqual(page1.save_as, 'blog/1/index.html')
+        self.assertEqual(page1.url, '//blog.my.site/1/')
+        page2 = paginator.page(2)
+        self.assertEqual(page2.save_as, 'blog/2/index.html')
+        self.assertEqual(page2.url, '//blog.my.site/2/')
+        page3 = paginator.page(3)
+        self.assertEqual(page3.save_as, 'blog/index.html')
+        self.assertEqual(page3.url, '//blog.my.site/')


arty-name:
Wow that thing is old

justinmayer:
Indeed. Very sorry for neglecting your pull request, Artemy. One of my goals is to handle this better in the future.

With that said, are these changes still relevant? Please let me know.

arty-name:
Yes, they are still relevant, and I am still using that branch for my blog. 

Looking at the big picture, I think this approach to pagination is good for the Internet overall, and should be even considered for a default. But obviously it is too much to ask, so I will be happy just to see that small patch included.

justinmayer:
@avaris / @ingwinlu / @iKevinY: Do you have a moment to review this pull request?

docs/settings.rst
```diff
@@ -599,6 +599,8 @@ Setting name (followed by default value, if any)    What does it do?
                                                     pagination.
 ``PAGINATION_PATTERNS``                             A set of patterns that are used to determine advanced
                                                     pagination output.
+``PAGINATION_FIXED = False``                        Set to true to have pages with fixed content,
```
iKevinY: In my opinion, `FIXED_PAGINATION` reads better than `PAGINATION_FIXED`. That being said, maybe a better way to describe this setting is `REVERSED_PAGINATION`? I'm not sure though.

pelican/paginator.py
```diff
@@ -28,20 +28,30 @@ def __init__(self, name, object_list, settings):
         if settings.get('DEFAULT_PAGINATION'):
             self.per_page = settings.get('DEFAULT_PAGINATION')
             self.orphans = settings.get('DEFAULT_ORPHANS')
+            self.pagination_fixed = settings.get('PAGINATION_FIXED')
```
iKevinY: This attribute name be updated appropriately if the setting name gets changed to `FIXED_PAGINATION`.

arty-name:
Thank you, @iKevinY! 

I’d argue that "fixed" tries to describe the point of having this kind of pagination, while "reversed" merely describes the means. "Fixed" is not doing a very good job at that, however I cannot come up with a better alternative. I guess it would be easier for a native English speaker. 

I will try to write tests as well.

Scheirle:
IMHO `static` and `dynamic` pagination sound right, where `dynamic` is the current default behavior.
`dynamic` -> articles move back in the stack to make room for newer ones.
`static` -> articles do not change their position.

---

Instead of a Boolean setting (`PAGINATION_FIXED`) you could also use string setting (`PAGINATION_MODE`), which would either be `static` or `dynamic` (default).

arty-name:
Okay, I have updated my commits with tests. Tests for Paginator class were effectively non-existent, so I tried to come up with something by myself. 

Regarding proper names, I do like the proposal by @Scheirle to use Enum for PAGINATION_MODE being static or dynamic, however I do not see any other enums in Setting, so I am not sure if it is okay to have them. @justinmayer, @iKevinY: do you agree?

arty-name:
Hey there! Believe it or not I have been using this branch of Pelican all these years. Now I took a fresh look at Pelican and noticed `ARTICLE_ORDER_BY`, which suits my purpose nicely. Having `ARTICLE_ORDER_BY = 'date'` I could use `object_list[::-1]` in `index.html` and `tag.html` of my theme to almost reach my goal. 

The only part missing would be [this small change](https://github.com/arty-name/pelican/commit/0dee597e57cf36af4937429b660929b7fb6893b2#diff-76400fd673719d4cb6ae2e7e53346ff922a1000ea722b809aaa43bee16f00871R122-R128) to allow [that usage of the PAGINATION_PATTERNS](https://github.com/arty-name/pelican/commit/0dee597e57cf36af4937429b660929b7fb6893b2#diff-1d78ba60e1efa4b7b08982e1bbaeceffefea079d402b7d58e57c0cfc7e6c78d8R630-R633).

What do you think, should I rather create a new PR in favor of this minimal change or update the current PR?

avaris:
@arty-name reusing this PR would be OK. Just to make sure I understand correctly, your changes would be adding a way (`-1`) to `PAGINATION_PATTERS` to reference the last page, right?

arty-name:
This way is already included in the current PR, I can just undo the rest of the changes since they are not strictly required anymore.

arty-name:
Okay, now this PR is much smaller: `+12 −2`. Maybe this would help it to get through the review!

avaris:
How about some tests? :)

arty-name:
I’ve added a test

pelican/tests/test_paginator.py
```diff
@@ -76,3 +76,29 @@ def test_custom_pagination_pattern(self):
...
+        paginator = Paginator('blog/index.html', '//blog.my.site/',
+                              object_list, settings, 1)
+        # The URL *has to* stay absolute (with // in the front), so verify that
+        page1 = paginator.page(1)
+        self.assertEqual(page1.save_as, 'blog/1/index.html')
+        self.assertEqual(page1.url, '//blog.my.site/1/')
+        page2 = paginator.page(2)
+        self.assertEqual(page2.save_as, 'blog/2/index.html')
+        self.assertEqual(page2.url, '//blog.my.site/2/')
+        page2 = paginator.page(3)
```
iKevinY, line 102: I think this variable should be `page3`. 😛

arty-name: Fixed!

docs/settings.rst
```diff
@@ -1067,6 +1067,11 @@ as follows::
   )
 
 
+If you want a pattern to apply to the last page in the list, use ``-1``
+as ``minimum page`` value like this::
```
iKevinY, line 1071: nit: Could this be worded as _"use `-1` as the `minimum_page` value:"_ instead?

iKevinY (approved):
This PR looks good to me; can the PR title & description be updated to reflect the new (smaller) set of changes (ie. the current commit message)?

avaris (approved):
Thanks! Good to go :+1:

docs/settings.rst
arty-name: Sure! I have updated this line and the PR title and description. Please have a look.

justinmayer (approved):
Thanks for your patience, Tom, as well as for being willing to update the endeavor for 2021. Much appreciated! Looks good to me.

@iKevinY: Any last thoughts before merging?

iKevinY (approved):
LGTM! Thank you for the PR and your patience @arty-name; glad we can finally get this merged in! 😄

justinmayer:
Many thanks to @arty-name for the enhancement and to everyone here who helped review it. 🥇

arty-name:
Yay! Thanks everyone for reviewing and accepting this PR and for all your work on Pelican!

Issue #1038, opened 2013-08-12
Allow nice URLs on devserver


ssbarnea:
When you publish the website is quite easy to implement nice URLs. For example an nginx configuraiton like this would provide you with nice URLs for your website:

```
index index.php index.html index.htm index.shtml;
root /var/www/example.com/public_html;
try_files $uri $uri.html $uri/index.html index.html;
```

Now the problem is tha the devserver does not have the save behaviour, and so far I am now aware of any way to configure it to do this.

```
/output/index.html              http://localhost:8000
/output/contact.html              http://localhost:8000/contact
/output/sales/index.html              http://localhost:8000/sales
/output/sales/index.html              http://localhost:8000/sales/    (yes the same) 
```

PR #1441, opened 2014-08-19
Server script should try to rstrip '/' when serving pages

eliben:
The server script currently has capability to try to serve pages without a .html
suffix, to support mapping:

site.com/2013/06/some-slug --> site.com/2013/06/some-slug.html

However, some internal links in websites add a trailing slash after the slug,
even though it's a HTML file and not a directory:

site.com/2013/06/some-slug/ --> site.com/2013/06/some-slug.html

Augments the solution for #1038

### DIFF
--- a/pelican/server.py
+++ b/pelican/server.py
@@ -16,28 +16,35 @@ from six.moves import socketserver
 
 class ComplexHTTPRequestHandler(srvmod.SimpleHTTPRequestHandler):
     SUFFIXES = ['', '.html', '/index.html']
+    RSTRIP_PATTERNS = ['', '/']
 
     def do_GET(self):
         # cut off a query string
         if '?' in self.path:
             self.path, _ = self.path.split('?', 1)
 
-        # Try to detect file by applying various suffixes
-        for suffix in self.SUFFIXES:
-            if not hasattr(self, 'original_path'):
-                self.original_path = self.path
-
-            self.path = self.original_path + suffix
-            path = self.translate_path(self.path)
-
-            if os.path.exists(path):
-                srvmod.SimpleHTTPRequestHandler.do_GET(self)
-                logging.info("Found `%s`." % self.path)
+        found = False
+        # Try to detect file by applying various suffixes and stripping
+        # patterns.
+        for rstrip_pattern in self.RSTRIP_PATTERNS:
+            if found:
                 break
+            for suffix in self.SUFFIXES:
+                if not hasattr(self, 'original_path'):
+                    self.original_path = self.path
 
-            logging.info("Tried to find `%s`, but it doesn't exist.",
-                         self.path)
-        else:
+                self.path = self.original_path.rstrip(rstrip_pattern) + suffix
+                path = self.translate_path(self.path)
+
+                if os.path.exists(path):
+                    srvmod.SimpleHTTPRequestHandler.do_GET(self)
+                    logging.info("Found `%s`.", self.path)
+                    found = True
+                    break
+
+                logging.info("Tried to find `%s`, but it doesn't exist.", path)
+
+        if not found:
             # Fallback if there were no matches
             logging.warning("Unable to find `%s` or variations.",
                             self.original_path)


justinmayer:

pelican/server.py
```diff
@@ -23,16 +24,17 @@ def do_GET(self):
...
+                path = self.translate_path(self.path)
+                if os.path.exists(path):
+                    srvmod.SimpleHTTPRequestHandler.do_GET(self)
+                    logging.info("Found: %s" % self.path)
+                    found = True
+                    break
+                logging.info("Tried to find file %s, but it doesn't exist. " % self.path)
         if not found:
-            logging.warning("Unable to find file %s or variations.", self.path)
+            logging.warning("Unable to find file %s or variations." % self.path)
```
leotrs: I think the previous version of this line, with a `,` instead of the `%` is the preferred way to use logging.

pelican/server.py
```diff
@@ -23,16 +24,17 @@ def do_GET(self):
...
-                break
-            logging.info("Tried to find file %s, but it doesn't exist. ", self.path)
+            for strip_pattern in RSTRIP_PATTERNS:
+                self.path = self.original_path.rstrip(strip_pattern) + suffix
+                path = self.translate_path(self.path)
+                if os.path.exists(path):
+                    srvmod.SimpleHTTPRequestHandler.do_GET(self)
+                    logging.info("Found: %s" % self.path)
+                    found = True
+                    break
```
leotrs: Does this break from the inner-most loop only? If a path is found with the first suffix, and we `break`, will we still try out all the other suffixes?

eliben: Thank you for the review! This code has fallen behind so I updated it with the recent pelican. Replacing the for...else logic with an explicit "found" flag now since we have a double loop. The strip pattern loop "prefers" no stripping to ensure we find the original path first.

pelican/server.py
eliben: OK, left as it was though now it's inconsistent within the same function :)

pelican/server.py
leotrs: What I meant is that when/if we find the correct path, we execute a `break`, but that only breaks from the inner-most loop, and the outer-most loop will continue iterating, even when we have found the path. I *think* you want to check the `found` flag at the top of every outer-most iteration. I hope that makes sense :)

eliben: Thanks, Done.

pelican/server.py
eliben: OK I changed everywhere to a `,`

eliben:
Thanks for the review!

I've now fixed conflicts with the latest Pelican code, and the PR should be up-to-date.

Hope this can be squashed/merged now.

pelican/server.py
```diff
@@ -16,24 +16,30 @@
 
 class ComplexHTTPRequestHandler(srvmod.SimpleHTTPRequestHandler):
     SUFFIXES = ['', '.html', '/index.html']
+    RSTRIP_PATTERNS = ['', '/']
```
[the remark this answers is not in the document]
oulenz: Sorry for entering this discussion at this late stage, but I wonder whether serving directories should really be prioritised over a html file. I would propose instead:
```
SUFFIXES = ['/index.html', '.html', '']
RSTRIP_PATTERNS = ['/', '']
```
In fact, as is, that final element `'/index.html'` doesn't do anything, since if `os.path.exists(path + '/index.html')`, then also `os.path.exists(path + '')`. The reason why it seems to work now is because `SimpleHTTPRequestHandler.send_head` goes looking for index files if the path in the get request is a directory.

I realise this was introduced by #1040, but it might be opportune to fix that here. If you'd rather merge this first I can open a new issue/pr.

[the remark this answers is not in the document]
eliben: This is probably a separate discussion? I'd prefer to get this merged separately, to be honest.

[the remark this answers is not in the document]
oulenz: Fair enough!

justinmayer:
Any last thoughts from @getpelican/reviewers regarding this PR?

avaris:
Looks good once it's squashed :).

eliben:
@justinmayer @avaris are you using the "Squash and Merge" feature on Github or do you prefer me to squash the commits and update the PR to have a single commit?

eliben:
Friendly ping :)

avaris:
I'm not sure about the Github's feature but [we usually want](https://github.com/getpelican/pelican/blob/master/CONTRIBUTING.rst#using-git-and-github) the author to cleanup the PR.

eliben:
No problem, I can squash the commits into one

eliben:
Squashed now!

[Travis fails in 3.5 only in what seems like some sort of tool failure unrelated to the code -- https://travis-ci.org/getpelican/pelican/jobs/362604537]

PR #1453, opened 2014-08-31
Fix #1405: No error message for invalid syntax in period_archives.html

### LINKED ISSUE #1405: No error message for invalid syntax in period_archives.html
I have noticed that my monthly archives are using archives.html template. With some debugger help I understood that whenever there is a syntax error in period_archives.html then no error is shown and archives.html is used instead.

Would be nice to show proper error message when jinja finds existing period_archives.html but cannot process it.

arty-name:
#1405 resulted from the wrapping of TemplateNotFound in generic Exception to adjust the error message. After that it was not possible to distinguish any other exception  from TemplateNotFound exception. 

I have introduced a new PelicanTemplateNotFound exception and started using it to wrap jinja2.TemplateNotFound and to fallback from period_archives to archives.

Another possible approach is not to wrap exception in this case when we want to throw it away. This could be achieved by either parametrising get_template or by extracting get_template_without_wrapping into a separate function. Both seem to me clumsier  compared to the approach I have chosen.

### DIFF
--- a/pelican/generators.py
+++ b/pelican/generators.py
@@ -29,6 +29,9 @@ from pelican import signals
 logger = logging.getLogger(__name__)
 
 
+class PelicanTemplateNotFound(Exception):
+    pass
+
 @python_2_unicode_compatible
 class Generator(object):
     """Baseclass generator"""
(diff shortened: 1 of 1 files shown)


justinmayer:
Thanks for the contribution, Artemy. Please accept my sincere apologies for the long delay in getting this merged.

arty-name:
Better late than never! Thanks @justinmayer!

PR #1455, opened 2014-09-05
Additional filter for TAG feed generation

aravindavk:
New setting "TAGS_FOR_FEED" is introduced, default is []
If TAGS_FOR_FEED is set then it will generate feed only for
those tags.

Example usage
TAGS_FOR_FEED = ["thats", "awesome"]

Signed-off-by: Aravinda VK mail@aravindavk.in

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -494,6 +494,9 @@ Setting name (followed by default value, if any)    What does it do?
 ``TAG_FEED_RSS = None``, i.e. no RSS tag feed       Relative URL to output the tag RSS feed
 ``FEED_MAX_ITEMS``                                  Maximum number of items allowed in a feed. Feed item
                                                     quantity is unrestricted by default.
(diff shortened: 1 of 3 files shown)


justinmayer:

avaris:
I'm failing to understand the incentive for this new setting. What's the problem with creating additional tag feeds? Why would you want to limit it? Just to save some disk space and minimal build time benefits?

If so, why don't the `author` and `category` get similar option?

justinmayer:
After thinking about this a bit more, I don't think it's worth adding a new setting to suppress feed generation for selected tags. It feels like an edge case that could be handled with Invoke automation, removing any tag feeds (post-generation) that aren't in a specified list.

davidag:
Probably the [feed-filter plugin](https://github.com/pelican-plugins/feed-filter) provides this feature in a more general way.

Issue #1467, opened 2014-09-16
Typogrify needs to be rewritten from scratch for Pelican


barrysteyn:
I have great respect and appreciation for the original authors of Typogrify, so let me begin by saying thank you to them and stating from the outset that this issue is not meant to reflect badly or be personal.

Typogrify as it is used currently in Pelican is broken for the following reasons:
 1. It has been hacked together (I know, I have been altering the code base for some time)
 2. It is quite messy, and the logic does not work as expected (if you look carefully, the Typogrify demo HTML does not apply the widont filter as advertised)
 3. It gives problems with tags that must not be typeset, and as it is used in the Pelican project, is not very configurable

Here is an example of what I mean: I suggested to @justinmayer that Typogrify specifically ignore JavaScript tags `<script>`. @justinmayer informed that it does not touch JavaScript tags, but this is slightly incorrect: Only the Caps and smartypants filters ignore JavaScript, and the Widont filter is applicable to certain tags (therefore ignoring script tags). But the other filters are applied to JavaScript. Since Typogrify is applied early on by the reader, JavaScript is not really affected.

I propose to fix Typogrify by providing a new class for Pelican that accomplishes what the current Typogrify does but in a very clean and neat fashion. It will do the following:
  1. Fixes up the regular expressions, making it less complex.
  2. Makes it easier to extend and configure for the Pelican project
  3. Fixes the logic to totally ignore certain tags (like the script tag)
  4. Provide a codebase that is well documented so that future developers can extend it

honzajavorek:
FWIW, I tried to code my own replacement for Typogrify called [tipi](https://github.com/honzajavorek/tipi), but it's not production-ready yet. I wanted to use it with Pelican.

justinmayer:
It's not clear to me that issues with Typogrify can't be addressed without ditching it and vendoring our own solution. But feel free to make that case here if you wish and I can re-open this issue for further discussion.

PR #1468, opened 2014-09-16
Addresses #1467

barrysteyn:
This is a first draft - I would like the community to debate and test for me. Unit tests have not been written (yet), but they will come if the community wants this contribution


justinmayer:
It's not clear to me that issues with Typogrify can't be addressed without ditching it and vendoring our own solution. But feel free to make that case here if you wish and I can re-open this pull request.

PR #1481, opened 2014-09-27
add objects details to theme docs

hrbonz:
This PR came out of a frustration of not knowing what I could do with the various objects in my templates. Just thought this could be a good addition to the docs.

I willingly made a decision not to describe all the attributes to keep only those I think are interesting in the context of a template.

Let me know what you think.


justinmayer:
Thanks for the documentation enhancement, Stefan!

PR #1489, opened 2014-10-03
Fix using {filename} on Windows

bugnano:
Python on Windows uses the '\\' character for paths on os.path functions, while on source files usually the '/' character is used to separate paths.

self._context['filenames'] has the file names separated by '\\' on Windows, so if a source file references a file with

```
{filename}/dir1/dir2/referenced_file.rst
```

this would fail, as self._context['filenames'] has this file mapped to `dir1\dir2\referenced_file.rst`

This patch solves the problem in an os-portable way, that is it separates the source path '/' characters, and it replaces them with the correct os-dependent path separator, so it works both on Windows and on Linux.


bugnano:
Hello? Is anyone still interested in Windows compatibility?

I can provide you more information on why this patch is required in order to make {filename} links work on Windows, if that is required for merging this pull request.

Or maybe we can discuss about a different solution that addresses this issue, but it needs to be addressed, because, as of now, {filename} links work only if the path separator used in the {filename} links is the same as os.sep, which means using the \\ character instead of / on Windows, which is a real issue if the same website is maintained by different people that use different operating systems.

smartass101:
Your solution looks quite sane. So you've made sure that it works on all the supported platforms? I suppose the test suite covers this already, or are new tests required?

smartass101:
Perhaps this should be documented at least.

bugnano:
I have tested this patch on both Windows (XP 32-bit) and Linux (Fedora 20 x86_64). I don't have any other systems to test.
Perhaps the problem with my patch could be that using the \\ character as a path separator for {filename} links will stop working, or maybe using the \\ characher as a path separator never worked, I didn't test this case neither before, nor after my patch, so I don't know whether this would break compatibility with some websites built using only Windows.

bugnano:
Ok, with this new commit I have documented what this patch is doing, plus I have addressed the possible problem with the \\ character in {filename} links. Now the {filename} links can have both the \\ and / characters as valid path separators on any platform, and Pelican will do the Right Thing (TM).

bugnano:
Thank you for noticing where the mistake was, the problem is that I modified the contents.py both in my local machine running Linux, and on a virtual machine running Windows, and I tested the patch only on the virtual machine, whereas I pushed the commit from my local machine. It should be all right now.

smartass101:
Looks quite good, I would just like to see some test with mixed path separators in the path. AFAIK `re` caches expressions, but if you can, please try whether it would be any faster when using `path_sep_regexp = re.compile('[\\/]')` at the module level and then just use `path_sep_regexp.match(path)`.

bugnano:
Being a single character class regular expression, I think the preformance difference between the precompiled expression and the module level functions is negligible, however having a regex object surely increases the maintainability of the project, because the regex string appears only once.

smartass101:
Good point @bugnano. Now what remains is the question of unit tests. Since the current test suite passed, it should be ok. However, there could be at least a test for mixed separator paths added. What do you think?

bugnano:
I agree that having a unit test for mixed path separators is a good idea, but I don't know how to do them, if you tell me how you do unit tests, I could try making one, or if there is someone willing to make this unit test I would appreciate it.

justinmayer:
Closed in favor of #1581. If there is work here that isn't addressed by that PR, please re-submit as a new pull request. Thanks!

PR #1491, opened 2014-10-06
Ignore empty metadata. Fixes #1469. Fixes #1398.

### LINKED ISSUE #1398: Tag with empty name created for line `tags: `
When migrating my old blog to pelican I have automatically exported old contents. This caused some of these files to have tags line without actual tags like that:

    tags: 

When pelican was parsing these files it has created a page for tag with empty name.

### LINKED ISSUE #1469: Empty slug causes generation of hidden file
Having this line in my markdown source:

    Slug:

Causes pelican to generate a file called `.html` in the output directory. Being a "dot file", it is hidden on unix-like systems, and probably not something that anyone wants. I think it would make more sense to behave as if there was no Slug: line at all.

foresto:
Some metadata values cause problems when empty.  For example, a markdown file containing a `Slug:` line with no additional text causing Pelican to produce a file named `.html` instead of generating a proper file name.  Others, like those created by a `PATH_METADATA` regex, must be preserved even if empty, so things like `PAGE_URL="filename{customvalue}.html"` will always work. Essentially, we want to discard empty metadata that we know will be useless or problematic.  This is better than raising an exception because (a) it allows users to deliberately keep empty metadata in their source files for filling in later, and (b) users shouldn't be forced to fix empty metadata created by blog migration tools (see #1398).

The metadata processors are the ideal place to do this, because they know the type of data they are handling and whether an empty value is wanted. Unfortunately, they can't discard items, and neither can `process_metadata()`, because their return values are always saved by calling code.  We can't safely change the calling code, because some of it lives in custom reader classes out in the field, and we don't want to break those working systems. Discarding empty values at the time of use isn't good enough, because that still allows useless empty values in a source file to override configured defaults.

My solution:
- When processing a list of values, a metadata processor will omit any unwanted empty ones from the list it returns.
- When processing an entirely unwanted value, it will return something easily identifiable that will pass through the reader code.
- When collecting the processed metadata, read_file() will filter out items that were identified by the processors as unwanted.

These metadata are affected by this change:
- author
- authors
- category
- slug
- status
- tags

I also removed a bit of now-superfluous code from generators.py that was discarding empty authors at the time of use.

### DIFF
--- a/pelican/generators.py
+++ b/pelican/generators.py
@@ -544,10 +544,8 @@ class ArticlesGenerator(CachingGenerator):
             if hasattr(article, 'tags'):
                 for tag in article.tags:
                     self.tags[tag].append(article)
-            # ignore blank authors as well as undefined
             for author in getattr(article, 'authors', []):
-                if author.name != '':
-                    self.authors[author].append(article)
+                self.authors[author].append(article)
         # sort the articles by date
         self.articles.sort(key=attrgetter('date'), reverse=True)
         self.dates = list(self.articles)
--- a/pelican/readers.py
+++ b/pelican/readers.py
@@ -28,15 +28,43 @@ from pelican.contents import Page, Category, Tag, Author
 from pelican.utils import get_date, pelican_open, FileStampDataCacher, SafeDatetime, posixize_path
 
 
+def strip_split(text, sep=','):
+    """Return a list of stripped, non-empty substrings, delimited by sep."""
+    items = [x.strip() for x in text.split(sep)]
+    return [x for x in items if x]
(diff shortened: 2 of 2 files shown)


justinmayer:
Hi Forest. So sorry for the delay in reviewing your pull request. If you would be so kind as to rebase on current master, I'd be happy to merge it posthaste.

foresto:
Done. Thanks, Justin. I also rebased #1515, in hopes that you'll get a chance to review that one as well.

justinmayer:
Many thanks, Forest!

PR #1495, opened 2014-10-09
Add more docs for `--write-selected` option

mlissner:
I confess I tried just about every combination of generated file or `md` file before finally getting this right. An example would have helped me tremendously and so I've put an example in here. I also struggled to find this (who looks in settings for details about generation?) so I added a reference from the quickstart page so people can find it more easily in the settings.

### DIFF
--- a/docs/publish.rst
+++ b/docs/publish.rst
@@ -1,6 +1,9 @@
 Publish your site
 #################
 
+
+.. _site_generation:
+
 Site generation
 ===============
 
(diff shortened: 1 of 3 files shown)


MinchinWeb:
Can this, or something like it be added to the documentation? Google is leading to this PR when trying to figure out how `--write-selected` works...

mlissner:
FWIW, it looks like this PR still merges cleanly. Is there something we're waiting on to merge it?

justinmayer:
@mlissner: Merged via 4aa60a09. Sincere apologies for the delay. Thanks for the contribution!

PR #1503, opened 2014-10-15
backslash(\) in pagination urls

ranqiangjun:
eg: 
http://sample.com/category\abc2.html
http://sample.com/category\abc3.html

### DIFF
--- a/pelican/themes/simple/templates/pagination.html
+++ b/pelican/themes/simple/templates/pagination.html
@@ -1,11 +1,11 @@
 {% if DEFAULT_PAGINATION %}
 <p class="paginator">
     {% if articles_page.has_previous() %}
-        <a href="{{ SITEURL }}/{{ articles_previous_page.url }}">&laquo;</a>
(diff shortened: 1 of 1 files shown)


trwired:
I just stumbled over this bug on Windows. However, shouldn't the problem be addressed by fixing the Page object rather than replacing backslashes in the template? Putting it there treats the symptoms, not the cause, and every theme that doesn't base upon the simple theme will need to reimplement that fix.

avaris:
@trwired I believe this was fixed by #1581. Did you test it with current master?

trwired:
I just did and after some testing it appears that the problem indeed went away. Thanks!

justinmayer:
@ranqiangjun: Please test with current master and comment here if the problem persists. Thanks!

ranqiangjundrupal:
I reported this bug last year, and it was fixed 18 Feb this year by https://github.com/getpelican/pelican/pull/1581.  @justinmayer

ranqiangjun:
I reported this bug last year, and it was fixed this year by #1581 as @avaris  mentioned. @justinmayer @trwired.

PR #1515, opened 2014-11-01
Guess mime type with python-magic if available. Fixes #1514.

### LINKED ISSUE #1514: Built-in server relies on file name suffix to guess Content-Type
Pelican's built-in web server relies on the standard `SimpleHTTPServer` module to guess the appropriate `Content-Type` header for the files it serves. Sadly, that implementation relies on file name suffix to make its guesses. When I configure my site to use URLs without suffixes...

    'PAGE_URL': 'pages/{slug}'

...the server sends `Content-Type: application/octet-stream`, and my browser refuses to render the HTML.

This could be better, at least on systems that have the python-magic package installed.

foresto:
This makes Pelican's built-in server better at guessing mime types, if a relatively recent version of the python-magic package is installed. It only kicks in when the default guessing code fails to guess, and gracefully degrades if python-magic is unavailable or old.

### DIFF
--- a/pelican/server.py
+++ b/pelican/server.py
@@ -12,6 +12,11 @@ try:
 except ImportError:
     import socketserver  # NOQA
 
+try:
+    from magic import from_file as magic_from_file
+except ImportError:
+    magic_from_file = None
+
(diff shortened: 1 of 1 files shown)


justinmayer:
Thanks, Forest!

PR #1525, opened 2014-11-10
Fix up the datetime comparison error caused by timezone.

tonyseek:
If the article's date is timezone-carried, "now" should be too. Or `can't compare offset-naive and offset-aware datetimes` will be raised.

### DIFF
--- a/pelican/contents.py
+++ b/pelican/contents.py
@@ -11,6 +11,7 @@ import os
 import re
 import sys
 
+import pytz
 
 from pelican import signals
 from pelican.settings import DEFAULT_CONFIG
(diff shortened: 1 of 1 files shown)


sassman:
seems that this issue is also on the sitemap generation:

`  File "/Users/d34dl0ck/workspaces/web/www.d34dl0ck.me/devblog/pelican-plugins/sitemap/sitemap.py", line 188, in generate_output
    self.set_url_wrappers_modification_date(self.context['categories'])
  File "/Users/d34dl0ck/workspaces/web/www.d34dl0ck.me/devblog/pelican-plugins/sitemap/sitemap.py", line 171, in set_url_wrappers_modification_date
    lastmod = max(lastmod, article.date)
TypeError: can't compare offset-naive and offset-aware datetimes
make: *** [html] Error 1
`
using pelican version 3.7.1

PR #1533, opened 2014-11-17
Replace underscores in dates with spaces before parsing

kernc:
This allows for parsing date, and time, from the filename
```
FILENAME_METADATA = '(?P<date>\d{4}-\d{2}-\d{2}(_\d{2}:\d{2})?)_(?P<slug>.*)'
```
The format version with `_` is somewhat nicer on Unix and more readable than the one with a `T`.
It's an edgy use case but the drawbacks are few, if any.


justinmayer:
Thanks for the contribution, @kernc. Apologies for the delay in getting this in.

kernc:
Oh, no problem. I wasn't sure this would ever get in. I'm glad it did. Thanks. :smiley:

PR #1560, opened 2014-12-15
Set logging level to INFO

FedericoCeratto:
Set logging level to INFO to make log messages visible to the user. 
By default, messages with lower priority than logging.WARNING are ignored, making the few calls to logging.info useless. Now the "Serving at port ..." and "Shutting down server" are being printed.

### DIFF
--- a/pelican/server.py
+++ b/pelican/server.py
@@ -13,6 +13,9 @@ except ImportError:
 from six.moves import SimpleHTTPServer as srvmod
 from six.moves import socketserver
 
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
 
 class ComplexHTTPRequestHandler(srvmod.SimpleHTTPRequestHandler):
     SUFFIXES = ['', '.html', '/index.html']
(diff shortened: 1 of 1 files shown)


justinmayer:
Hi @FedericoCeratto. Any thoughts regarding the feedback from @smartass101?

justinmayer:
Hi @FedericoCeratto. Would you please take another pass at this PR based on the provided feedback so we can review and potentially merge your contribution?

FedericoCeratto:
Switching to a logger is a little bit more invasive but nicer. PR updated.

justinmayer:
@FedericoCeratto: This PR is showing conflicts. Could you resolve?

justinmayer:
Perhaps due to recent changes to Pelican's built-in server module, I am seeing the following messages when using `pelican --server` on current master:
```
Serving site at: 127.0.0.1:8000 - Tap CTRL-C to stop
^C
Keyboard interrupt received. Shutting down server.
```
So I get the impression the impetus behind this pull request may no longer be relevant. If that assumption is wrong, and there are still reasons for related follow-up work regarding this topic, please comment here and I will re-open.

PR #1566, opened 2014-12-23
Extract tag_cloud generation into its own method

recursify:
The `generate_context` method is getting kind of long... starting to strip it down by moving some of the logic into their own methods.  The `tag_cloud` stuff is pretty self-contained.


ingwinlu:
@justinmayer shoot your lazer

justinmayer:
Since tag clouds are now out of core and implemented via a plugin, it would appear this pull request is no longer relevant. (Thanks for the reminder, @ingwinlu.)

PR #1578, opened 2015-01-02
Use Shields.io as source for Travis badge

iKevinY:
Currently, the Travis badge is rendering using the "flat" style while the Coveralls badge uses the "plastic" style, which looks a little odd. Since the Coveralls badge already uses Shields, I think it would make sense for the Travis badge to do the same, for the sake of consistency. This way, any changes made to badge styling by the Shields team will be applied to both simultaneously.


kylef:
Thanks.

iKevinY:
No worries; thanks for the merge!

PR #1579, opened 2015-01-03
docs: internal link typo

kermit666:
The link didn't work without the `/`.

### DIFF
--- a/docs/content.rst
+++ b/docs/content.rst
@@ -188,7 +188,7 @@ and ``article2.md``::
 
     See below intra-site link examples in Markdown format.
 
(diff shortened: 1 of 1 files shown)


avaris:
No, that's not a typo. The two are different. One is a relative path to `content` folder, the other is a relative path to the file that links it.

justinmayer:
Deniz is correct. In the future, it might be better to describe the perceived problem, along with a detailed list of steps to reproduce.

kermit666:
Ah, OK. I didn't catch that part - I was skimming through the docs a bit too quickly, sorry. Thanks for the clarification!

PR #1581, opened 2015-01-05
Fix Pelican rendering and unit tests on Windows.

georgevreilly:
In order to [hand off webmaster duties](http://georgevreilly.github.io/blog/2015/01/03/ForWantOfANail.html) on a [Pelican-based site](https://github.com/nwcpp/pelican-site), I had to ensure that the website built on Windows, in addition to Mac. I've fixed the output rendering issues and fixed the unit tests too.

The fixes fall into the following categories.
* Assuming that file system relative paths (`/` or `\`) and URL paths (always `/`) can be used interchangeably. Output URL paths should be constructed with `posixize_path` or `posix_join`.
* CRLF vs LF problems. `pelican_open` replaces CRLFs with LFs. I also had to strip CRLF warnings from `git diff` output.
* Locale support. There were three locale-related tests that I just could not get to pass with Python 2.7.6 (default, Nov 10 2013, 19:24:18) [MSC v.1500 32 bit (Intel)]. I suspect that the Python binary links with ANSI (not Unicode) MSVC libraries. Those tests are now skipped on Windows.

I believe you should be able to set up [free Windows-based Continuous Integration at Appveyor](https://packaging.python.org/en/latest/appveyor.html) (via [ionel](http://blog.ionelmc.ro/2015/01/02/speedup-pip-install/)). This will stop the fixes from rotting.

### DIFF
--- a/docs/content.rst
+++ b/docs/content.rst
@@ -157,6 +157,9 @@ the other content will be placed after site generation).
 
 To link to internal content (files in the ``content`` directory), use the
 following syntax for the link target: ``{filename}path/to/file``
+Note: forward slashes, ``/``,
+are the required path separator in the ``{filename}`` directive
+on all operating systems, including Windows.
 
 For example, a Pelican project might be structured like this::
 
--- a/docs/publish.rst
+++ b/docs/publish.rst
@@ -100,6 +100,18 @@ separately. Use the following command to install Fabric, prefixing with
 
     pip install Fabric
 
+.. note:: Installing PyCrypto on Windows
+
+    Fabric depends upon PyCrypto_, which is tricky to install
+    if your system doesn't have a C compiler.
+    For Windows users, before installing Fabric, use
+    ``easy_install http://www.voidspace.org.uk/downloads/pycrypto26/pycrypto-2.6.win32-py2.7.exe``
+    per this `StackOverflow suggestion <http://stackoverflow.com/a/11405769/6364>`_
+    You're more likely to have success
+    with the Win32 versions of Python 2.7 and PyCrypto,
+    than with the Win64—\
+    even if your operating system is a 64-bit version of Windows.
+
 Take a moment to open the ``fabfile.py`` file that was generated in your
 project root. You will see a number of commands, any one of which can be
 renamed, removed, and/or customized to your liking. Using the out-of-the-box
@@ -179,3 +191,4 @@ executables, such as ``python3``, you can set the ``PY`` and ``PELICAN``
 environment variables, respectively, to override the default executable names.)
 
 .. _Fabric: http://fabfile.org/
+.. _PyCrypto: http://pycrypto.org
--- a/docs/tips.rst
+++ b/docs/tips.rst
@@ -36,9 +36,15 @@ already exist). The ``git push origin gh-pages`` command updates the remote
 ``gh-pages`` branch, effectively publishing the Pelican site.
 
 .. note::
+    The ``github`` target of the Makefile (and the ``gh_pages`` task of the Fabfile)
+    created by the ``pelican-quickstart`` command
+    publishes the Pelican site as Project Pages, as described above.
 
-    The ``github`` target of the Makefile created by the ``pelican-quickstart``
-    command publishes the Pelican site as Project Pages, as described above.
+.. note:: ghp-import on Windows
+
+    Until `ghp-import Pull Request #25 <https://github.com/davisp/ghp-import/pull/25>`_
+    is accepted, you will need to install a custom build of ghp-import:
+    ``pip install https://github.com/chevah/ghp-import/archive/win-support.zip``
 
 User Pages
 ----------
@@ -85,6 +91,12 @@ output directory. For example::
 
     STATIC_PATHS = ['images', 'extra/CNAME']
     EXTRA_PATH_METADATA = {'extra/CNAME': {'path': 'CNAME'},}
+
+Note: use forward slashes, ``/``, even on Windows.
+
+.. hint::
+    You can also use the ``EXTRA_PATH_METADATA`` mechanism
+    to place a ``favicon.ico`` or ``robots.txt`` at the root of any site.
 
 How to add YouTube or Vimeo Videos
 ==================================
--- a/pelican/contents.py
+++ b/pelican/contents.py
@@ -17,7 +17,7 @@ from pelican import signals
 from pelican.settings import DEFAULT_CONFIG
 from pelican.utils import (slugify, truncate_html_words, memoized, strftime,
                            python_2_unicode_compatible, deprecated_attribute,
-                           path_to_url, set_date_tzinfo, SafeDatetime)
+                           path_to_url, posixize_path, set_date_tzinfo, SafeDatetime)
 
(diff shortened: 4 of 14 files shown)


justinmayer:
Many thanks for the enhancements, George. Would you mind taking a look at #1489 and determine whether your changes render it moot? (i.e., whether we can close that pull request in favor of this one)

georgevreilly:
I'm going to have to experiment with #1489. That might be a bit cleaner than some of what I did. Mine is more comprehensive, as I fixed a pile of tests. Maybe tonight.

georgevreilly:
I experimented with #1489. The one thing that it does that my patch doesn't do is that it accepts `\` as well as `/` in `{filename}` links, such as `{filename}\foo\bar.rst`. I'm not sure that this is actually desirable in the input.

Moreover, if you run on Windows, all the links rendered in HTML contain backslashes; e.g., `<link href="http://nwcpp.org/feeds\all.atom.xml"`. While most browsers will  accept backslashes in URLs, the canonical form of a URL contains a forward slash. Indeed, I can find no mention of backslashes in [RFC 3986](https://tools.ietf.org/html/rfc3986). The unit tests don't pass with #1489. One reason is that a number of the tests compare rendered output against `pelican/tests/output` where all the slashes lean to the right.

PR #1489 also fails to handle backslashes in several important cases, such as 
```
EXTRA_PATH_METADATA = { 'extras/CNAME': {'path': 'CNAME'} }
```

I commend @bugnano for fixing the most serious problem on Windows. However,  I recommend not accepting #1489.

bugnano:
Well, the backslash character in URLs is an error that must be fixed, however I see #1489 as the first step in the right direction, there is only some more work to do.
Allowing the backslash character in {filename} links, if handled correctly, allows more flexibility, because Windows users are more used to the backslash character as a path separator, and {filename} links refer to a local file, not to an URL.
As far as the EXTRA_PATH_METADATA is concerned, we must ensure that the keys are coherent with the internal representation.
Let me explain:
I started doing the #1489 patch because the keys of self._context['filenames'] use the os.sep character as path separator.
Now, there are 2 ways of ensuring that the {filename} links are coherent with self._context['filenames']:
1. Change the internal representation of files, in this case self._context['filenames'], in order to have only the '/' character as a path separator
2. Change the {filename} link string, in order to ensure that it uses os.sep as the path separator.
Both solutions work, but I felt that solution number 2 requires less work.
In the same way, we must ensure that EXTRA_PATH_METADATA uses the same path separator as the internal representation, I don't know what object, as I didn't look that part.
Another important thing to take into consideration is not only that the path separators accepted in {filename} links, EXTRA_PATH_METADATA, and any other user definable path are coherent with the current operating system, but we must ensure that they work on Windows, and also on other operating systems, as in my case where the same website is maintained both by me using Linux, and by another guy using Windows.
In order to correctly address this issue, we could instead go with option number 1, that is replacing every os.path function with its posixpath equivalent everywhere. This would ensure that the internal representation of the paths always have the '/' character as a path separator, which is accepted on Windows anyway, but I understand that such change would be enormous, and should be tested extensively.

georgevreilly:
I'm not convinced that backslash needs to be supported in input text on Windows. Anyone who uses Pelican has managed to install Python; install Pelican and all its dependencies; perhaps installed Fabric and PyCrypto; learned ReStructuredText or Markdown; and figured out how to publish their content to a website. They can certainly cope with putting `/` instead of `\` in their content source.

My patch does your option number 1, by the way.

bugnano:
Sounds good to me, it only needs to be documented.
As long as this pull request fixes the {filename} links problem on Windows, I'm happy to have this pull request accepted instead of #1489 .

georgevreilly:
I can write some documentation if you tell me roughly what you want covered and where in the docs it should go.

bugnano:
In the content.rst doc at the "Linking to internal content" section it would be nice to specify explicitly that paths should be separated by forward slashes even on Windows.

I don't know whether there are other parts of Pelican where paths should be separated by forward slashes (EXTRA_PATH_METADATA?), but it would be nice to specify that on the documentation of those parts too.

georgevreilly:
I added some notes about forward slashes, as well as some notes about Fabric, ghp-import, and PyCrypto on Windows.

justinmayer:
@bugnano: What do think about George's documentation additions?

bugnano:
I like them very much, especially the part that describes the PyCrypto installation, which has always been tricky on Windows.

justinmayer:
Thanks for the feedback, Franco.

@avaris, @ingwinlu, et al...  Any other comments before we merge?

avaris:
I just tested this and works fine. There is one issue though, the following is not the way to skip tests:

        if platform == 'win32':
            return unittest.skip("Doesn't work on Windows")

This will mark the tests as passed, which is obviously not the case. It would be better to mark them as actually skipped, like:

        @unittest.skipIf(platform == 'win32', "Does not work on Windows")
        @unittest.skipUnless(locale_available('fr_FR.UTF-8'), 'French locale needed')
        def test_french_strftime(self):
            # ...

Similar for other tests that uses same logic to 'skip' them.

avaris:
I'm tinkering with (force) skipped tests. Most of them doesn't need to be skipped.

* `test_utils.TestUtils.test_strftime_locale_dependent_turkish` will pass if `locale.LC_ALL` is used instead of `locale.LC_TIME`.

* `test_utils.TestUtils.test_strftime_locale_dependent_french` doesn't have any issue but it might make sense to change it to `LC_ALL` as well (I like symmetry).

* `test_utils.TestDateFormatter.test_french_strftime` will pass if it set correct locale for windows. IOW, instead of:

        locale.setlocale(locale.LC_ALL, str('fr_FR.UTF-8'))

    it needs to do

        if platform == 'win32':
            locale.setlocale(locale.LC_ALL, str('French'))
        else:
            locale.setlocale(locale.LC_ALL, str('fr_FR.UTF-8'))

* `test_settings.TestSettingsConfiguration.test_default_encoding`... I got nothing for this. This has to be skipped I guess. `locale.getlocale()` and `locale.getdefaultlocale()` return completely different things on Windows.

**PS**: `test_french_locale` and `test_turkish_locale` (in `test_utils.TestDateFormatter`) are false positives on Windows because of the way `settings.read_settings` works. `read_settings` will ignore bad `LOCALE` option and `'fr_FR.utf-8'`/`'tr_TR.utf-8'` is invalid on Windows. As a result, they don't test stuff in different locale. `LOCALE` settings has to be set conditionally depending on `platform` (`'French'` vs `'fr_FR.utf-8'`) like other tests do. They will still pass after the change, but right now they don't test what they supposed to test.

(Yeah, I know... I wrote these tests and now I'm going to hide :).)

georgevreilly:
@avaris, thanks for digging into those tests. It's good to get i18n right. I spent a little time trying to make them pass, but failed. Unlike the other tests that I fixed under Windows, these weren't as crucial in getting Pelican running there.

justinmayer:
@avaris: What, if anything, do you think is needed here in order to proceed?

justinmayer:
@avaris will follow up in a separate PR.

Many thanks for all your work on this, George. Much appreciated!

PR #1582, opened 2015-01-05
DOCS: add comment to 'preview your site' section mentioning...

ingwinlu:
...the http server module for python3 is different then python2


ingwinlu:
thanks to `norg` from IRC for reporting

PR #1583, opened 2015-01-05
Documentation fixes

ingwinlu:
this pr fixes some minor and a big issue with the current documentation.

* http://pelican.readthedocs.org/en/latest/contribute.html#testing-on-python-2-and-3 <- if you look here you will notice missing headers due to errors regarding the rst structure. this is due to the include to ../CONTRIBUTE.rst using a different struc. 
* minor whitespace fixes i noticed while scrolling over
* pelican themes hyperlink was double defined with different targets

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -1,5 +1,5 @@
 Filing issues
--------------
+=============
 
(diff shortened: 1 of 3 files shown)


justinmayer:
Thanks, @ingwinlu!

PR #1588, opened 2015-01-10
Adding ability to listen on addresses other than localhost.

dhermes:
This is helpful for mobile testing of Pelican sites by allowing broadcasting on the local network.

Using port 80 requires running as root on most machines.

### DIFF
--- a/pelican/server.py
+++ b/pelican/server.py
@@ -12,7 +12,8 @@ try:
 except ImportError:
     import socketserver  # NOQA
 
-PORT = len(sys.argv) == 2 and int(sys.argv[1]) or 8000
+PORT = len(sys.argv) in (2, 3) and int(sys.argv[1]) or 8000
+SERVER = len(sys.argv) == 3 and sys.argv[2] or ""
 SUFFIXES = ['', '.html', '/index.html']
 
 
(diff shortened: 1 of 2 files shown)


justinmayer:
Thanks for the contribution, Danny. It seems to me the proposed name for the new target (`serve-local`) sounds like the opposite of what it actually does — i.e., serve globally. Perhaps call this new target `serve-global`?

dhermes:
Yeah I [struggled][1] to find a good name. I don't think `serve-global` really works either, since it isn't really serving to the whole web.

What about `serve-local-network`? (Seems too long, do you have rules about length?) Maybe `serve-lan` / `serve-LAN`? Or `serve-nonlocal` (since not `localhost`)?

[1]: http://martinfowler.com/bliki/TwoHardThings.html

justinmayer:
> I don't think serve-global really works either, since it isn't really serving to the whole web.

How is it not? Doesn't using 0.0.0.0 as the IP address accomplish just that?

dhermes:
OK SGTM. Will push and ping you when updated.

dhermes:
Added change and pushed. Would you prefer I squash the two commits into a single one or is two commits for one PR OK with you?

justinmayer:
Squashed would be preferable, if it's not too much trouble.

dhermes:
Squashed and `push --force`ed.

justinmayer:
Thanks, Danny.

dhermes:
Sure thing.

dhermes:
@justinmayer Thanks!

What is the release schedule like? Is there some list I should look for to know when I can drop my custom `Makefile` / `pelican/serve.py` from my blog?

justinmayer:
My guess is that 3.6 will land in the next 4-5 weeks. Perhaps keep an eye on the blog's Atom feed or via Twitter `@getpelican`?

dhermes:
Will do. Thanks.

PR #1592, opened 2015-01-17
add argparse argument for relative urls

ingwinlu:
This PR adds the `--relative-urls` option to argparse.
Useful when combined with the `-r` flag, enabling easy local development without touching the configuration files.


justinmayer:
Thanks for the enhancement, Winlu!

PR #1593, opened 2015-01-18
Change the documented type of various settings

iKevinY:
Some of Pelican's settings are documented as tuples, but I think lists or dictionaries would be more appropriate when considering the data being represented (tuples being more appropriate for heterogeneous data). `DEFAULT_METADATA` in particular is listed as a tuple, when in reality it's supposed to be a tuple of tuples (which gets converted into a dictionary). This could potentially be misleading, especially since [this page](http://pelican.readthedocs.org/en/latest/content.html#publishing-drafts) defines it as a dictionary.

(This PR is really just an improvement on #1368, but I seem to have deleted that specific fork at some point, so I can't make any revisions to it.)


justinmayer:
Many thanks, Kevin!

iKevinY:
No worries!

PR #1595, opened 2015-01-18
make html's IDs unique on using multiple code-blocks with 'linespans' option activate

ghost:
Html's attribute "id" must be unique.

If you use multiple code-bocks with 'linespans' option activate, this affirmation isn't correct.
I suggest a little patch to correct this.

### DIFF
--- a/pelican/rstdirectives.py
+++ b/pelican/rstdirectives.py
@@ -71,7 +71,10 @@ class Pygments(Directive):
             if flag in self.options:
                 self.options[flag] = True
 
-        self.options['linespans'] += next(code_block_id)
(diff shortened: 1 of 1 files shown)


iKevinY:
You could also use [`itertools.count`](https://docs.python.org/2/library/itertools.html#itertools.count) to avoid having to writing an `integers()` function.

```python
import itertools
code_block_id = itertools.count(1)
```

justinmayer:
The individual who submitted this pull request has subsequently deleted the corresponding GitHub account. Anyone want to take this over and re-submit?

justinmayer:
Superseded by PR# 1599

PR #1597, opened 2015-01-21
Fix static path watchers

avaris:
There was an issue with static path watchers, where they were watching wrong paths.
They need to be prefixed with the `content` path. So, they were not working at all.

It was also possible to overwrite default watchers like `'content'`, `'settings'` and
`'theme'` by mistake if any of them were present in `STATIC_PATHS`. This is fixed
by adding a prefix to static watchers.

And static watchers were "too static", meaning, they stayed the same even if
`STATIC_PATHS` was changed in the settings during autoreload. Now static watchers
reflect those changes (i.e. new paths are added to watch list, and removed ones
are no longer watched).

### DIFF
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -359,8 +359,13 @@ def main():
                                         pelican.ignore_files),
                 'settings': file_watcher(args.settings)}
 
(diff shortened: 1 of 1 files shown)


justinmayer:
Thanks, @avaris!

PR #1598, opened 2015-01-22
Implements tag cloud sorting

magnunleno:
For those (like me) who don't like the tag cloud to be randomly organized, I've implemented some sorting options:

* Alphabetically;
* Alphabetically Reverse (descending);
* Size;
* Size Reverse (descending);
* Random.

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -632,6 +632,9 @@ Setting name (followed by default value)            What does it do?
 ``TAG_CLOUD_STEPS = 4``                             Count of different font sizes in the tag
                                                     cloud.
 ``TAG_CLOUD_MAX_ITEMS = 100``                       Maximum number of tags in the cloud.
+``TAG_CLOUD_SORTING = 'random'``                    The tag cloud ordering scheme.  Valid values:
+                                                    random, alphabetically, alphabetically-rev, size and
+                                                    size-rev
 ================================================    =====================================================
 
 The default theme does not include a tag cloud, but it is pretty easy to add one::
@@ -666,6 +669,7 @@ For example::
 
     ...
 
+By default the tags in the cloud are sorted randomly, but if you prefers to have it alphabetically use the `alphabetically` (ascending) and `alphabetically-rev` (descending). Also is possible to sort the tags by it's size (number of articles with this specific tag) using the values `size` (ascending) and `size-rev` (descending).
 
 Translations
 ============
--- a/pelican/generators.py
+++ b/pelican/generators.py
@@ -576,8 +576,20 @@ class ArticlesGenerator(CachingGenerator):
             )
             for tag, count in tag_cloud
         ]
-        # put words in chaos
-        random.shuffle(self.tag_cloud)
+
+        sorting = self.settings.get('TAG_CLOUD_SORTING')
+
+        if sorting == 'alphabetically':
+            self.tag_cloud.sort(key=lambda elem: elem[0].name)
+        elif sorting == 'alphabetically-rev':
+            self.tag_cloud.sort(key=lambda elem: elem[0].name, reverse=True)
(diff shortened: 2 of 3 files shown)


ingwinlu:
Hi,
this was always possible in themes, i.e. jinja2's build in [sort](http://jinja.pocoo.org/docs/dev/templates/#sort) filter.

Might be worth adding this as a plugin instead of an addition to pelican.

magnunleno:
Hi,

Despite it being possible to implement it in themes (as I used to do in my site), it will cause the site's building time to grow considerably. Since the sorting will be done in the theme, the tag cloud list sorting will be repeatedly executed for **every generated page**. Nowadays my site has something around 2000 pages, and it's taking 180~190 seconds to compile. With this feature implemented my site build time went down to 30~40 seconds.

Thanks anyway.

ingwinlu:
set in my `base.html`
```
{% set tag_cloud = tag_cloud|sort(attribute='0')%}
{% set tag_cloud_rev = tag_cloud|sort(attribute='0',reverse=True)%}
{% set tag_cloud_size = tag_cloud|sort(attribute='1')%}
{% set tag_cloud_size_rev = tag_cloud|sort(attribute='1',reverse=True)%}
{% macro list_tag_cloud(tags) -%}
<ul>
    {% for tag in tags %}
        <li class="tag-{{ tag.1 }}">
            <a href="{{ SITEURL }}/{{ tag.0.url }}">
                {{ tag.0 }}
            </a>
        </li>
    {% endfor %}
</ul>
{%- endmacro %}
```

used like this in `tags.html`:  [here](https://www.heroicdebugging.biz/tags.html) (only quickly hacked it together, i don't use a tag list usually)
```
{% block content %}
<div class="row">
    <div class="col-lg-12">
    <h2>tag_cloud</h2>
    {{ list_tag_cloud(tag_cloud)}}
    <h2>tag_cloud_rev</h2>
    {{ list_tag_cloud(tag_cloud_rev)}}
    <h2>tag_cloud_size</h2>
    {{ list_tag_cloud(tag_cloud_size)}}
    <h2>tag_cloud_size_rev</h2>
    {{ list_tag_cloud(tag_cloud_size_rev)}}
    </div>
</div>
{% endblock %}
```
no noticable slowdown on a rpi, 125s vs 130s
`Done: Processed 37 article(s), 1 draft(s) and 1 page(s) in 129.08 seconds.`

then again, there is use in having this functionality as a *plugin*.

magnunleno:
Setting it in `base.html` is even worse, it will sort the tag_cloud even when it isn't needed. In my theme I set the sorting to occur only when the tag cloud is included.

Your haven't had any noticeable slowdowns because you have a "small" site. My site has nearly 180 articles and something around 400 tags.

ingwinlu:
ah i thought the jinja environment would take care of caching then

magnunleno:
As far as I'm concerned it doesn't cache, since the tagcloud is altered and is never "written back" to the context that pelican uses.

ingwinlu:
that might be worth considering together with #1566.

feel like it needs tests even though you are basically just testing the build in sorting function.

avaris:
Maybe `tag_cloud` should become a plugin.

ingwinlu:
that certainly is an even better way forward

iKevinY:
I agree with turning `tag_cloud` into a plugin, since tag clouds aren't really a core feature for blogs, nor are they relevant at all to regular static sites.

ingwinlu:
I will prepare a PR then

ingwinlu:
@justinmayer pewpew here as well

justinmayer:
Since tag clouds are now out of core and implemented via a plugin, it would appear this pull request is no longer relevant. (Thanks for the reminder, @ingwinlu.)

PR #1599, opened 2015-01-22
make html's IDs unique on using multiple code-blocks with 'lineanchors' options

mothsART:
Html's attribute "id" must be unique.

If you use multiple code-bocks with 'linespans' option activate, this affirmation isn't correct.
I suggest a little patch to correct this.

Sorry for re-iterate a last PR : https://github.com/getpelican/pelican/pull/1595


mothsART:
See example here: http://195.154.252.48/linkmanager/webservice.html#code137-9

and source : https://github.com/mothsART/linkmanager_website

justinmayer:

pelican/rstdirectives.py
```diff
@@ -11,6 +11,16 @@
...
 
 
+def integers():
+    """Infinite sequence of integers."""
+    i = 1
+    while True:
+        yield str(i)
+        i = i + 1
+
+code_block_id = integers()
```
iKevinY: You could also use `itertools.count(start=1)` to create a generator that yields the sequence of natural numbers. Also, would this still work if it was a class variable belonging to `Pygments` rather than a global?

mothsART:
Hi,

I plunged in this old PR and suggest this new stashing commit.

First, thks to **iKevinY** for reduce code and also complexity.

Second, to pass unit tests, i need to init generator on each change of pelican settings. (due to the successive calls methods **test_basic_generation_works**, **test_custom_generation_works**, **test_custom_locale_generation_works**)

I'm not familiar to Pelican's core and this way is probably inelegant.

Third, to response to **iKevinY** :
global settings has no unit test.

To test, configure your pelicanconf.py like this
```
PYGMENTS_RST_OPTIONS = {
    'linenos': 'table',
    'anchorlinenos': '',
    'linespans': 'foo', 'lineanchors': 'foo'
}
```

**Unbelievable.rst** file bring test on a local directives (with a lineanchors prefix)

```
.. sourcecode:: python
    :anchorlinenos:
    :classprefix: testing
    :hl_lines: 10,11,12
    :lineanchors: foo
    :linenos: inline
    :linenospecial: 2
    :linenostart: 8
    :linenostep: 2
    :lineseparator: <br>
    :linespans: foo
    :nobackground:

       code
```
Fourth, I don't undersand why travis failed on docs.

Finaly,

This patch does work exclusively on **Rst** file, not **Markdown** (no real solution on pelican => need a patch on https://pypi.python.org/pypi/Markdown)

To my personnal usage, i create the **ace_editor** (https://github.com/getpelican/pelican-plugins) plugin and he cover all my needs.

mothsART:
I have found for the travis check : it's missing a dependence on tox.ini

PR #1602, opened 2015-01-27
Removed `PDF_GENERATOR=False`

ltiao:
It might give the impression that the PDF generation is still part of the core.


ingwinlu:
+1

PR #1603, opened 2015-01-27
Removed PDF Generation settings

ltiao:
Since PDF Generation is no longer a core feature, these settings are irrelevant. If the `pdf` plugin from pelican-plugins is used, it disregards the `PDF_GENERATOR` setting anyways.


ingwinlu:
+1

justinmayer:
Thanks, Louis!

PR #1608, opened 2015-02-01
Docs: Note how posts are published when 'draft' is default

abackstrom:
The docs show how posts can be defaulted to "draft" mode, but not how to publish a post when your site is in this state.

### DIFF
--- a/docs/content.rst
+++ b/docs/content.rst
@@ -505,6 +505,9 @@ publish an article before it is finished) include the status in the ``DEFAULT_ME
         'status': 'draft',
     }
 
+To publish a post when the default status is ``draft``, update the post's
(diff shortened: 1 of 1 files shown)


justinmayer:
I believe one can also remove the `Status: draft` line altogether, no?

iKevinY:
I think the PR intends to clarify how to publish a post if the user has set `draft` as the default `status` via `DEFAULT_METADATA` in Pelican's configuration.

justinmayer:
Ah, that makes sense. Thanks for pointing me in the right direction, Kevin.

iKevinY:
@justinmayer No problem!

justinmayer:
Thanks for the clarification, Annika!

PR #1616, opened 2015-02-07
Add signal: all_generators_finalized (see getpelican/pelican-plugins#314)

kernc:
Some plugins have used `content_object_init` signal and read `summary` or `content` properties of the content object. This resulted in internal (`{filename}`) links being unresolved. When used, this signal should hopefully mitigate that.

See also:
* https://github.com/getpelican/pelican-plugins/issues/314
* https://github.com/getpelican/pelican-plugins/pull/410


justinmayer:
Excellent work, @kernc. Many thanks for this addition!

PR #1617, opened 2015-02-07
Add articles with subparts.

pxquim:
Sub-articles are useful when an article has logically related parts
that are too large to be conveniently included in a single article,
yet are not independent enough to merit a separate article.

The site pxquim.pt uses sub-articles to associate multiple photo galleries
with a single article. Each photo gallery goes into a sub-article.

By convention, subparts of articles are articles with "--" in their name.
The part before the "--" should be the slug of the parent article.

For example:

* article.md        (the parent article)
* article--one.md   (a sub-article of "article.md")
* article--other.md (another sub-article of "article.md")

The article object gets a list of subparts[].

The sub-article object gets the following attributes parent:

* parent -- Reference to the parent article object.
* title -- Changed to start with the parent article plus a comma.
* subtitle -- The original title of the sub-article.

Sub-articles do not appear in blog indexes or in feeds, but tagged sub-articles appear in tag indexes.

--

NOTE: I am quite certain that this pull request is not up to the right standard, because it lacks tests and documentation. However:

- I would like feedback on whether sub-articles would be a useful feature.
- I could not figure out ant way to have sub-articles as a plugin.

Also, I am new to git and to GitHub. Sorry if I did something horribly wrong.

### DIFF
--- a/pelican/generators.py
+++ b/pelican/generators.py
@@ -346,7 +346,7 @@ class ArticlesGenerator(CachingGenerator):
 
     def generate_articles(self, write):
         """Generate the articles."""
-        for article in chain(self.translations, self.articles):
+        for article in chain(self.translations, self.articles, self.article_subparts):
             signals.article_generator_write_article.send(self, content=article)
             write(article.save_as, self.get_template(article.template),
                   self.context, article=article, category=article.category,
@@ -536,6 +536,30 @@ class ArticlesGenerator(CachingGenerator):
 
         signals.article_generator_pretaxonomy.send(self)
 
+        slugs= {}
+        for article in self.articles:
+            slugs[article.slug]= article
+        self.article_subparts= []
+        articles_main= []
+        for article in self.articles:
+            if '--' in article.slug:
+                (pslug, _)= article.slug.split('--', 1)
+                if pslug in slugs:
+                    parent= slugs[pslug]
+                    if not hasattr(parent, 'subparts'):
+                        parent.subparts= []
+                    parent.subparts.append(article)
+                    article.subpart_of= parent
+                    self.article_subparts.append(article)
+                    article.subtitle= article.title
+                    article.title= article.title + ", " + parent.title
+                else:
+                    logger.error('No parent for {}'.format(pslug))
+                    articles_main.append(article)
+            else:
+                articles_main.append(article)
+        self.articles= articles_main
+
         for article in self.articles:
             # only main articles are listed in categories and tags
             # not translations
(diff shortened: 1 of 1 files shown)


justinmayer:
Have you see the [Series plugin](https://github.com/getpelican/pelican-plugins/tree/master/series)? I suspect you'll find that it provides very similar functionality.

ingwinlu:
i am with @justinmayer, this has no part in pelican core

goanpeca:
What about subparts in a page, not an article. 

The series plugin works great with series of articles, but what if you want to keep editing of content simple (markdown) but provide a "complex" (grids rows columns ala bootstrap) for the styling of lets say a homepage or any other (non blog article) page?

Is it possible somehow to have subparts in a markdown file they would get rendered in the specific location in the template. Right now there is a content tag in the templates, but that's it.

Lets say a stupid... template `page.html` is:

```
{% extends "base.html" %}
{% block title %}{{ page.title }} - {{ SITENAME }}{% endblock %}

    {% block content %}
        {{ page.content }}
    {% endblock content %}

    {% block content-left %}
        {{ page.content-1 }}
    {% endblock content-right %}

    {% block content-right %}
        {{ page.content-2 }}
    {% endblock content-right %}
```

And you would write content in markdown for each of the parts.

```markdown
Title: My super title
Date: 2010-12-03 10:20
Modified: 2010-12-05 19:30
Category: Python
Tags: pelican, publishing
Slug: my-super-post
Authors: Alexis Metaireau, Conan Doyle
Summary: Short version for index and feeds

Content:
This is the content of my super blog post.

Content-1:
This is the additional content 

Content-2:
This is the additional content ...
```

There are some extensions to include a grid in markdown, but these force extra nomenclature... right now I am doing these "complex" templates by writing the pages directly in html, but of course you loose the flexibility and cleanness of Markdown.

Would something like this be worth doing?

pxquim:
Hi @justinmayer. I am using the Series plugin as well. For example, if I go to an event with three shows, each show gets its own article, possibly with different tags.

However, I also had karate competitions where I publish a thousand photos in several galleries, often 10-20. I attach each gallery to its own article, but I have very little to say about each one. And these little articles would pollute the blog indexes, hiding older content.

My sub-article solution allows me to have a single entry in the blog index (avoiding the pollution). The article ends with (template-generated) links to the sub-parts. Sub-articles have their own tags (and I have a tag for all the "karate medal" galleries, for example).

So, sub-articles are valuable to me, and I could not figure out a way to achieve the same with a plugin.

Of course, like @ingwinlu says, being useful for me does not mean that it should be a part of core Pelican.

justinmayer:
@pxquim / @goanpeca: You both raise some good points, and I can see how the use cases you describe would benefit from the functionality provided here. As @ingwinlu suggested, perhaps this might be best suited as an optional plug-in.

@pxquim: Would you be so kind as to reformulate this as a plug-in and resubmit it in the pelican-plugins repository?

justinmayer:
This functionality has been moved to the `sub-parts` plugin, as noted above.

pxquim:
@goanpeca, about pages with parts:

Note that reStructuredText has the notion of [Sidebar](http://docutils.sourceforge.net/docs/ref/rst/directives.html#sidebar), which is like a separate mini-document with its own title and name. Perhaps you could extract the sidebars for separate processing, either in Pelican (with a plugin) or at the browser (with Javascript)?

PR #1622, opened 2015-02-10
Format custom metadata fields.

patrickfournier:
This patch allows the use of reStructuredText or Markdown content in metadata fields other than "summary".

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -214,6 +214,7 @@ Setting name (followed by default value, if any)                                
                                                                                  in this list are written. Paths should be either absolute or relative
                                                                                  to the current Pelican working directory. For possible use cases see
                                                                                  :ref:`writing_only_selected_content`.
+``FORMATTED_FIELDS = ['summary']``                                               A list of metadata fields to be parsed and translated to HTML.
 ===============================================================================  =====================================================================
 
 .. [#] Default is the system locale.
--- a/pelican/readers.py
+++ b/pelican/readers.py
@@ -136,13 +136,15 @@ class RstReader(BaseReader):
 
     def _parse_metadata(self, document):
         """Return the dict containing document metadata"""
+        formatted_fields = self.settings['FORMATTED_FIELDS']
+
         output = {}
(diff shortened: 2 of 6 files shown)


docs/settings.rst
```diff
@@ -78,6 +78,7 @@ Setting name (followed by default value, if any)
                                                                                  date and the slug, you could set something like:
                                                                                  ``'(?P<date>\d{4}-\d{2}-\d{2})_(?P<slug>.*)'``.
                                                                                  See :ref:`path_metadata` and ``SLUGIFY_SOURCE``.
+``FORMATTED_FIELDS = []``                                                        A list of metadata fields to be parsed and translated to HTML.
```
kernc: This setting (with specified default value) should be specified in settings.py also.

pelican/readers.py
```diff
@@ -136,13 +136,19 @@ def __init__(self, *args, **kwargs):
 
     def _parse_metadata(self, document):
         """Return the dict containing document metadata"""
+        if 'FORMATTED_FIELDS' in self.settings:
```
kernc: So if you put this setting in settings.py as well, you can just rely on it being there. Additionally, I wouldn't hardcode 'summary' being always parsed. Perhaps make it a default?

kernc:
You would definitely need to [squash/fixup the commits](https://www.google.com/search?q=github+squash+commits) into one meaningful one.

patrickfournier:
I added FORMATTED_FIELDS in the default settings with ['summary'] as the default value, and squashed everything.

kernc:
Looks much better to me. Hope it gets accepted. :+1:

avaris:
Since this only applies to reStructuredText, documentation should mention that.

patrickfournier:
It applies to Markdown too.

avaris:
Oh, right. Sorry, apparently I'm blind (or I need some sleep).

justinmayer:
@kernc & @avaris: Thank you for your review and feedback. Much appreciated!

@patrickfournier: Thanks for the enhancement!

PR #1607, opened 2015-02-01
Import drafts from WordPress XML

arulrajnet:
The fix for accepting draft post while importing from wordpress xml file. The correct meta also added in the generated `.rst` or `.md` file.

The meta for `draft` is look like

```
Build environment for flex with eclipse
#######################################
:date: 2010-10-03 04:16
:author: admin
:category: Uncategorized
:slug: build-environment-for-flex-with-eclipse
:status: draft
```

The meta for `published` is look like

```
About
#####
:date: 2009-04-03 13:22
:author: admin
:slug: about
:status: published
```

Currently its only works with wordpress. Other import also will not break because of `status=None` in line no 452

PR #1623, opened 2015-02-11
Fix for tests that were broken in #1607

avaris:


justinmayer:
A thousand thanks, Deniz!

PR #1625, opened 2015-02-12
Provides Custom URLs for FEEDS

linickx:
Provides an optional FEED URL. Feed variables have been split into two, _SAVE_AS and _URL.

 _SAVE_AS variables are used to define the RSS/ATOM Files.
 _URL variables are the feed slugs (aligns to post/page variables)

 Using rewrite rules (nginx/apache) it is now possible to have _pretty_ feed URLS like /feed or /feed/tag/pelican

I have uploaded [my publishconf.py](https://github.com/linickx/dotcom/blob/master/publishconf.py) as an example config. ( _The default pelicanconf.py uses only the _SAVE_AS URLs as that is all thats explicitly necessary_)

Themes will need updating, my themes [base.html](https://github.com/linickx/dotcom/blob/master/themes/linickx/templates/base.html) includes [head.html](https://github.com/linickx/dotcom/blob/master/themes/linickx/templates/head.html) to show how link formatting can be used.

[This gist](https://gist.github.com/linickx/3c319b72e444603f298e) is the first 5 lines of my sites [/tag/python/feed](http://www.linickx.com/tag/python/feed) which shows the links matching the feed URL. The feed URL works using an [nginx rewrite](https://github.com/linickx/dotcom/blob/master/nginx/inc.d/linickx.conf)

:warning: This pull-request depreciates the old FEED variables therefore custom and community themes will need updating. (_The stock themes have been updated to pass unit tests_)

:notebook: The _URL variables are optional, the _SAVE_AS variables are used in place of the old variables of those who are happy with .xml feed urls

Documentation & Unit Tests have been updated, please let me know if this pull-request is missing anything that could stop it from being accepted, I'm happy to make changes suggested by the lead developers.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -154,11 +154,11 @@ To disable feed generation, all feed settings should be set to ``None``.
 All but three feed settings already default to ``None``, so if you want to
 disable all feed generation, you only need to specify the following settings::
 
-    FEED_ALL_ATOM = None
-    CATEGORY_FEED_ATOM = None
-    TRANSLATION_FEED_ATOM = None
-    AUTHOR_FEED_ATOM = None
-    AUTHOR_FEED_RSS = None
+    FEED_ALL_ATOM_SAVE_AS = None
+    CATEGORY_FEED_ATOM_SAVE_AS = None
+    TRANSLATION_FEED_ATOM_SAVE_AS = None
+    AUTHOR_FEED_ATOM_SAVE_AS = None
+    AUTHOR_FEED_RSS_SAVE_AS = None
 
 The word ``None`` should not be surrounded by quotes. Please note that ``None``
 and ``''`` are not the same thing. 
@@ -191,6 +191,24 @@ generates ``'feeds/all.atom.xml'`` by default and ``FEED_ATOM`` now defaults to
 ``None``. The following feed setting has also been renamed::
 
     TRANSLATION_FEED -> TRANSLATION_FEED_ATOM
+
+Starting in 3.5, the feed variables have been split into two. FEED_ATOM
+has now become FEED_ATOM_SAVE_AS allowing FEED_ATOM_URL to be introduced; FEED_ATOM_URL
+allows you to assign URL Slugs to feeds such as /feed/atom. (URL rewriting will still
+be required on the web server). Here is an exact list of the renamed settings::
+
+    FEED_ATOM -> FEED_ATOM_SAVE_AS
+    FEED_RSS -> FEED_RSS_SAVE_AS
+    FEED_ALL_ATOM -> FEED_ALL_ATOM_SAVE_AS
+    FEED_ALL_RSS -> FEED_ALL_RSS_SAVE_AS
+    CATEGORY_FEED_ATOM -> CATEGORY_FEED_ATOM_SAVE_AS
+    CATEGORY_FEED_RSS -> CATEGORY_FEED_RSS_SAVE_AS
+    AUTHOR_FEED_ATOM -> AUTHOR_FEED_ATOM_SAVE_AS
+    AUTHOR_FEED_RSS -> AUTHOR_FEED_RSS_SAVE_AS
+    TAG_FEED_ATOM -> TAG_FEED_ATOM_SAVE_AS
+    TAG_FEED_RSS -> TAG_FEED_RSS_SAVE_AS
+
+To use the new URLs see :doc:`feed settings <settings>`.
 
 Older themes that referenced the old setting names may not link properly.
 In order to rectify this, please update your theme for compatibility by changing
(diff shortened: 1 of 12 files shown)


justinmayer:
Thanks for the detailed pull request. I agree that "pretty" feed URLs are a worthwhile goal, but I believe that they can already be achieved in the current version of Pelican. For example, my production settings file contains:

    FEED_ATOM = "feeds/main"

Voila — pretty URLs, without any web server rewrite rules. I can't think of a reason to save your feeds with crufty `.xml` extensions if you don't want them as part of the feed link URLs. Ergo, it's not clear to me what the advantage would be to bifurcate this setting into `_URL` and `_SAVE_AS` components.

linickx:
Hi. The advantage would be exactly the same as having 'ARTICLE_URL' / 'ARTICLE_SAVE_AS'.

The use case I'm trying to achieve was a nested feed structure, the following currently fails:

```
   FEED_RSS = "feed"
   FEED_ALL_RSS = "feed/all"
   AUTHOR_FEED_RSS = "feed/author/%s"
   TAG_FEED_RSS = "feed/tag/%s"
   FEED_ATOM = "feed/atom"
   FEED_ALL_ATOM = "feed/atom/all"
   AUTHOR_FEED_ATOM = "feed/atom/author/%s"
   TAG_FEED_ATOM = "feed/atom/tag/%s"
```

One option is just to _bodge_ it with rewrite but then the `<link>` tag in the xml is wrong which can cause issues with feed readers as the subscribed (_user configured_) URL does not match the `_self` found in the feed.... hence I opt'ed for aligning the `FEED` variables to the `ARTICLE` & `PAGE` variables.

justinmayer:
Without further detail, I don't know what you mean by "fails". That said, it seems to me that `FEED_RSS = "feed"` could be a problem, since that will create a file called "feed" in the same place that your subsequent directives are trying to create a directory. That collision could be causing the failure you mentioned.

If you have your heart set on your primary RSS feed being named "feed", then I suggest that all of the subsequent directives be changed so that the directory that they are trying to create is instead named "feeds/[...]".

linickx:
That's exactly the issue, the current feed implementation is limited by file / directory requirements. 

In the above example it's not just that `/feed` creates a block, but also having `/feed/atom` blocks creating `/feed/atom/all`

linickx:
Hi - I created [a gist with example failure output and example pelicanconf.py](https://gist.github.com/linickx/1b5a826bfbd6972bba9b)  (_I ran it against the samples_)

avaris:
So, you need a bunch of `FEED_*_URL`s to use in your templates? Well, you can already do that. Pelican lets you define arbitrary settings to use in your templates. e.g. you can do something like:

```python
FEED_RSS = 'feed.xml'
FEED_RSS_URL = 'feed'
FEED_ALL_RSS = 'feed/all.xml'
FEED_ALL_RSS_URL = 'feed/all'
# ...
```

and in the template

```jinja
<link href="{{ FEED_DOMAIN }}/{{ FEED_RSS_URL }}" ... />
<link href="{{ FEED_DOMAIN }}/{{ FEED_ALL_RSS_URL }}" ... />

<!--- or with a fall back to default if *_URL variables aren't defined -->
<link href="{{ FEED_DOMAIN }}/{{ FEED_RSS_URL or FEED_RSS }}" ... />
<link href="{{ FEED_DOMAIN }}/{{ FEED_ALL_RSS_URL or FEED_ALL_RSS }}" ... />
```

Or, if you don't mind a little performance loss (i.e. doing these replacements for every render), you can strip `.xml` from non-URL settings and negate the need to define `*_URL` settings.

```python
import re
def strip_xml(text):
    return re.sub('\.xml$', '', text)

JINJA_FILTERS = {'strip_xml': strip_xml}
```

and

```jinja
<link href="{{ FEED_DOMAIN }}/{{ FEED_RSS|strip_xml }}" ... />
<link href="{{ FEED_DOMAIN }}/{{ FEED_ALL_RSS|strip_xml }}" ... />
```

linickx:
Hello,

> So, you need a bunch of FEED_*_URLs to use in your templates?

No. I see what you're saying but .... The mail goal of this PR is to ensure that the XML contained the correct _self_ link

    <atom:link href="http://www.linickx.com/tag/python/feed" rel="self">

^that tag was taken from [this gist](https://gist.github.com/linickx/3c319b72e444603f298e) which was generated using my patched version of pelican. 

The advantage of maintaining the URL inside the XML feed is that my feeds are updating correctly with 3rd party services after my transition from [wp](http://Wordpress.org) . Without this as soon as my new pelican site went live, feed services (like ifttt) would have thought there were 10 new articles in the XML (because the src of the articles has changed).

justinmayer:

justinmayer:
Hi Nick. I really appreciate the contribution, and I'm very sorry for the long delay in coming back to your pull request. I understand the intention behind this effort, but for now, I get the impression it is an edge case that does not benefit enough people to warrant the implementation's maintenance cost.

If someone comes across this PR and feels strongly enough about this feature to be willing to modernize it, please comment here and we can discuss further.

Issue #1541, opened 2014-11-20
double quotes not always picked up by TYPOGRIFY in ReST documents


howthebodyworks:
In my blog, double quotes never get the Smartypants treatment despite TYPOGRIFY=True in my settings file. Not sure if this is code, documentation or PEBCAK error. First, let's see if I'm crazy.

`quote_test_1.rst` contains the following:
````
:title: quote test 1
:date: 2014-11-20 13:27:54 +0100

'This' is some "content" -- so --- I --- claim.
````
The resulting page on my blog is
![screen shot 2014-11-20 at 13 40 08](https://cloud.githubusercontent.com/assets/21740/5124688/e20b9426-70ba-11e4-901f-07167dec42dd.png)
Single quotes and dashes are "smartified" but double quotes are not.

On the other hand, `quote_test_2.md` contains the following:
````
Title: quote test 2
Date: 2014-11-20 13:30:32 +0100

'This' is some "content" -- so --- I --- claim.
````
![screen shot 2014-11-20 at 13 32 44](https://cloud.githubusercontent.com/assets/21740/5124613/010a6cfe-70ba-11e4-8fa7-114cd1711cd7.png)
Single quotes, dashes *and* double quotes are "smartified".

AFAICT the settings to choose which smartypants transformations to apply are *not* exposed, so this shouldn't really be happening.

Running current pelican master, which corresponds to the 3.5.0 release version.

* smartypants==1.8.6
* typogrify==2.0.7
* python==2.7.5

Unit tests pass, although I couldn't work out how to tell if the typogrify filter test (which does doest for double quote substitution in ReST) was being skipped. I'll return later to work that out.
````
$ python -m unittest discover
.....................................................................ssssssssssssssssss............................................................
----------------------------------------------------------------------
Ran 147 tests in 5.900s

OK (skipped=18)
````

Is anyone else experiencing this issue?

-Dan

johnmastro:
I also noticed this. As far as I can tell, the issue is that `docutils` has already replaced the double quotes with `&quot;` by the time `typogrify` and `smartypants` see them.

However, I don't have enough knowledge of the various pieces to say whether that's a mistake or if there's an easy fix for it. Perhaps `smartypants` should optionally perform the replacement on the `&quote;` entities, or maybe there's a way to prevent `docutils` from replacing the double quotes at the beginning.

justinmayer:
Good sleuthing. Anyone want to tackle this and come up with a fix?

[...]

PR #1627, opened 2015-02-13
Tell smartypants to also process &quot; entities

johnmastro:
This is necessary because Docutils has already replaced double quotes with `&quot;` HTML entities by the time the `typogrify` filter is applied.

This topic has recently been discussed in issue #1541.


justinmayer:
Thanks for the enhancement, John!

PR #1632, opened 2015-02-19
ArticlesGenerator: Set blog=True consistently. Fixes #1631.

### LINKED ISSUE #1631: ArticlesGenerator sets blog template variable inconsistently
Some of the ArticlesGenerator methods set `blog=True` when rendering templates, which could be very helpful for theme developers wanting to selectively include/exclude certain elements in blog templates vs. page templates. Unfortunately, the variable is set inconsistently. For example, it is present in the list of articles for a time period and in the list of authors, but not the list of articles for any one author. As far as I can tell, this makes it rather useless.

foresto:
Some of the `ArticlesGenerator` methods set `blog=True` when rendering templates, which could be very helpful for theme developers wanting to selectively include/exclude certain elements in blog templates vs. page templates. Unfortunately, the variable is set inconsistently. For example, it is present in the list of articles for a time period and in the list of authors, but not the list of articles for any one author. As far as I can tell, this makes it rather useless. This patch corrects the inconsistency.

### DIFF
--- a/pelican/generators.py
+++ b/pelican/generators.py
@@ -350,7 +350,8 @@ class ArticlesGenerator(CachingGenerator):
             signals.article_generator_write_article.send(self, content=article)
             write(article.save_as, self.get_template(article.template),
                   self.context, article=article, category=article.category,
-                  override_output=hasattr(article, 'override_save_as'))
+                  override_output=hasattr(article, 'override_save_as'),
+                  blog=True)
 
     def generate_period_archives(self, write):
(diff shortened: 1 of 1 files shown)


iKevinY:
Looks good to me. It's nice to have more information to work with when it comes to writing themes.

justinmayer:
@iKevinY: Thanks for the review and input!

@foresto: Thank you for the enhancement. Much appreciated.

PR #1635, opened 2015-02-22
remove useless if-condition in index template

SkyLothar:
useless if-condition on articles_page's attribute.

in this case, if loop.first is `True`, articles_page.has_previous() is always `True`


iKevinY:
Looks good to me. For the sake of being explicit, this redundancy is due to the `if` condition in [line 8](https://github.com/getpelican/pelican/blob/master/pelican/themes/notmyidea/templates/index.html#L8).

justinmayer:
Thanks to @SkyLothar for the fix and to @iKevinY for the review!

PR #1636, opened 2015-02-23
Add support for YAML metadata in Markdown files

kernc:
Meta extension from Python-Markdown>=2.6 supports YAML (Jekyll-like) headers and an optional `yaml` switch which, when used, parses data with PyYAML and hence a wee bit different metadata object gets provided. This is what I came up with. I think it may work well perhaps.

Related: https://github.com/getpelican/pelican-plugins/pull/382

~~Would YAML deserve a mention in _docs/content.rst_?~~Done.

Would additional test cases be desired? :sweat_smile:

### DIFF
--- a/pelican/readers.py
+++ b/pelican/readers.py
@@ -30,7 +30,9 @@ from pelican.utils import get_date, pelican_open, FileStampDataCacher, SafeDatet
 
 def strip_split(text, sep=','):
     """Return a list of stripped, non-empty substrings, delimited by sep."""
-    items = [x.strip() for x in text.split(sep)]
+    if not isinstance(text, list):
+        text = text.split(sep)
+    items = [x.strip() for x in text]
     return [x for x in items if x]
 
 
@@ -90,6 +92,8 @@ class BaseReader(object):
 
     def process_metadata(self, name, value):
         if name in METADATA_PROCESSORS:
+            if isinstance(value, list) and len(value) == 1:
+                value = value[0]
             return METADATA_PROCESSORS[name](value, self.settings)
         return value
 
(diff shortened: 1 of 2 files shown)


kernc:
The discussion on [dropping internal metadata altogether](https://github.com/waylan/Python-Markdown/issues/390) should be considered. Some of the changes here are relevant even with an [external yaml-meta-data markdown plugin](https://github.com/teoric/python-markdown-yaml-meta-data), though.

robclancy:
Just add it as an option? Why is this on the milestone with no activity? We want the release but all I see is 3 random issues stagnating. Is there other issues actually holding up the release?

kernc:
@robclancy How would you make it an option? Everything is already in place as-is anyway, just a couple of invariants need to be evened out (e.g. if it's already a datetime object, don't parse it, if it's already a list of tags, don't split it, etc.)

Will rebase this asap.

robclancy:
I don't know I just see this stagnating for no reason other than not being able to decide to put it in or not? I don't know if I will use this but as I come from other static site generators I would like the option (we have a LOT of metadata atm). I just want to be able to go to 3.6 and stay there instead of checking back every week for a new release.

justinmayer:
Thanks for the rebase, @kernc. From your perspective, I assume this is good to go and ready for inclusion in 3.6, without any follow-on work needed before release?

ingwinlu:
I would expect to see some test cases for this kind of metadata

justinmayer:
@kernc: How useful/relevant is this PR given that Python-Markdown [removed YAML metadata support](https://github.com/waylan/Python-Markdown/commit/4f9d4fff86c638ab2edb569ad9eb14f82728808d)?

kernc:
Indeed. And the YAML meta data extension seems to have just [recently reverted to outputting strings](https://github.com/teoric/python-markdown-yaml-meta-data/commit/305ba3eca9c85e44c4352b6504fdebde5d9803ff). So this PR seems to be irrelevant. Thanks for checking again. :relieved:

PR #1637, opened 2015-02-23
rsync -avc, not --avc

pkirk:
As in the title, rsync wants -avc, not --avc. Thanks for your time.


justinmayer:
Thanks for the fix, @pkirk!

PR #1638, opened 2015-02-23
Avoid Markdown 2.6 deprecations

kernc:
* Short extension names ('extra', 'meta') are deprecated  
  https://pythonhosted.org/Markdown/release-2.6.html#shortened-extension-names-deprecated
* Extension config as part of extension name is deprecated  
  https://pythonhosted.org/Markdown/release-2.6.html#extension-configuration-as-part-of-extension-name-deprecated

The `MD_EXTENSIONS` documentation in `docs/settings.rst` hasn't been touched, so it still shows the old value.

### DIFF
--- a/docs/internals.rst
+++ b/docs/internals.rst
@@ -50,7 +50,10 @@ Take a look at the Markdown reader::
         def read(self, source_path):
             """Parse content and metadata of markdown files"""
             text = pelican_open(source_path)
-            md = Markdown(extensions = ['meta', 'codehilite'])
+            md_extensions = {'markdown.extensions.meta': {},
+                             'markdown.extensions.codehilite': {}}
+            md = Markdown(extensions=md_extensions.keys(),
+                          extension_configs=md_extensions)
             content = md.convert(text)
 
             metadata = {}
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -112,15 +112,15 @@ Setting name (followed by default value, if any)                                
                                                                                  of these patterns will be ignored by the processor. For example,
                                                                                  the default ``['.#*']`` will ignore emacs lock files, and
                                                                                  ``['__pycache__']`` would ignore Python 3's bytecode caches.
-``MD_EXTENSIONS =`` ``['codehilite(css_class=highlight)','extra']``              A list of the extensions that the Markdown processor
-                                                                                 will use. Refer to the Python Markdown documentation's
+``MD_EXTENSIONS =`` ``{...}``                                                    A dict of the extensions that the Markdown processor
+                                                                                 will use, with extensions' settings as the values.
+                                                                                 Refer to the Python Markdown documentation's
                                                                                  `Extensions section <http://pythonhosted.org/Markdown/extensions/>`_
-                                                                                 for a complete list of supported extensions. (Note that
-                                                                                 defining this in your settings file will override and
-                                                                                 replace the default values. If your goal is to *add*
-                                                                                 to the default values for this setting, you'll need to
-                                                                                 include them explicitly and enumerate the full list of
-                                                                                 desired Markdown extensions.)
+                                                                                 for a complete list of supported extensions and their options.
+                                                                                 Default is ``{'markdown.extensions.codehilite' : {'css_class': 'highlight'},
+                                                                                 'markdown.extensions.extra': {}, 'markdown.extensions.meta': {}}``.
+                                                                                 (Note that the dictionary defined in your settings file will
+                                                                                 update this default one.)
 ``OUTPUT_PATH = 'output/'``                                                      Where to output the generated files.
 ``PATH``                                                                         Path to content directory to be processed by Pelican. If undefined,
                                                                                  and content path is not specified via an argument to the ``pelican``
--- a/pelican/readers.py
+++ b/pelican/readers.py
@@ -246,9 +246,8 @@ class MarkdownReader(BaseReader):
 
     def __init__(self, *args, **kwargs):
         super(MarkdownReader, self).__init__(*args, **kwargs)
-        self.extensions = list(self.settings['MD_EXTENSIONS'])
-        if 'meta' not in self.extensions:
-            self.extensions.append('meta')
+        self.extensions = self.settings['MD_EXTENSIONS']
+        self.extensions.setdefault('markdown.extensions.meta', {})
         self._source_path = None
 
     def _parse_metadata(self, meta):
@@ -284,7 +283,8 @@ class MarkdownReader(BaseReader):
         """Parse content and metadata of markdown files"""
 
         self._source_path = source_path
-        self._md = Markdown(extensions=self.extensions)
+        self._md = Markdown(extensions=self.extensions.keys(),
+                            extension_configs=self.extensions)
         with pelican_open(source_path) as text:
             content = self._md.convert(text)
 
--- a/pelican/settings.py
+++ b/pelican/settings.py
@@ -101,7 +101,11 @@ DEFAULT_CONFIG = {
     'PELICAN_CLASS': 'pelican.Pelican',
     'DEFAULT_DATE_FORMAT': '%a %d %B %Y',
     'DATE_FORMATS': {},
-    'MD_EXTENSIONS': ['codehilite(css_class=highlight)', 'extra'],
+    'MD_EXTENSIONS': {
+        'markdown.extensions.codehilite': {'css_class': 'highlight'},
+        'markdown.extensions.extra': {},
+        'markdown.extensions.meta': {},
+    },
     'JINJA_EXTENSIONS': [],
     'JINJA_FILTERS': {},
     'LOG_FILTER': [],
@@ -361,6 +365,14 @@ def configure_settings(settings):
                            "(must be a list), falling back to the default",
                            PATH_KEY)
             settings[PATH_KEY] = DEFAULT_CONFIG[PATH_KEY]
+
+    # Save people from declaring MD_EXTENSIONS as a list rather than a dict
+    if not isinstance(settings.get('MD_EXTENSIONS', {}), dict):
+        logger.warning('The format of the MD_EXTENSIONS setting has '
+                       'changed. It should now be a dict mapping '
+                       'fully-qualified extension names to their '
+                       'configurations. Falling back to the default.')
+        settings['MD_EXTENSIONS'] = DEFAULT_CONFIG['MD_EXTENSIONS']
 
     # Add {PAGE,ARTICLE}_PATHS to {ARTICLE,PAGE}_EXCLUDES
     mutually_exclusive = ('ARTICLE', 'PAGE')
--- a/pelican/tests/test_readers.py
+++ b/pelican/tests/test_readers.py
@@ -471,7 +471,12 @@ class MdReaderTest(ReaderTest):
         # expected
         page = self.read_file(
             path='article_with_markdown_markup_extensions.md',
-            MD_EXTENSIONS=['toc', 'codehilite', 'extra'])
+            MD_EXTENSIONS={
+                'markdown.extensions.toc': {},
+                'markdown.extensions.codehilite': {},
+                'markdown.extensions.extra': {}
+            }
+        )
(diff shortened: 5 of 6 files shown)


pelican/readers.py
```diff
@@ -233,7 +233,13 @@ def read(self, source_path):
...
 
         self._source_path = source_path
-        self._md = Markdown(extensions=self.extensions)
+        extension_configs = {}
+        if 'markdown.extensions.codehilite' in self.extensions:
+            extension_configs['markdown.extensions.codehilite'] = {
+                'css_class':'highlight'
+            }
+        self._md = Markdown(extensions=self.extensions,
+                            extension_configs=extension_configs)
```
kernc: I don't think this will surprise too many unweary users. If they want some other `css_class`, they will have had something else in `MD_EXTENSIONS` anyway.

ingwinlu:
Hey,

gonna pile on this: wasn't headerid part of extra before it now got `deprecated`? I don't think we run tests on that currently, but the behaviour of markdownreader will change through an update (i.e. header ids will not be available anymore)

About the config thing, Can we not just blindly overwrite or correct settings? Would be better to use our deprecation functions somehow and warn the user that we are changing his settings to the now correct ones.

ingwinlu:
Appearently I was wrong and extra never included headerid

kernc:
What about a code like this:

**_settings.py_**
```python
'MD_EXTENSIONS': [
    'codehilite-INTERNAL-ONLY',  # something that would never have been entered by a user
    'markdown.extensions.extra'
],
```

**_readers.py_**
```python
if 'codehilite-INTERNAL-ONLY' in self.extensions:
    ... # correct the plugin name
    ... # set the default css_class setting as above
```
I wouldn't know how to use the deprecation machinery here, nor how it applies.

avaris:
It's probably better to expose `extension_configs` as a setting since any string extension has to be configured with that. It'll also make setting YAML meta easier:

``` python
'MD_EXTENSIONS': [
    'markdown.extensions.codehilite'
    'markdown.extensions.extra'
],
'MD_EXTENSION_CONFIGS' : {'markdown.extensions.codehilite' : {'css_class': 'highlight' }},
```

and the user defined config will `.update` the default one, so people can do:

```python
MD_EXTENSION_CONFIGS = {'markdown.extensions.meta': {'yaml': True}}
```

to turn on YAML processing.

Besides that, `settings.py` has to correct old `MD_EXTENSIONS` settings defined by the user (and also warn them) to the new one. If correcting becomes too tricky, we can just warn that the syntax is changed and fall back to default values.

kernc:
So then when Markdown decides to break its API with 3.0, wrap everything once again to provide backwards compatibility? I'm against it. The `MD_EXTENSIONS` interface is clear: If you need special config parameters, you can instantiate the plugin class directly as you want it.

docs/settings.rst
```diff
@@ -39,184 +39,186 @@ Here is a list of settings for Pelican:
 Basic settings
 ==============
 
-===============================================================================  =====================================================================
-Setting name (followed by default value, if any)                                 What does it do?
-===============================================================================  =====================================================================
```
kernc: `git diff --ignore-space-change` shows the true changes here.

pelican/readers.py
```diff
@@ -200,6 +200,16 @@ class MarkdownReader(BaseReader):
...
+        self.extension_configs = self.settings['MD_EXTENSION_CONFIGS']
+        for extension in self.extensions:
+            try: name, params = extension.split('(')
+            # Else, the extension is a str without config parameters,
+            # or an extension object instance
+            except (ValueError. AttributeError): continue
+            logger.warning('Configuration of Markdown extensions in '
+                           'MD_EXTENSIONS is deprecated. Use '
+                           'MD_EXTENSION_CONFIGS instead.')
+            self.extension_configs[name.strip()] = eval('dict(' + params)
```
kernc: Hopefully no one will find this evil.

kernc:
How about now? :sweat_smile:

justinmayer:
@avaris: Thoughts on @kernc's latest?

avaris:
@justinmayer Ah, sorry, I somehow missed the last commit.

@kernc Looks mostly OK. Use of `eval` is a bit troublesome. I can see why it was used but still it would be best to avoid.

ingwinlu:
Am I the only one who is against introducing setting deprectiation in readers.py?

ingwinlu:
@kernc any update on this? Can you move the depreciation into the [appropriate section](https://github.com/getpelican/pelican/blob/master/pelican/__init__.py#L89)?

@justinmayer might want to tag this to keep track of it for soonish release

justinmayer:
@ingwinlu: Tagged. Thanks!

kernc:
Have considered both @avaris' and @ingwinlu's recommendations. Particularly the provided link was really helpful to a project novice like myself. That `dict` `eval`-ing is kind of hackish, but good enough and should work in all practical cases, I think.

ingwinlu:
hey @kernc good to see you are still on board with this.

some food for thought:
* can we reduce the configs and replace them with a single dict? we can iterate over that dict's keys to get the names identifying the extensions
* can we somehow get the eval out? either replace it with a regex or abort when the config is not correct and let users handle it manually

there also appears quite some ws in this commit, is that really necessary (in the doc)?

kernc:
I agree with your considerations, will implement asap.

Re table whitespace: the default value string should get pretty long. Should I put the default value in the right-hand column instead?

justinmayer:
@kernc: I think putting the default value in the right-hand column for this particular setting might indeed make sense, obviating the need to adjust the spacing for the entire table.

justinmayer:
Hi @kernc. Just following up on this PR. Any chance you might be able to get it ready for merge as noted above?

kernc:
@justinmayer sorry, thanks. :flushed: 

@ingwinlu How would I ["reduce the configs and replace them with a single dict"](https://github.com/getpelican/pelican/pull/1638#issuecomment-132809863)? Isn't that exactly the opposite of what @avaris [recommended](https://github.com/getpelican/pelican/pull/1638#issuecomment-75896410)? 

I also did (almost) drop the eval? Is it better now or is it still _unsafe_?

ingwinlu:
@kernc yes it is different, but I still maintain that pelican has to many options, and there is no value in duplicating settings across two settings that contain the same data:
```
    'MD_EXTENSIONS': [
        'markdown.extensions.codehilite',
        'markdown.extensions.extra',
    ],
    'MD_EXTENSION_CONFIGS': {
        'markdown.extensions.codehilite' : {'css_class': 'highlight'},
    },
```
to ->
```
    'MD_EXTENSIONS': {
        'markdown.extensions.codehilite' : {'css_class': 'highlight'},
        'markdown.extensions.extra' : {},
    },
```

can then be used like
```
self._md = Markdown(extensions=self.extensions.keys(),
                                   extension_configs=self.extension_configs)
```

tada, one less setting, and less points of failures (mismatching keys, how to handle partially configured settings, ...)

kernc:
> if you want to do settings conversion

I don't. What's the alternative?

ingwinlu:
check if `md_extensions` is a list, print a warning that the setting changed and needs to be updated manually, exit

ingwinlu:
please do not put my handle in your commit message and then push it multiple times, i just got 3 emails because of that in rapid succession...

kernc:
Alas, this is how it works. :stuck_out_tongue: 

Updated, coverage is missing a test for the new "deprecation."

pelican/readers.py
```diff
@@ -246,9 +246,10 @@ class MarkdownReader(BaseReader):
 
     def __init__(self, *args, **kwargs):
         super(MarkdownReader, self).__init__(*args, **kwargs)
-        self.extensions = list(self.settings['MD_EXTENSIONS'])
-        if 'meta' not in self.extensions:
-            self.extensions.append('meta')
+        self.extensions = self.settings['MD_EXTENSIONS']
+        # Ensure 'meta' extension gets loaded
+        self.extensions['markdown.extensions.meta'] = \
```
ingwinlu: a more readable way to do this:
```
self.extensions.setdefault('markdown.extensions.meta', {})
```
```
setdefault(key[, default])¶
If key is in the dictionary, return its value. If not, insert key with a value of default and return default. default defaults to None.
```

pelican/settings.py
```diff
@@ -101,7 +101,10 @@ def load_source(name, path):
     'PELICAN_CLASS': 'pelican.Pelican',
     'DEFAULT_DATE_FORMAT': '%a %d %B %Y',
     'DATE_FORMATS': {},
-    'MD_EXTENSIONS': ['codehilite(css_class=highlight)', 'extra'],
+    'MD_EXTENSIONS': {
+        'markdown.extensions.codehilite': {'css_class': 'highlight'},
+        'markdown.extensions.extra': {},
+    },
```
ingwinlu: might be worth it to add meta to the default settings

kernc: And since the dict is _updated_, does inclusion here obsolete the test in MarkdownReader?

docs/internals.rst
```diff
@@ -50,7 +50,8 @@ Take a look at the Markdown reader::
         def read(self, source_path):
             """Parse content and metadata of markdown files"""
             text = pelican_open(source_path)
-            md = Markdown(extensions = ['meta', 'codehilite'])
```
ingwinlu: needs to be updates as well

pelican/settings.py
ingwinlu: no?

docs/internals.rst
kernc: Thanks for catching that. :sweat_smile:

pelican/settings.py
kernc: Great. :smiley:

avaris:
Why are we hard-quitting after a misconfiguration that doesn't have any major affect? I seriously disagree with that. It should be treated [same](https://github.com/getpelican/pelican/blob/master/pelican/settings.py#L358) as any other misconfiguration: ignore and fall back to default with a warning.

kernc:
That's exactly why I asked, thanks.

pelican/settings.py
```diff
@@ -362,6 +366,14 @@ def configure_settings(settings):
                            PATH_KEY)
             settings[PATH_KEY] = DEFAULT_CONFIG[PATH_KEY]
 
+    # Save people from declaring MD_EXTENSIONS as a list rather than a dict
+    if not isinstance(settings.get('MD_EXTENSIONS', {}), dict):
+        logger.warning('The format of the MD_EXTENSIONS setting has '
+                       'changed. It should now be a dict mapping '
+                       'fully-qualified extension names to their '
+                       'configurations. Falling back to the default')
```
ingwinlu: missing tailing '.'

pelican/tests/test_readers.py
```diff
@@ -471,7 +471,10 @@ def test_article_with_markdown_markup_extension(self):
         # expected
         page = self.read_file(
             path='article_with_markdown_markup_extensions.md',
-            MD_EXTENSIONS=['toc', 'codehilite', 'extra'])
+            MD_EXTENSIONS={
+                'markdown.extensions.toc': {},
+                'markdown.extensions.codehilite': {},
+                'markdown.extensions.extra': {}})
```
ingwinlu: might want to newline the enclosing dict } here for better readability

ingwinlu:
other than the comments in the code, this looks good to me now

kernc:
You're a nitpicker, aren't you. :stuck_out_tongue:
Got rid of it, thanks.

ingwinlu:
I would love not have to have to correct you.

Squash is still missing, after that we are good to go @justinmayer

justinmayer:
@kernc: As Winlu suggested, would you be so kind as to squash your commits? Many thanks for patiently working with us to get this ready for merge.  (^_^)

kernc:
Oh, no problem at all.

justinmayer:
Sincere thanks to @kernc, @ingwinlu, and @avaris for your work on this. Bravo! :tada:

PR #1642, opened 2015-02-26
Remove PDF-related elements from notmyidea

### LINKED ISSUE #1610: notmyidea theme: article pdf link is generated twice
While reading the notmyidea theme's templates, I noticed that the following code appears in both article_infos.html and taglist.html:

    {% if PDF_PROCESSOR %}<a href="{{ SITEURL }}/pdf/{{ article.slug }}.pdf">get the pdf</a>{% endif %}

Since article_infos.html includes taglist.html, I expect this generates a duplicate link for people who use the pdf plugin. The code should probably be removed from taglist.html.

iKevinY:
Closes #1610.

### DIFF
--- a/pelican/themes/notmyidea/templates/article_infos.html
+++ b/pelican/themes/notmyidea/templates/article_infos.html
@@ -16,7 +16,7 @@
                 {% endfor %}
         </address>
         {% endif %}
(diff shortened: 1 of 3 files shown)


justinmayer:
Thanks, Kevin!

(Related to #1602)

PR #1643, opened 2015-02-28
switch to pytest test runner for better testing experience

bubenkoff:
As a part of pytest adoption initiative
http://pytest.org/latest/adopt.html
Introduces a switch to pytest test runner.

[Review on Reviewable](https://reviewable.io/reviews/getpelican/pelican/1643)


theskumar:
+1

bubenkoff:
@ametaireau can you please review?

justinmayer:
I'm a bit swamped at the moment. Anyone have time to review and test this?

ingwinlu:
any reason to not run python3.3 and python3.4 tests under travis just because of the switch to another testing tool?

iKevinY:
Looking at the [build log](https://travis-ci.org/getpelican/pelican/builds/52531496), it seems like tests for 2.7, 3.3, and 3.4 are all being run by Tox under a single Python 2.7 job.

bubenkoff:
We will run on same set of envs
See tox.ini
Its just an entry point python version to be sure on which interpreter tox
is executed

bubenkoff:
@iKevinY, @ingwinlu, @justinmayer 
here is the list of major pytest benefits:

- fully compartible with existing unittest/nose tests, no need to change anything
- simple but powerful plain ```assert  x == y``` statements instead of ```self.assertEquals(y, y)``` with detailed explanation when assert fails
- well maintained plugin ecosystem https://pytest.org/latest/plugins_index/index.html ('killer' plugins are: pytest-xdist (parallelize test run), pytest-cache (run last failed tests), pytest-pep8 (pep8 code checks), pytest-flakes (flake8 checks), pytest-cov (test coverage) and more!) 
- test parametrization (single test function, multiple parameters to run it!)
- fixtures with dependency injection (move your test setup/teardown code out of the actual test function/class)
- Part of the "adopt pytest month" work would be to create any fixtures that the project needs.
- I would expect that making use of pytest features would result in
tests that had less boilerplate and were "cleaner" to read -
basically, more readable, and it will be easier to write new tests.

avaris:
All nice and good... But without refactoring current tests to "pytest way" all this switch would achieve is adding a new dependency.

bubenkoff:
but this is the test dependency, not the installation dependency
is it critical?

On 26 March 2015 at 10:33, Deniz Turgut <notifications@github.com> wrote:

> All nice and good... But without refactoring current tests to "pytest way"
> all this switch would achieve is adding a new dependency.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/getpelican/pelican/pull/1643#issuecomment-86420635>.
>



-- 
Anatoly Bubenkov

ingwinlu:
also, as a sideeffect of letting tox handle the test environments, travis test time doubles from around 5 min to 10, which is not something major but it is not shown in the list that would change, so this pr now does the following:

* add new test dependencies
* slow down testing
* force maintainers to familiarize with pytest
* does not improve existing tests

can you provide examples on why pytest would be better then nose for pelican?

bubenkoff:
i'll reconfigure travis to use multiple builds - that's not a problem, then the test time will be exactly the same, tox allows to pass environments via env variables

bubenkoff:
on last your point: ```does not improve existing tests```
im all to improve tests, just i thiought that it will be too much for one PR.
i can do next PR to refactor tests to use fixtures/parametrization

ingwinlu:
keep in mind that `does not improve exising tests` also means does not add new features that the current testing environment does not have in any form.

nose has support for fixtures and while it does not have fancy 'parametizing' out of the box there is a plugin which would add it. nose also supports test_generators which is more or less the same concept just nicer to read (though that is certainly a personal opinion)

so yes I would like to see examples on what pytest can do better then nose, no need to rewrite all tests for that, just something simple would suffice.

bubenkoff:
@ingwinlu ok, challenge accepted  :)

avaris:
If test refactoring is going to follow, they can (and IMO should) be part of this PR in different commits. No need to divide the work.

justinmayer:
Thanks for the contribution, Anatoly. I like py.test and would be interested to see your continued work on this effort.

I agree with what @avaris said above. Would you be willing to refactor the tests for py.test and push that additional commit so it appears as part of this pull request?

bubenkoff:
sorry guys, will not have time to do it :(

Issue #1065, opened 2013-08-27
Add use case of pelican as general static site generator


diimpp:
It can be done by moving pages to / and articels to  /blog

Here is a example configuration.

```
ARTICLE_DIR = 'blog'
ARTICLE_URL = 'blog/{slug}.html'
ARTICLE_SAVE_AS = 'blog/{slug}.html'
PAGE_DIR = 'pages'
PAGE_URL = '{slug}.html'
PAGE_SAVE_AS = '{slug}.html'
AUTHOR_SAVE_AS = False
CATEGORY_SAVE_AS = 'blog/category/{slug}.html'
CATEGORY_URL = 'blog/category/{slug}.html'
TAG_SAVE_AS = 'blog/tag/{slug}.html'
TAG_URL = 'blog/tag/{slug}.html'
DIRECT_TEMPLATES = (('index', 'blog/tags', 'blog/categories', 'blog/archives', 'blog/index'))
PAGINATED_DIRECT_TEMPLATES = (('index', 'blog/index', ))
```

Details can be found in first comment to this post http://stackful-dev.com/static-site-jinja-pelican-shared-templates.html

foresto:
Pelican really is shaping up into a decent general-purpose site generator. I suspect more people would choose it if features useful for creating/migrating non-blog sites were more obvious to newcomers. For example:

* Pages can be generated in a site's root directory, and articles can live in a
  subdirectory instead of the root (or be left out entirely).
* The generated directory structure need not be flat or simple; it can be as complex
  as any existing website, either by replicating the source directory structure or through
  per-page path overrides.
* Page sources and static files can either share the same source directories or be divided
  into separate ones. Pelican can work with whichever approach works best for the
  author and his existing tools.
* Individual pages can have their file names and URLs overridden to be different from
  the the pattern used by other pages. This is especially useful for special pages
  like directory indexes.
* Generated file names can be pretty much anything. They don't have to be based
  on a slug, or contain dashes, or end in .html. Their suffixes can be whatever an
  existing site uses, including .php, .asp, or nothing at all.

Most of this stuff can be discovered if someone is willing to invest a bunch of time carefully reading the docs and experimenting, but someone who just wants a convenient website generator is more likely to see Pelican as blogger's tool and move on to something else. That's a pity, because Pelican is definitely capable of more if configured with the right combination of settings.

[...]

PR #1645, opened 2015-03-01
Change phrasing and formatting of README

iKevinY:
There was some good discussion in #1065 about Pelican being a great general-purpose static site generator in addition to a solid blog generator. However, it feels like the README currently emphasizes Pelican's position as a blogging platform, which might be leading prospective users to assume that Pelican is only suitable for blogs.

I've made some changes to make the README a little more content-agnostic, and condensed the final sections into a singular, less disjointed "FAQ" section. Here's a [rendered version](https://github.com/iKevinY/pelican/tree/readme-changes) of these changes.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -3,48 +3,45 @@ Pelican |build-status| |coverage-status|
 
 Pelican is a static site generator, written in Python_.
 
-* Write your weblog entries directly with your editor of choice (vim!)
-  in reStructuredText_ or Markdown_
-* Includes a simple CLI tool to (re)generate the weblog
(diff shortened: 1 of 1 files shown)


justinmayer:
Great changes, Kevin. I merged most of them via 87d86d7, mainly demurring on the FAQ section seeing as how we already have a dedicated FAQ.

I also took the liberty of making a few other minor changes via e35ca1d.

Thanks, Kevin!

iKevinY:
Ah, I forgot about the dedicated FAQ — I can see why that could be misleading. Thanks for the merge!

Issue #1542, opened 2014-11-21
AttributeError: 'unicode' object has no attribute 'name'


yysfire:
When running `pelican content -s pelicanconf.py -D -t notmyidea`, I got the following error:

```
DEBUG: Read file vim/vimwiki-syntaxhighlighter.md -> Article
DEBUG: Read file web/how-to-add-google-custom-search-to-github-pages.md -> Article
CRITICAL: 'unicode' object has no attribute 'name'
Traceback (most recent call last):
  File "D:/DevPack/msys64/mingw64/lib/python2.7/runpy.py", line 162, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "D:/DevPack/msys64/mingw64/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "D:/DevPack/msys64/mingw64/bin/pelican.exe/__main__.py", line 9, in <module>
  File "D:/DevPack/msys64/mingw64/lib/python2.7/site-packages/pelican/__init__.py", line 425, in main
    pelican.run()
  File "D:/DevPack/msys64/mingw64/lib/python2.7/site-packages/pelican/__init__.py", line 169, in run
    p.generate_context()
  File "D:/DevPack/msys64/mingw64/lib/python2.7/site-packages/pelican/generators.py", line 550, in generate_context
    if author.name != '':
AttributeError: 'unicode' object has no attribute 'name'
```

More details:

Windows 7
Python 2.7.8
Pelican 3.5.0

I am the only author of my site, so I set `DEFAULT_METADATA = (('author', 'Yysfire'),)` and
 ```
AUTHOR_SAVE_AS = ''
AUTHORS_SAVE_AS = ''
```
in my pelicanconf.py

Issue #1464, opened 2014-09-15
Support string formatted dates for DEFAULT_DATE


Naatan:
Currently the documentation implies that it supports various formats, but does not explain which. The most intuitive one however is not supported.

It would make sense that this option would support the format of the meta tag that it seeks to provide a default value for, as this would be the most intuitive value type.

justinmayer:
Merged via #1959

PR #1653, opened 2015-03-06
Make sure Content uses URLWrappers

### LINKED ISSUE #1547: `category.slug` is not honored when determining URLs for articles
I'm trying to write a plugin which allows me to manually adjust the slug for a category, i.e. the thing which goes into the URL when you put `{category}` in `ARTICLE_URL` and friends.  I have it working, except that `category.slug` is *ignored* by `Content.url_format` (in favor of `slugify(category)`, which uses the *name*, so the generated pathname for the category itself honors the slug setting, but the generated pathnames for articles in the category don't.

It appears to me that this is a straight-up bug in `Content.url_format`, possibly due to its having been written before `URLWrapper` objects were a thing?

avaris:
This cleans up pre-`URLWrapper` code/logic:

 * `Content.url_format` was sluggify-ing `Author` and `Category` whereas it could've used the `.slug` attribute (see #1547). It was unnecessary computation and was causing issues if you wanted to use custom slugs (via plugins) for those. It now uses `.slug`
 * `DEFAULT_METADATA` was able to put un-processed values as metadata (see #1542).  It is now processed like any other metadata before it's used.
 * Some tests were providing plain strings in place of `URLWrapper` objects for metadata like `Author` and `Category`. They are changed to `URLWrapper` objects.

So, this fixes #1547 and #1542. And coincidentally provides a solution for #1464, because you can now do:

    DEFAULT_METADATA = {'date': '2015-03-06'}

### DIFF
--- a/pelican/contents.py
+++ b/pelican/contents.py
@@ -53,7 +53,7 @@ class Content(object):
         self._context = context
         self.translations = []
 
-        local_metadata = dict(settings['DEFAULT_METADATA'])
+        local_metadata = dict()
         local_metadata.update(metadata)
 
         # set metadata as attributes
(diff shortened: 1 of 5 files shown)


justinmayer:
Great work, Deniz. Many thanks!

PR #1663, opened 2015-03-14
Use `--relative-urls` only if it is specified

avaris:
Otherwise, `RELATIVE_URLS` in the config file is ignored and
`RELATIVE_URLS` is set to `False` if `--relative-urls` is not
specified.

Fixes an issue introduced in #1592


iKevinY:
Looks good to me.

justinmayer:
Many thanks, Deniz!

PR #1669, opened 2015-03-22
Support semicolon-separated author/tag lists

zackw:
Idea borrowed from Docutils.  This allows one to write author lists in
lastname,firstname format.  The code change also means that readers with
fancy metadata that can natively represent lists (e.g. Docutils itself,
or MD-Yaml) don't have to merge 'em back together for process_metadata's
sake.

### DIFF
--- a/docs/content.rst
+++ b/docs/content.rst
@@ -35,6 +35,12 @@ this metadata in text files via the following syntax (give your file the
     :authors: Alexis Metaireau, Conan Doyle
     :summary: Short version for index and feeds
 
+Author and tag lists may be semicolon-separated instead, which allows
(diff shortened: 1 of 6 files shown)


ingwinlu:
seems reasonable / useful enough and contains docs + tests; what do others think?

justinmayer:
I agree.

@zackw: Would you care to rebase on current master so we can get this merged?

zackw:
@justinmayer Sorry for the delay; the branch has now been rebased onto 940eb76b7f70b1c9c7f833d5328d44cb19bde406.

justinmayer:
Thanks for the contribution, Zack. Much appreciated!

PR #1670, opened 2015-03-24
Add Org support *in* existing base readers

jo-tham:
I wanted to throw this into the public sphere to get some feedback/thoughts.

I've been interesting in blogging using pelican and [org-mode](http://orgmode.org/). In fact, I [http://jotham-city.com/blog/2015/03/21/adding-yasnippets-snippets/](have) done so. Orgmode is a great tool for writing and I suspect it has many users who would appreciate native support for it in pelican.

I know additional readers are supposed to be implemented as plugins. One such solution has be created for [org documents](https://github.com/getpelican/pelican-plugins/blob/master/org_reader/org_reader.py). It invokes emacs to generate html from org documents. This (subjectively) strikes me as a bit breakage prone and harder to get more people started with pelican and org-mode. In contrast, I think using pandoc would be a bit more robust and accessible. The option to use various markup standards is also available, making it more flexible.  Additionally, it's not an additional reader as implemented here, but a 'helper' to the base readers, if you will. That said, the exact motivation for keeping additional readers out of the main library are not known to me.

What does the pelican dev community think about building in org support to the main library? 
Even if the thoughts are overwhelmingly, "no!"; I have a few thoughts about the implementation I'm using in this PR which would be interesting to get some feedback on.

- what if there were an option for which standard reader to use for org files, i.e. which intermediate format to convert the org files to? rst, markdown. My preference is to avoid converting it to html thinking it may results in some wierd artefacts 
- would additional tests are needed to make this implementation solid?
- ideas for improving this implentation?

Still to do:
- fix pandoc template has to be in pelican site root (i.e. at the level of content/)

Thanks!

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,3 +1,3 @@
 include *.rst
-recursive-include pelican *.html *.css *png *.in *.rst *.md *.mkd *.xml *.py
+recursive-include pelican *.html *.css *png *.in *.rst *.md *.mkd *.xml *.py *.markdown
 include LICENSE THANKS docs/changelog.rst
--- a/pelican/pandoc_templates/md_org_template.markdown
+++ b/pelican/pandoc_templates/md_org_template.markdown
@@ -2,11 +2,10 @@ $if(title)$Title: $title$$endif$
 $for(author)$Author: $author$$endfor$
 $if(date)$Date: $date$$endif$
 $if(category)$Category: $category$$endif$
-$if(modified)$Modified: $modified$$endif$
-$if(slug)$Slug: $slug$$endif$
 $if(summary)$Summary: $summary$$endif$
-$if(modified)$Modified: $modified$$endif$
+$if(slug)$Slug: $slug$$endif$
 $for(tags)$Tags: $tags$$endfor$
+$if(modified)$Modified: $modified$$endif$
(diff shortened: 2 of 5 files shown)


ingwinlu:
>That said, the exact motivation for keeping additional readers out of the main library are not known to me.

Less clutter / stuff to maintain / less dependencies of the base package

>What does the pelican dev community think about building in org support to the main library? 
>Even if the thoughts are overwhelmingly, "no!"; I have a few thoughts about the implementation I'm using in this PR which would be interesting to get some feedback on.

Not much, reasons above, pelican is cluttered as is.

>what if there were an option for which standard reader to use for org files, i.e. which intermediate format to convert the org files to? rst, markdown. My preference is to avoid converting it to html thinking it may results in some wierd artefacts

probably reStructuredText, as markdown should be more limiting

> ideas for improving this implentation?

you are missing docs, updates to the requirements, tests

just create your own plugin and add it to the plugins repo as well, that way everyone who wants to use org can do so, without all the drawbacks of adding more readers

jo-tham:
Cool - will proceed in that manner.

PR closed

jo-tham:
In case anyone looking for org reader stumbles on this
PR for pluings below
getpelican/pelican-plugins#523
can also find the path of the reader in that diff

PR #1679, opened 2015-04-03
Make `pelican.server` importable and use it in `fab serve`

avaris:
`fab serve` and `make devserver` use different HTTP Handlers and as a
result they behave differently. This makes sure `fab serve` also uses
the Handler defined in `pelican.server` in order to get rid of the
inconsistency.

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -61,10 +61,10 @@ Preview your site
 -----------------
 
 Open a new terminal session and run the following commands to switch to your
-``output`` directory and launch Python's built-in web server::
+``output`` directory and launch Pelican's web server::
 
     cd ~/projects/yoursite/output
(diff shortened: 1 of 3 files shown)


avaris:
On a slightly related topic, we might change the [Quickstart Preview](https://github.com/getpelican/pelican/blob/master/docs/quickstart.rst#preview-your-site) to `pelican.server` as well. And the command will be the same regardless of the python version (2vs3).

    python -m pelican.server [port] [server]

I didn't include that into this PR because it's not entirely related but I can, if there is agreement.

justinmayer:
Thanks, Deniz. Seems like a sensible change. Feel free to add that to this PR.

avaris:
@justinmayer, added. I decided to leave out port/server settings for the sake of simplicity.

justinmayer:
Awesome. Thanks, Deniz!

PR #1680, opened 2015-04-06
[DOCS] fix broken internal metadata link

ingwinlu:
the current link at http://docs.getpelican.com/en/latest/settings.html#path-metadata to http://docs.getpelican.com/en/latest/content.html#file-metadata is broken.

this commit removes the anonymous and hard to maintain link and replaces it with a ref to the correct section


justinmayer:
Thanks for the fix!

PR #1681, opened 2015-04-06
maintain test_readers

ingwinlu:
while tinkering with the idea of moving `extra_path_metadata` into a more regexy format i found that test_readers had some errors.

this replaces all the individual metadata tests with a single call to `assertDictHasSubset` which checks if a Dictionary contains a Subset Dictionary.

that makes it really obvious that some tests (`RstReaderTest.test_article_with_filename_metadata`, ...) had wrong metadata tests, i.e. iterating over the content metadata instead of iterating over the expected fields, resulting in the error that the test would succeed if a value was unset, but was expected to have a value.

The commit also fixes those falsely setup tests.

Additionally it also makes the whole `MdReaderTest` class skipped if Markdown is not available instead of skipping each method individually

### DIFF
--- a/pelican/tests/test_readers.py
+++ b/pelican/tests/test_readers.py
@@ -22,6 +22,52 @@ class ReaderTest(unittest.TestCase):
         r = readers.Readers(settings=get_settings(**kwargs))
         return r.read_file(base_path=CONTENT_PATH, path=path)
 
+    def assertDictHasSubset(self, dictionary, subset):
+        for key, value in subset.items():
+            if key in dictionary:
+                real_value = dictionary.get(key)
+                self.assertEqual(
+                    value,
+                    real_value,
+                    str('Expected %r to have value %r, but was %r')
+                        % (key, value, real_value))
+            else:
+                self.fail(
+                    str('Expected %r to have value %r, but was not in Dict')
+                        % (key, value))
+
+class TestAssertDictHasSubset(ReaderTest):
+    def setUp(self):
+        self.dictionary = {
+        'key-a' : 'val-a',
+        'key-b' : 'val-b'}
+
+    def tearDown(self):
+        self.dictionary = None
+
+    def test_subset(self):
+        self.assertDictHasSubset(self.dictionary, {'key-a':'val-a'})
+
+    def test_equal(self):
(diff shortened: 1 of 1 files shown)


pelican/tests/test_readers.py
```diff
@@ -22,6 +22,50 @@ def read_file(self, path, **kwargs):
...
         return r.read_file(base_path=CONTENT_PATH, path=path)
 
+    def assertDictHasSubset(self, dictionary, subset):
+        for key, value in subset.items():
+            if key in dictionary:
+                real_value = dictionary.get(key)
+                self.assertEqual(
+                    value,
+                    real_value,
+                    "Expected %r to have value %r, but was %r" % (key, value, real_value))
```
avaris: This needs to be `str("Expected %r to have value %r, but was %r") % (key, value, real_value))` because of the way `repr` works in py2 and py3. Otherwise, in py2, the output of `repr` will be implicitly converted to unicode to apply formatting, which caused the errors you had previously. With the change you made to `URLWrappers.__repr__`, you just made it sure that the `repr` is ascii-only and implicit conversion (using `ascii` codec) succeeds.

pelican/tests/test_readers.py
```diff
@@ -22,6 +22,50 @@ def read_file(self, path, **kwargs):
...
+            else:
+                self.fail('Expected ' + repr(key) + ' to have value ' +
```
avaris: It'd be better if this also used string formatting like above (with `str` wrapper as well).

avaris:
As for `URLWrappers.__repr__` change, I don't have a real issue with it. Previously it was `utf-8` encoded string in py2, because of the `str(self)` and `@python_2_unicode_compatible` decorator. Now it's ascii.

ingwinlu:
i think I am ok with how it is now, tackled the main issues I had with the tests, managed to resolve the urlwrappers encoding issues I had thanks to @avaris.

One more thing @avaris mentioned is that error output on the tests are weird because they implicitly test the string representations of urlwrappers instead of objects themself.

however this PR is getting quite large so I would rather introduce type checks in another pr.

justinmayer:
Many thanks to @ingwinlu for the excellent work, and thanks to @avaris for the review.  (^_^)

PR #1683, opened 2015-04-08
fix the meta tags added in #1057

alb-i986:
The attribute 'contents' should be 'content'.
See http://www.w3schools.com/tags/att_meta_content.asp

Fixes #1057 .


justinmayer:
Thanks, Alberto!

PR #1687, opened 2015-04-10
[DOCS] fix broken refs

ingwinlu:
/home/winlu/gitrepos/pelican/docs/themes.rst:374: WARNING: undefined label: object-author_cat_tab (if the link has no caption the label must precede a section header)
/home/winlu/gitrepos/pelican/docs/themes.rst:429: WARNING: undefined label: object-author_cat_tab (if the link has no caption the label must precede a section header)

Another case of PR + merge without even trying to build the docs.


justinmayer:
Apologies for the errant merge. Thank you for fixing the issue!

Issue #1688, opened 2015-04-10
[DOCS] PR's breaking docs


ingwinlu:
Hi all,

everytime I build the docs because I want to document something new I notice that past PR's broke some parts of it.

I vote to add the sphinx doc to the travis build and abort it on warnings as I am fed up with people not testing their documentation changes.

References:
https://github.com/getpelican/pelican/pull/1680
https://github.com/getpelican/pelican/pull/1687

Way forward:
* split dev_requirements into dev_requirements and dev_requirements_docs (containing sphinx and the theme for it)
* build the docs on travis
   * either via a separate environment (faster, need to switch to tox commands to setup environments)
   * build the docs in every environment
* make the travis build fail if warnings pop up during the sphinx run

While this will not catch all errors it will at least prevent some of the poor doc PR's.

PR #1690, opened 2015-04-11
Test docs via Travis

ingwinlu:
this is a followup PR to the issue I raised earlier this week #1688

to handle the additional environment we have to utilize tox more than we did until now and let travis only do the main container spawning.

* move build environment into tox
* add new environment installing sphinx and testing for doc errors
* reorganize dependency installs for easier management

TODO:
-   [x] check for build errors in docs
-   [x] ~~check if docs need an update~~ -> no additions needed in my opinion
-   [x] ~~include move to container based travis testing (http://docs.travis-ci.com/user/workers/container-based-infrastructure/)~~ -> would need move of importer to pelican tools repo first

### DIFF
--- a/dev_requirements.txt
+++ b/dev_requirements.txt
@@ -10,6 +10,3 @@ typogrify
 # To perform release
 bumpr==0.2.0
 wheel
-
-# For docs theme
-sphinx_rtd_theme
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,11 @@
 language: python
 python:
   - "2.7"
-  - "3.3"
-  - "3.4"
+env:
+  - TOX_ENV=docs
+  - TOX_ENV=py27
+  - TOX_ENV=py33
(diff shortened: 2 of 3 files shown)


dev_requirements.txt
```diff
@@ -1,5 +1,9 @@
 # Tests
 mock
+nose
+nose-cov
+coveralls
+tox
```
avaris: These should go to `deps` in `tox.ini`.

justinmayer:
Excellent work, Winlu. My apologies for the delay in getting this in. Thanks for the contribution and for being so understanding of my travel schedule.  (^_^)

justinmayer:
@ingwinlu: Oops. Looks like Travis is reporting [test failures](https://travis-ci.org/getpelican/pelican/builds/62428655).

ingwinlu:
very curious, no idea why travis suddenly changed its mind

PR #1691, opened 2015-04-11
add skips for tests relying on dev_requirements modules

ingwinlu:
add skips for tests relying on non dependency modules in 
* test_generators.py
* test_importer.py
* test_rstdirectives.py


justinmayer:
Thanks, @ingwinlu!

Issue #1692, opened 2015-04-11
Static generator should honor `IGNORE_FILES`


zackw:
Files that match an `IGNORE_FILES` glob should not be copied by the static generator.  This is not just cosmetic; emacs lock files (`'.#*'`, included in `IGNORE_FILES` by default) are pseudo-symlinks (that is, symlinks used as an atomic way to store a small amount of data in the file system -- the destination of the symlink never exists) and `shutil.copy2` will crash if it tries to copy them.

avaris:
Static generator does honor `IGNORE_FILES`. If you're getting tracebacks, please add them.

zackw:
@avaris I suppose it would be more precise to say 'the static generator doesn't apply `IGNORE_PATHS` to files nested inside a directory copied by `_copy_paths`', but I don't see why, with my user hat on, I should have to know that those are not treated the same as other static files.

Example traceback:

~~~~
--- Logging error ---
Traceback (most recent call last):
  File "/home/zack/owlfolio/pelican-core/pelican/__init__.py", line 437, in main
    pelican.run()
  File "/home/zack/owlfolio/pelican-core/pelican/__init__.py", line 177, in run
    p.generate_output(writer)
  File "/home/zack/owlfolio/pelican-core/pelican/generators.py", line 720, in generate_output
    os.curdir)
  File "/home/zack/owlfolio/pelican-core/pelican/generators.py", line 689, in _copy_paths
    os.path.join(output_path, destination, final_path))
  File "/home/zack/owlfolio/pelican-core/pelican/utils.py", line 310, in copy
    recurse(source_, destination_)
  File "/home/zack/owlfolio/pelican-core/pelican/utils.py", line 306, in recurse
    shutil.copy2(entry_path, destination)
  File "/home/zack/owlfolio/pelican-venv/lib/python3.4/shutil.py", line 245, in copy2
    copyfile(src, dst, follow_symlinks=follow_symlinks)
  File "/home/zack/owlfolio/pelican-venv/lib/python3.4/shutil.py", line 108, in copyfile
    with open(src, 'rb') as fsrc:
FileNotFoundError: [Errno 2] No such file or directory: '/home/zack/owlfolio/www/style/static/.#main.scss'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib/python3.4/logging/__init__.py", line 978, in emit
    msg = self.format(record)
  File "/usr/lib/python3.4/logging/__init__.py", line 828, in format
[...]

PR #1693, opened 2015-04-11
Honor 'IGNORE_FILES' in StaticGenerator._copy_paths (#1692).

zackw:
generators.StaticGenerator._copy_paths just passes the setting
down; the bulk of the change is in utils.copy, since we want to
have IGNORE_FILES apply at every level of a recursively copied
directory (e.g. the theme static directory).

Since I was overhauling utils.copy anyway, I changed it to use
os.walk, which should be more efficient.

Other users of utils.copy should be unaffected.

### DIFF
--- a/pelican/utils.py
+++ b/pelican/utils.py
@@ -273,51 +273,73 @@ def slugify(value, substitutions=()):
     return value.decode('ascii')
 
 
-def copy(source, destination):
+def copy(source, destination, ignores=None):
     """Recursively copy source into destination.
 
     If source is a file, destination has to be a file as well.
-
     The function is able to copy either files or directories.
 
     :param source: the source file or directory
     :param destination: the destination file or directory
+    :param ignores: either None, or a list of glob patterns;
+        files matching those patterns will _not_ be copied.
     """
+
+    def walk_error(err):
+        logger.warning("While copying %s: %s: %s",
+                       source_, err.filename, err.strerror)
 
     source_ = os.path.abspath(os.path.expanduser(source))
     destination_ = os.path.abspath(os.path.expanduser(destination))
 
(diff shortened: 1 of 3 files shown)


pelican/utils.py
```diff
@@ -273,51 +273,73 @@ def slugify(value, substitutions=()):
...
+            logger.warning('Cannot copy %s (a directory) to %s (a file)',
+                           source_, destination_)
+            return
+
+        for src_dir, subdirs, others in os.walk(source_):
+            dst_dir = os.path.join(destination_,
+                                    os.path.relpath(src_dir, source_))
+
+            subdirs[:] = (s for s in subdirs if not any(fnmatch.fnmatch(s, i)
+                                                        for i in ignores))
```
avaris: Ignoring folders with `IGNORE_FILES` isn't documented and it doesn't match the use of `IGNORE_FILES` in other places. Similar issue with the early `return` on line 301.

zackw:
@avaris As there is no `IGNORE_DIRECTORIES`, I am inclined to say that `IGNORE_FILES` should work for both, and if it doesn't include directories elsewhere, *that* should be changed.

avaris:
@zackw sure, that's reasonable. If you include that in your PR and update documentation, that would be great. But, ignoring directories in one place but not in other places isn't consistent.

zackw:
@avaris Sorry for the long delay getting back to you.  As you suggested, I have made `IGNORE_FILES` apply to directories in the one other place that it's used, and updated the documentation.  I have also split the branch into three commits to clarify what piece does what.

avaris:
@zackw looks good, :+1:. Thank you for the work.

justinmayer:
Thanks, @zackw!

PR #1696, opened 2015-04-13
Clarify phrasing in quickstart documentation

dmd:
The quickstart was worded confusingly - it said "from your project directory", which implied doing a `cd ..` down to the `projects` dir, which would cause `pelican content` to fail. In fact, you need to be in the `yoursite` directory, which is the directory that has the `content` directory in it.


justinmayer:
Thanks, Daniel!

PR #1700, opened 2015-04-16
Fixes #1686: posixize paths in context['filenames'] to fix windows issues

### LINKED ISSUE #1686: "Mixed content in the same directory" broken on Windows
When a folder is both in `STATIC_PATHS` and `ARTICLE_PATHS` only non-source files are supposed to be copied raw. This does not seem to work on Windows, where articles are both parsed and copied.

The problem might be in *generators.py* with `add_source_path()`, which unlike `_is_potential_source_path()` does not perform a `os.path.normpath()` on the path, leading to slash/backslash mismatches. While adding it in fixes the problem, that breaks `{filename}` (and similar) links due to them not using `os.path.normpath()` either.

I'm using the latest master.

I'd appreciate if someone could verify this, I'm only just starting to use pelican and noticed this in the middle of developing a plugin (which I did disable to check I wasn't causing it, but still...) so this may not be the cleanest of test cases.

avaris:


ingwinlu:
+1

justinmayer:
Nice work! Thanks to @avaris for the contribution and to @ingwinlu for the review. (^_^)

Issue #1172, opened 2013-11-30
utils.slugify is very slow for non-English words


calfzhou:
`utils.slugify` calls `unidecode.unidecode` to convert all words (especially non-English characters) to their ASCII representation. This is very slow. My site have less than 100 posts, and dozens of Chinese tags, it takes 30+ seconds to build the site. If I manually change all tags to English, it takes only 15 seconds.

It might be better to let user disable `unidecode.unidecode` and define custom slugify mapping (from Unicode character to ASCII representation).

leotrs:
@justinmayer #1701 has been merged and states this issue can now be deferred to plugins.

justinmayer:
Ostensibly addressed via 1701.

PR #1701, opened 2015-04-16
Change URLWrapper.slug to a property

avaris:
Main goal is to delay `slugify` call until `slug` is needed. `slugify`
can be expensive depending on the input string (see #1172). Changing it
to a property gives plugins time to place custom `slug`s before `name`
is unnecessarily slugified.

With this change, default behavior is same except the time slugification
happens. But if you set custom slug, `slugify` won't be used at all.

So, this is a partial solution to #1172. The rest, setting a custom slug,
would best be handled by a plugin.

### DIFF
--- a/pelican/urlwrappers.py
+++ b/pelican/urlwrappers.py
@@ -13,12 +13,10 @@ logger = logging.getLogger(__name__)
 @functools.total_ordering
 class URLWrapper(object):
     def __init__(self, name, settings):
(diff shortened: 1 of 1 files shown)


ingwinlu:
looks good

justinmayer:
Nice work! Thanks to @avaris for the contribution and to @ingwinlu for the review.  (^_^)

PR #1703, opened 2015-04-20
remove tag_cloud from core

ingwinlu:
As discussed in #1598 and #1566 this breaks out tag cloud creation into a plugin, for which there is a PR in the [pelican-plugins repo](https://github.com/getpelican/pelican-plugins/pull/470).

- [x] remove tag_cloud from core
- [x] add tag_cloud to pelican-plugins
- [x] add deprecation warnings that tag clouds are not available without plugin

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -241,3 +241,12 @@ purposes. This can be achieved by explicitly specifying only the
 filenames of those articles in ``ARTICLE_PATHS``. A list of such
 filenames could be found using a command similar to ``cd content;
 find -name '*.md' | head -n 10``.
+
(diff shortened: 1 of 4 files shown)


ingwinlu:
options for deprecation:
*   hack something together that checks if there are tag_cloud settings present but no tag_cloud plugin loaded later, will be hacky and does not integrate with how we do deprecation right now, will also not catch cases where a tag_cloud was in use without additional configuration
*  add a f.a.q. entry ala 'since i upgraded pelican my tag cloud does not show up anymore'
* don't do anything

justinmayer:
Thanks for taking care of this, Winlu. Much appreciated!

PR #1708, opened 2015-04-29
Fix intra-site links to drafts

### LINKED ISSUE #865: Intra-site links for draft posts
Probably a really low-priority issue: When you link to a file that is currently under `:status: draft`, the links will not work.

```
.. _some href: |filename|my-drafted-article.rst
```

This will create a link to the article as if it was not in the drafts folder. Not a very pressing issue but a little inconsistent.

iksteen:
Overwrite an article's source path registration when it is 'upgraded' to a draft.

Fixes #865


justinmayer:
Many thanks for the fix, Ingmar!

PR #1715, opened 2015-05-08
Clarify STATIC_EXCLUDE_SOURCES documentatoin.

foresto:
Add some text to the documentation, explaining that this setting is for backward compatibility, and noting another setting that might be better for users wanting to publish their site sources.


justinmayer:
Thanks, Forest!

PR #1716, opened 2015-05-10
Add backtick missing from my last docs commit.

foresto:
Oops... I missed a closing backtick in my last settings.rst edit. (Sorry about that.) This fixes it.


ingwinlu:
#1690 would have helped ;)

PR #1720, opened 2015-05-13
make tox 2 compatible

ingwinlu:
* specify tox version in travis build
* fix test failures because ENV variables not set correctly


justinmayer:
Thanks for the quick fix, Winlu.

ingwinlu:
no problem, did not expect tox to release an updated that might break existing configurations.

however the build still failed since one builder could not install packages because of a connection reset; you should be able to demand a manual rebuild however somewhere on the travis site (somewhere under the gearwheel) and it should build fine

justinmayer:
Rebuild triggered. Passed. Thanks!

PR #1724, opened 2015-05-20
Remove duplicate tags

fernandezcuesta:
When a tag is entered more than once in metadata, pelican should discard duplicates. Same when there's no tag at all.

### DIFF
--- a/pelican/readers.py
+++ b/pelican/readers.py
@@ -25,13 +25,15 @@ from six.moves.html_parser import HTMLParser
 
 from pelican import signals
 from pelican.contents import Page, Category, Tag, Author
-from pelican.utils import get_date, pelican_open, FileStampDataCacher, SafeDatetime, posixize_path
+from pelican.utils import (get_date, pelican_open, FileStampDataCacher,
+                           SafeDatetime, posixize_path)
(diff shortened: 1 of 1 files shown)


ingwinlu:
might be better to handle this case at https://github.com/getpelican/pelican/blob/master/pelican/readers.py#L258 , so it has the same behaviour for any listed metadata

fernandezcuesta:
Just modified https://github.com/getpelican/pelican/blob/master/pelican/readers.py#L34 which applies as well to any listed metadata.

avaris:
Tests are failing because of the semi-random behavior of the `set` order. Might be better to use `OrderedDict`.

ingwinlu:
please look at http://pelican.readthedocs.org/en/latest/contribute.html#using-git-and-github and squash your commits

justinmayer:
Hi @fernandezcuesta. Would you please take a look at the comments on your pull request and make the relevant changes so we can get your contribution merged?

fernandezcuesta:
@justinmayer you may drop it, the fix is already achieved in c9183808021e54aa9b949939ff289b07c4443b7e

PR #1726, opened 2015-05-24
Add publishconf.py, "make publish" details to docs

StevenMaude:
Explain that the default `publishconf.py` overrides `pelicanconf.py` and add `make publish` to the make documentation.

The default of `publishconf.py` overriding `pelicanconf.py` didn't seem to be explained anywhere in the docs. This might be considered reasonably obvious to Python coders looking at the settings files, but may be overlooked otherwise.

### DIFF
--- a/docs/publish.rst
+++ b/docs/publish.rst
@@ -160,9 +160,15 @@ and thus doesn't require installing anything else in order to use it. The
 downside is that non-POSIX systems (e.g., Windows) do not include ``make``,
 and installing it on those systems can be a non-trivial task.
 
-If you want to use ``make`` to generate your site, run::
+If you want to use ``make`` to generate your site using the settings in
+``pelicanconf.py``, run::
 
     make html
+
+To generate the site for production, using the settings in ``publishconf.py``,
+run::
+
+    make publish
 
 If you'd prefer to have Pelican automatically regenerate your site every time a
 change is detected (which is handy when testing locally), use the following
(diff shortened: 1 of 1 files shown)


jonathan-s:
:+1:

ingwinlu:
I am not sure on this one, I would prefer a wording that sounds less like it was some kind of pelican feature that handles this.

IMO there should be something along the lines of: To base your publish configuration on top of your pelicanconf you can import it via `import blabla`, which is generated by default in the pelican-quickstart publishconf.

StevenMaude:
@ingwinlu Thanks for taking a look. Your suggested wording was clearer, so I've reworked the pull request with that in mind.

docs/publish.rst
```diff
@@ -61,6 +61,15 @@ feeds, etc.) that you may have defined::
 
     pelican content -s publishconf.py
 
+To base your publish configuration on top of your ``pelicanconf.py``, you
+can import your ``pelicanconf`` settings by including in ``publishconf.py``
+the line::
```
ingwinlu: you can import your pelicanconf settings by including the following line in your publishconf

ingwinlu:
@StevenMaude can you fix that one sentence? Then I am ok with merging this.

StevenMaude:
@ingwinlu Done.

justinmayer:
Thanks to @StevenMaude for the contribution and to @ingwinlu for the review!

PR #1728, opened 2015-05-24
Fix for stopping devserver

jonathan-s:
When running `make devserver` and then running `make stopserver` it doesn't stop the server. This patch fixes that.


justinmayer:
Thanks for the fix, Jonathan.

PR #1737, opened 2015-06-01
use protocol relative urls in notmyidea

### LINKED ISSUE #1736: Load GitHub ribbon via https
http://s3.amazonaws.com/github/ribbons/forkme_right_red_aa0000.png results in a "Mixed Content" warning in Chrome if the blog itself is loaded via https. You should change it to https

ingwinlu:
that's all I could find with a quick `grep -r "http:" .` 

resolves #1736

### DIFF
--- a/pelican/tests/output/basic/a-markdown-powered-article.html
+++ b/pelican/tests/output/basic/a-markdown-powered-article.html
@@ -7,7 +7,7 @@
         <link href="/feeds/all.atom.xml" type="application/atom+xml" rel="alternate" title="A Pelican Blog Atom Feed" />
 
         <!--[if IE]>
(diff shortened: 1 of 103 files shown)


ingwinlu:
yeah shame on me, still need to correct tests

justinmayer:
The [general consensus](http://www.paulirish.com/2010/the-protocol-relative-url/) regarding this topic is that the protocol-relative URL is now considered to be an anti-pattern.

In other words, if a resource is available via HTTPS, we should always request it via HTTPS.

justinmayer:
Thanks for taking the initiative so quickly, @ingwinlu. I addressed the issue via https://github.com/getpelican/pelican/commit/940eb76b7f70b1c9c7f833d5328d44cb19bde406, utilizing the aforementioned HTTPS-everywhere approach.

PR #1740, opened 2015-06-03
Improve pluralization, fixing display count of hidden pages

### LINKED ISSUE #1739: Hidden pages not counted when generating
If you set status to hidden on a page it won't be counted when generating. These pages are nonetheless generated. So there should be a counter for hidden as well.

ingwinlu:
*  display count of hidden pages
*  pluralize only if necessary
*  add maybe_pluralize to utils
*  add tests for maybe_pluralize

resolves #1739


jonathan-s:
Looks good :)

justinmayer:
Thanks, @ingwinlu!

PR #1741, opened 2015-06-03
RST example in quickstart

abrahamvarricatt:
The quickstart tutorial in our docs does not have any examples in reStructuredText format. This is to add it in alongside the existing Markdown one.

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -46,6 +46,21 @@ text editor to create your first article with the following content::
 Given that this example article is in Markdown format, save it as
 ``~/projects/yoursite/content/keyboard-review.md``.
 
+Alternatively, here is the same article in reStructuredText format::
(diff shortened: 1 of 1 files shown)


ingwinlu:
As mentioned in IRC I don't think this is needed.

Quickstart should be a very brief example on what you have to do to produce something with pelican for people who are not familiar with static site generation.

There are examples on rst files throughout the net and pelican has some rst examples in the documentation and ships with a whole [directory](https://github.com/getpelican/pelican/tree/master/samples/content) full of examples.

abrahamvarricatt:
What about a link or some mention in the quickstart pointing to that directory? I'd never known about those examples until you just pointed it out.

justinmayer:
Added a link via ecc5d0b, so closing this out as complete.

PR #1723, opened 2015-05-17
Improve exception logging a bit more. Now it always dumps a traceback…

ionelmc:
…. Ref: #1717, #1718.

PR #1718, opened 2015-05-11
Improve internal error reporting. Fixes #1717.

ionelmc:

PR #1743, opened 2015-06-03
Fix exception logging

avaris:
Adds exception type to the exception logging.
Removes the extra logging for autoreload in debug mode, since `raise`
will make it caught by the global `try/except` below and it'll be
logged there.

This replaces #1723 and also removes extra exception logging caused
in #1718.


justinmayer:
Many thanks, @Avaris!

Issue #1356, opened 2014-05-20
Draft posts should not depend on a date


kylef:
Since it's a draft, you won't know the date yet. It makes very little sense to require the date for drafts.

ehashman:
I was hoping to help close this bug but I'm not sure what the problem is. If you create an article with `status: draft`, no date is required for the draft to generate an article. @kylef, what is the desired fix?

ingwinlu:
@ehashman see the above linked open issue I opened that explains the issue

leotrs:
If drafts now don't really depend on a date, then we can probably continue discussion at #1748, reference this issue from there and close this.

justinmayer:
@leotrs: Agreed. If someone determines that draft posts do indeed require a date, please comment here and I will re-open this issue.

PR #1745, opened 2015-06-04
[WIP] cache versioning

ingwinlu:
if pelican is updated, the cache should be invalidated automatically

# Todo
- [x] ~~break out caching from utils into seperate module~~
      - [x] ~~break out caching classes into `cache.py`~~
      - [x] ~~move caching tests to `test_cache.py`~~
      - [x] ~~cleanup unneeded / duplicate code after move~~
- [x] ~~fix existing cache tests and remove `assert_called_count` which does not seem to be part of mock anymore~~
- [x] ~~use cPickle instead of pickle for increased speed~~
- [x] ~~set pelican's version to 3.6.0.dev or something so master branch installs don't get the same treatment as 3.5.0~~
- [x] ~~invalidate 'bad' cache automatically~~
    - [x] ~~invalidate when version of pelican changes~~
    - [x] ~~invalidate when settings change~~
- [ ] redo generate_context
    - [x] ~~don't cache broken content~~
    - [x] ~~cache drafts~~
    - [ ] restructure generate_context some more to solve #1356 and produce cleaner code
- [ ] find a way to not use `import pelican` in `cache.py` to access `__version__`
- [ ] check if settings invalidation is handled correctly
- [ ] check if there is anything we can do to track if plugins changed
- [ ] make sure the introduced overhead is ok, maybe make cache invalidation optional via settings
- [ ] benchmark

# Out of Scope
* restructure caching?

  FileDataCacher is currently the top level caching instance, it might make sense to introduce a 'caching manager' which contains most of the setting, pathing logic (and then in the future keeps track of versions) and let FileDataCacher just do what the name implies

### DIFF
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -25,7 +25,7 @@ from pelican.settings import read_settings
 from pelican.utils import clean_output_dir, folder_watcher, file_watcher
 from pelican.writers import Writer
 
-__version__ = "3.5.0"
+__version__ = "3.6.0.dev"
 
 DEFAULT_CONFIG_NAME = 'pelicanconf.py'
 
--- a/pelican/cache.py
+++ b/pelican/cache.py
@@ -8,6 +8,9 @@ try:
 except:
     import pickle
 
+#TODO don't load pelican module, maybe have to move version definition
+import pelican
+from pelican.settings import settings_check_equal
 from pelican.utils import mkdir_p
 
 
@@ -25,6 +28,7 @@ class FileDataCacher(object):
         Sets caching policy according to *caching_policy*.
         """
         self.settings = settings
+        self._cache = {}
         self._cache_path = os.path.join(self.settings['CACHE_PATH'],
                                         cache_name)
         self._cache_data_policy = caching_policy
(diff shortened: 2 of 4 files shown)


ingwinlu:
pinging @smartass101 

the current cache tests all feature `assert_called_count` which I can not find in the mock documentation; whats up with that?
https://github.com/getpelican/pelican/blob/master/pelican/tests/test_generators.py#L377

smartass101:
Perhaps it is a remnant of some obsolete unittests version. Try using `assertEqual(mock.called_count, 0)` if that still works.

ingwinlu:
thanks for the quick reply @smartass101 
i now found https://github.com/getpelican/pelican/commit/aee103544079aa6405b0b274859edb598709563a

apparently you stumbled upon that as well in another function, one of the now fixed tests fails now sadly (from the quick look I had at it it tries to use another folder for caching then what is expected)

ingwinlu:
just to further document:

the tests failed because the generator code tried to cache invalid files. during the fix of those problems it came up that generate_context in ArticlesGenerator and PagesGenerator could use a restructure to be more readable and not generally awful and hacked as it was before

ingwinlu:
going to close this for now:

I feel like scanning all plugins (since plugins are not required to set a version) and hashing their module contents feels like too much overhead.

Still not wasted, showed that cache tests were broken, and we now have a better caching structure, leveraging cPickle for more speed.

smartass101:
@ingwinlu, there are ways to hash stuff with less overhead by using non-cryptographic algorithms, e.g. [xxHash](https://github.com/Cyan4973/xxHash). However, it should be optinal as this dependency is harder to obtain.

ingwinlu:
efficiency aside, I think I am going to follow my own goto response to new features: 'this should be an optional plugin'

Issue #1689, opened 2015-04-11
Decide on default content caching setting


smartass101:
I think it's obvious that content caching is a useful setting and it does speed up things. However, since 3.4 there have been far too many instances when people encountered a problem that was due to caching being on by default, particulary when experimenting with different options at the beginning of their pelican usage.

I think we should decide whether to turn caching off by default in 3.6 and mention in docs that it can be turned on.

Rationale:
* Most new users will start by experimenting with settings and will enable caching once they arrive at a final shape of their website.
* Long-term users can just add enable caching if they want to continue to use it. Until they do, the original cache will still be there in their cache/ folder so they won't lose anything. However, it would be fair to log a message of some sort that caching is off.

Not sure what the "caching is off" warning should look like. maybe something like in settings.py
```Python
if not settings['LOAD_CONTENT_CACHE']:
    logger.info('Content cache not loaded, all content will be re-read')
if not settings['CACHE_CONTENT']:
    logger.info('Content caching is disabled, newly read content will be re-read next time')
```
These info messages would also make things clearer in general IMHO.

What do you think?
Once decided, I'll commit the changes to the settings, docs and logging.

iKevinY:
I agree with making this change. The progression from experimenting with site options to learning how to enable the cache seems more logical than the current situation.

avaris:
Given that I advocated this for 3.5, I'm all for it.

ingwinlu:
go for it.

smartass101:
What do you think @justinmayer?

justinmayer:
I agree that disabled-by-default is probably best from a *principle of least astonishment* perspective.

I'm not sure that the warning is needed. If generation is slow, I presume folks will be incentivized to, ya know, read the docs and learn about caching. What are everyone's opinions on that?

pxquim:
I'm not even certain that caching is such a big win, at least not always. For example, some plugins process articles into BeautifulSoup and back, and that this processing is not cached. I've only seen modest performance wins, which are offset by the occasional cache issue.

smartass101:
@justinmayer, I was thinking of a warning mainly to give a  heads up and simplify debugging for people on large sites when suddenly the new version slows down.

@pxquim, it is mainly useful on larger sites. By default content is cached only at the reader level, but if plugins modify content extensively, the generator level might help a lot more.

robclancy:
We have builds going past 10 minutes WITH caching, so yes it is needed. We have big sites.

[...]

PR #1747, opened 2015-06-06
Fix caching and disable by default

ingwinlu:
this is a bugfixing pr resulting from my work in #1745  and just fixes existing errors and bugs

adds changes discussed in #1689


justinmayer:
Many thanks, Winlu. Much appreciated.

Any comments on this from other folks? (cc: @kylef / @smartass101 / @avaris / @iKevinY)

pelican/__init__.py
```diff
@@ -101,6 +101,12 @@ def _handle_deprecation(self):
                             'PAGE_LANG_URL'):
                 logger.warning("%s = '%s'", setting, self.settings[setting])
 
+        if self.settings.get('AUTORELOAD_IGNORE_CACHE'):
+            logger.warning('Found deprecated `AUTORELOAD_IGNORE_CACHE` in '
+                           'settings. It has no effect now and can be safely '
+                           'removed.')
```
avaris: Umm, why? People can still enable cache and might want to disable it in ``--autoreload``.

iKevinY:
Docstrings should probably use `"""` rather than `'''` for the sake of consistency.

ingwinlu:
@iKevinY 
I did not touch those, they are the same as they were in utils.py (there we use ')

I am hesitant to do a unification in this pr as it is not strictly part of it, but I agree with you that we should do a pep8 pr soon (maybe even before the next release)

iKevinY:
@ingwinlu Yeah, I realized that after reading through the changes more closely. I think it would be safe to make that change in this PR since it's purely syntactic.

Definitely agree on the need for a style-standardization pull request in the near future.

ingwinlu:
@avaris 
The same behaviour without that setting can still be archived via `--autoreload --ignore-cache`. So no loss of functionality.

personal opinion:
settings.py is pretty packed, people might look at it and not find what they are looking for. Cache alone had 7. I don't feel like the level of customization offered is needed at all. Any setting we can remove is a good thing (less complexity, less to document, less to break)

ingwinlu:
@iKevinY flake8 gently reminded me that i forgot a hashlib import, so definitely was worth checking style ;)

avaris:
@ingwinlu, I'm fine with that. But, I think it would be better if we mention ``--ignore-cache`` in the deprecation warning. Something like:

```python
logger.warning('Found deprecated `AUTORELOAD_IGNORE_CACHE` in ' 
               'settings. Use --ignore-cache option instead.')
```

ingwinlu:
When everyone is happy with the review commits I will squash them into the 2 main ones and we should be good to go

avaris:
:+1:

justinmayer:
Excellent work, Winlu.  (^_^)

PR #1753, opened 2015-06-09
Make Pelican codebase compliant with PEP8

### LINKED ISSUE #1752: make source code pretty
* docutils strings are `'''` and `"""` in some places.
* lots of pep8 violations
* add a tox environment that runs flake8 checks
```
[testenv:flake8]
basepython = python2.7
deps =
    flake8
commands =
    flake8 pelican
```

ingwinlu:
* add flake8 testenv to tox.ini
* make existing code compatible with pep8 standard

TODO:
* find someone insane who actually reviews this :/

resolves #1752

### DIFF
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -1,45 +1,41 @@
 # -*- coding: utf-8 -*-
-from __future__ import unicode_literals, print_function
-import six
+from __future__ import print_function, unicode_literals
 
+import argparse
+import collections
+import locale
+import logging
 import os
 import re
 import sys
 import time
-import logging
-import argparse
-import locale
-import collections
+
+import six
 
 # pelican.log has to be the first pelican module to be loaded
 # because logging.setLoggerClass has to be called before logging.getLogger
-from pelican.log import init
-
+from pelican.log import init  # noqa
 from pelican import signals
-
 from pelican.generators import (ArticlesGenerator, PagesGenerator,
-                                StaticGenerator, SourceFileGenerator,
+                                SourceFileGenerator, StaticGenerator,
                                 TemplatePagesGenerator)
 from pelican.readers import Readers
 from pelican.settings import read_settings
-from pelican.utils import (clean_output_dir, folder_watcher,
-                           file_watcher, maybe_pluralize)
+from pelican.utils import (clean_output_dir, file_watcher,
+                           folder_watcher, maybe_pluralize)
 from pelican.writers import Writer
 
 __version__ = "3.6.4.dev0"
-
 DEFAULT_CONFIG_NAME = 'pelicanconf.py'
-
-
 logger = logging.getLogger(__name__)
 
 
 class Pelican(object):
 
     def __init__(self, settings):
-        """
-        Pelican initialisation, performs some checks on the environment before
-        doing anything else.
+        """Pelican initialisation
+
+        Performs some checks on the environment before doing anything else.
         """
 
         # define the default settings
@@ -152,7 +148,7 @@ class Pelican(object):
         context = self.settings.copy()
         # Share these among all the generators and content objects:
         context['filenames'] = {}  # maps source path to Content object or None
-        context['localsiteurl'] = self.settings['SITEURL'] 
+        context['localsiteurl'] = self.settings['SITEURL']
 
         generators = [
             cls(
@@ -190,23 +186,23 @@ class Pelican(object):
                                if isinstance(g, PagesGenerator))
 
         pluralized_articles = maybe_pluralize(
-            len(articles_generator.articles) +
-                len(articles_generator.translations),
+            (len(articles_generator.articles) +
+             len(articles_generator.translations)),
             'article',
             'articles')
         pluralized_drafts = maybe_pluralize(
-            len(articles_generator.drafts) +
-                len(articles_generator.drafts_translations),
+            (len(articles_generator.drafts) +
+             len(articles_generator.drafts_translations)),
             'draft',
             'drafts')
         pluralized_pages = maybe_pluralize(
-            len(pages_generator.pages) +
-                len(pages_generator.translations),
+            (len(pages_generator.pages) +
+             len(pages_generator.translations)),
             'page',
             'pages')
         pluralized_hidden_pages = maybe_pluralize(
-            len(pages_generator.hidden_pages) +
-                len(pages_generator.hidden_translations),
+            (len(pages_generator.hidden_pages) +
+             len(pages_generator.hidden_translations)),
             'hidden page',
             'hidden pages')
 
@@ -243,8 +239,8 @@ class Pelican(object):
         return generators
 
     def get_writer(self):
-        writers = [ w for (_, w) in signals.get_writer.send(self)
-                    if isinstance(w, type) ]
+        writers = [w for (_, w) in signals.get_writer.send(self)
+                   if isinstance(w, type)]
         writers_found = len(writers)
         if writers_found == 0:
             return Writer(self.output_path, settings=self.settings)
@@ -254,15 +250,15 @@ class Pelican(object):
                 logger.debug('Found writer: %s', writer)
             else:
                 logger.warning(
-                    '%s writers found, using only first one: %s', 
+                    '%s writers found, using only first one: %s',
                     writers_found, writer)
             return writer(self.output_path, settings=self.settings)
 
 
 def parse_arguments():
     parser = argparse.ArgumentParser(
-        description="""A tool to generate a static blog,
-        with restructured text input files.""",
+        description='A tool to generate a static blog, '
+                    ' with restructured text input files.',
         formatter_class=argparse.ArgumentDefaultsHelpFormatter
     )
 
@@ -354,7 +350,7 @@ def get_config(args):
     # argparse returns bytes in Py2. There is no definite answer as to which
     # encoding argparse (or sys.argv) uses.
     # "Best" option seems to be locale.getpreferredencoding()
-    # ref: http://mail.python.org/pipermail/python-list/2006-October/405766.html
+    # http://mail.python.org/pipermail/python-list/2006-October/405766.html
     if not six.PY3:
         enc = locale.getpreferredencoding()
         for key in config:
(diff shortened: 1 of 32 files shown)


ingwinlu:
As @avaris mentioned in IRC, `flake8` is picky;

however, while some lines in this pr might be less readable than their original form it provides some notable improvements when we have an automated check running:
* more consistent codebase
* checks for unneeded / wrong imports
* brings the codebase inline with what we list in `CONTRIBUTING.rst`, running flake8 is a must
* cleaning up the codebase should be a priority, as it has grown into a lot of directions, by a lot of people (which is good). but diving into it right now is not exactly easy and straightforward. 

since this is quite a large pr I suggest reviewing it line by line and commenting on the diff if you are ok with it.

the usual suspects: @avaris @iKevinY @smartass101 @kylef @justinmayer

pelican/contents.py
```diff
@@ -93,13 +93,16 @@ def __init__(self, content, metadata=None, settings=None,
...
         # setting of SLUG_ATTRIBUTE
         if not hasattr(self, 'slug'):
-            if settings['SLUGIFY_SOURCE'] == 'title' and hasattr(self, 'title'):
-                self.slug = slugify(self.title,
-                                settings.get('SLUG_SUBSTITUTIONS', ()))
-            elif settings['SLUGIFY_SOURCE'] == 'basename' and source_path != None:
-                basename = os.path.basename(os.path.splitext(source_path)[0])
-                self.slug = slugify(basename,
-                                settings.get('SLUG_SUBSTITUTIONS', ()))
+            if settings['SLUGIFY_SOURCE'] == 'title' and \
```
iKevinY: Should probably avoid using backslashes by writing this like the `elif` block below:

```python
if (settings['SLUGIFY_SOURCE'] == 'title' and
    hasattr(self, 'title')):
```

pelican/writers.py
```diff
@@ -139,9 +139,10 @@ def write_file(self, name, template, context, relative_urls=False,
         :param **kwargs: additional variables to pass to the templates
         """
 
-        if name is False or name == "" or\
-           not is_selected_for_writing(self.settings,\
-               os.path.join(self.output_path, name)):
+        if name is False or \
```
iKevinY: More backslash removal here.

```python
if (name is False or
    name == "" or
    not is_selected_for_writing(self.settings,
                                os.path.join(self.output_path, name)):
```

pelican/utils.py
```diff
@@ -546,12 +543,12 @@ def file_times(path):
             dirs[:] = [x for x in dirs if not x.startswith(os.curdir)]
 
             for f in files:
-                if (f.endswith(tuple(extensions)) and
-                    not any(fnmatch.fnmatch(f, ignore) for ignore in ignores)):
-                    try:
-                        yield os.stat(os.path.join(root, f)).st_mtime
-                    except OSError as e:
-                        logger.warning('Caught Exception: %s', e)
+                if f.endswith(tuple(extensions)) and \
```
iKevinY:
```python
if (f.endswith(tuple(extensions)) and
    not any(fnmatch.fnmatch(f, ignore) for ignore in ignores)):
```

pelican/utils.py
```diff
@@ -546,12 +543,12 @@ def file_times(path):
...
+                   not any(fnmatch.fnmatch(f, ignore) for ignore in ignores):
+                        try:
```
iKevinY: Is there supposed to be indentation added to the `try/except` statement here?

iKevinY:
One fairly controversial thing that we should decide on is whether the 80-column rule should be enforced or not. There are quite a few places in Pelican's codebase where code reaches several indentation levels deep, and with an 80-column limit, this doesn't leave much room to work with, resulting in this:

```diff
- function(arg1, arg2, arg3)
+ function(
+    arg1,
+    arg2,
+    arg3)
```

Which in some cases is good, but in other cases just results in a drop in readability. There are some alternatives, like an 80-column "soft limit", where people try to end their lines at 80 columns, but going over by ten characters or so isn't the end of the world.

pelican/contents.py
ingwinlu: sadly that is not possible because it would line up with the intendation of the next logical block (4 chars `if (`) which throws another error.

another solution for this is double intending the second and line which results in the following:

pelican/writers.py
ingwinlu: not sure removing all backslashes is a thing, in this case i opted against it to keep it from producing ))) at the end which would not help readability

pelican/utils.py
ingwinlu: the extra indentation is added to distinguish from the not line before; else there would only be 1 space or so difference, which makes it hard to read

ingwinlu:
I opt for the hard limit on the !!79!! character limit as it is enforced by flake8 and pep8.

If you need more than that chances are that you are nesting a lot of ifs (and that is the reason why you run out of space, example would be the autoreload section in __init__) or other similarly unwieldy stuff.

you also don't need to intend every arg after a function, that is purely a choice I make when the function readability profits from it. (example would be the assertEqual calls in the tests, where you can see more easily what gets compared).

I would prefer not to have a discussion on char limit. If you want do discuss then keep it to if we want flake8 enforced or if we want to continue ignoring it even though we have it in contributing.rst.

pelican/tests/test_readers.py
```diff
@@ -185,18 +187,18 @@ def test_article_extra_path_metadata(self):
...
         expected_without_metadata = {
-            'category' : 'misc',
-            'author'   : 'Charlès Overwrite',
-            'title'    : 'Article title',
-            'reader'   : 'rst',
+            'category': 'misc',
+            'author':   'Charlès Overwrite',
+            'title':    'Article title',
+            'reader':   'rst',
         }
```
ingwinlu: i removed those on @avaris suggestion

pelican/tests/test_readers.py
```diff
@@ -212,13 +214,13 @@ def test_article_extra_path_metadata_dont_overwrite(self):
...
-            'reader': 'rst',
-            'key-1b': 'value-1b'
+            'author':   'Alexis Métaireau',
+            'title':    'Rst with filename metadata',
+            'date':     SafeDatetime(2012, 11, 29),
+            'slug':     'rst_w_filename_meta',
+            'mymeta':   'foo',
+            'reader':   'rst',
+            'key-1b':   'value-1b'
         }
```
ingwinlu: i removed those on @avaris suggestion

pelican/tests/test_readers.py
```diff
@@ -165,14 +167,14 @@ def test_article_extra_path_metadata(self):
...
-            'key-1b': 'value-1b'
+            'author':   'Alexis Métaireau',
+            'title':    'Rst with filename metadata',
+            'date':     SafeDatetime(2012, 11, 29),
+            'slug':     'rst_w_filename_meta',
+            'mymeta':   'foo',
+            'reader':   'rst',
+            'key-1a':   'value-1a',
+            'key-1b':   'value-1b'
         }
```
ingwinlu: i removed those on @avaris suggestion

pelican/contents.py
iKevinY: Ah I see; that's rather inconvenient. Up to you for the best course of action then.

pelican/__init__.py
```diff
@@ -330,10 +332,12 @@ def get_config(args):
         config['RELATIVE_URLS'] = args.relative_paths
     config['DEBUG'] = args.verbosity == logging.DEBUG
 
-    # argparse returns bytes in Py2. There is no definite answer as to which
-    # encoding argparse (or sys.argv) uses.
-    # "Best" option seems to be locale.getpreferredencoding()
-    # ref: http://mail.python.org/pipermail/python-list/2006-October/405766.html
+    """
```
QuLogic: What is the benefit of changing these comments to a docstring? It's not actually documenting anything.

pelican/utils.py
ingwinlu: not sure removing all backslashes is a thing, in this case i opted against it to keep it from producing )) at the end which would not help readability

ingwinlu:
updated;

also cleaned up all headers and sorted imports alphabetically (at least the modules) and incorporated most of the feedback gathered so far

pelican/cache.py
```diff
@@ -1,15 +1,17 @@
+# -*- coding: utf-8 -*-
 from __future__ import unicode_literals
 
 import hashlib
 import logging
 import os
+
+from pelican.utils import mkdir_p
```
kylef: I don't think this change is correct. From PEP-8:

> Imports should be grouped in the following order:
>
> 1. standard library imports
> 2. related third party imports
> 3. local application/library specific imports
>
> You should put a blank line between each group of imports.

This is a local import and should be below the standard library import of pickle.

ingwinlu: First, `pep8` tool currently does not check the order of imports, it only makes sure that there are no unneeded imports or double imports;

You are right that the pickle import is not local and should be above, my reasoning for keeping this order is the try block.

When looking at the file you first got the imports, which generally do not contain any logic, but whlie trying to import cpickle we suddenly have logic (if an error occurs, do something different). For me that is something different to a normal import and should be below the normal import blocks.

I was also the original author of this line, so from a coder point of few I say this change was not needed. While reviewing however I felt differently and wanted to move it.

All in all pep8 only defines the order of imports as 'should' and not 'must'. I will however bow to popular demand on this one. (Things like this are the reason I did not just run autopep on it and went home)

QuLogic: To avoid the logic here, you can import `cPickle` from `six.moves`.

ingwinlu: while this is true in this case, we can not always apply six.moves

ingwinlu:
adding a flake8 plugin for import order additionally wields [another set of errors](https://gist.github.com/ingwinlu/3d4ac96493229f7b7c6e)

ingwinlu:
last update now ignoring the import order statement errors as I am not happy with either directives suggestions.

to get it in check we could however handle lib imports in pelican utils and import them from there to get a nicer ordering

justinmayer:
Thank you, Winlu, for what I'm sure was a very significant undertaking here. I've reviewed the changes and they look fine to me. Personally, I'd prefer to set `max-line-length = 85`, but I don't feel strongly enough about it to make a fuss.  (^_^)

Anyone else have any comments on these changes before they are merged?

justinmayer:
Thanks again for all the hard work, Winlu!

PR #1755, opened 2015-06-11
fix outdated content_object_init refs

### LINKED ISSUE #1711: signals.content_object_init.connect not working as described in docs
The docs suggest adding a sender parameter when connecting the `content_object_init` signal. However, when I do that, just copying the example in the docs, nothing happens. When I just put the method name, with no additional argument, the signal appears to be sent correctly. Searching through the plugins repo, it appears that none of the plugins use the additional sender argument as suggested in the docs. Is this functionality no longer there? If so, the docs should be corrected.

The relevant section of the documentation:

> The signal content_object_init can send a different type of object as the argument. If you want to register only one type of object then you will need to specify the sender when you are connecting to the signal.

> ```python
from pelican import signals
from pelican import contents

> def test(sender, instance):
        print "%s : %s content initialized !!" % (sender, instance)

> def register():
        signals.content_object_init.connect(test, sender=contents.Article)
```

ingwinlu:
* remove content_object_init section from docs
* improve content_object_init test

The content_object_init signal used to set its class as sender and pass
the instance as additional arg in 6100773. Commit ed907b4 removed this
behaviour to bring it inline with other signals, on the basis that
you can test for the class of the object anyway.
We also had a test in place, checking this behaviour, but it was poorly
implemented, not actually checking if the function ever got called.

closes #1711

### DIFF
--- a/docs/plugins.rst
+++ b/docs/plugins.rst
@@ -116,30 +116,10 @@ static_generator_preread            static_generator               invoked befor
                                                                    staticfiles list.
 static_generator_init               static_generator               invoked in the StaticGenerator.__init__
 static_generator_finalized          static_generator               invoked at the end of StaticGenerator.generate_context
-content_object_init                 content_object                 invoked at the end of Content.__init__ (see note below)
+content_object_init                 content_object                 invoked at the end of Content.__init__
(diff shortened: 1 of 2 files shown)


justinmayer:
Thanks, Winlu!

PR #1756, opened 2015-06-12
Fix #1647: Fix ARTICLE_ORDER_BY and add the ability to reverse order

### LINKED ISSUE #1647: Metadata field not recognized
I can't seem to get ARTICLE_ORDER_BY working.  I see this message in the log irrespective of the custom metadata field I try to sort by:

WARNING: There is no "date_parsed" attribute in the item metadata.Defaulting to slug order.

I know the metadata is present because I can use it in Jinja templates like so:  {{ article.date_parsed }}

I'm using Pelican 3.5.0 and generating RST files that contain the metadata.  Am I missing something obvious?

avaris:
ARTICLE_ORDER_BY wasn't doing anything because the ArticlesGenerator
was sorting articles after ARTICLE_ORDER_BY was applied. This fixes
that by adding the ability to reverse metadata order by adding the
option prefix 'reversed-' to metadata and changing the default value
to 'reversed-date'.

Relevant documentation is also updated and moved into a more appropriate
place ('Ordering Content' instead of 'URL settings').

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -277,14 +277,6 @@ Setting name (followed by default value, if any)        What does it do?
 ======================================================  ==============================================================
 ``ARTICLE_URL = '{slug}.html'``                         The URL to refer to an article.
 ``ARTICLE_SAVE_AS = '{slug}.html'``                     The place where we will save an article.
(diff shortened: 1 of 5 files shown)


justinmayer:
Thanks, @avaris!

PR #1760, opened 2015-06-16
Export hidden pages in context

jml:
The `PageGenerator` was building hidden pages, but not making them
available in the context. This makes it difficult for other plugins to
operate on hidden pages.

This patch updates `PageGenerator` to export the hidden pages it finds
in the context as `hidden_pages`.

[Review on Reviewable](https://reviewable.io/reviews/getpelican/pelican/1760)

### DIFF
--- a/pelican/generators.py
+++ b/pelican/generators.py
@@ -577,7 +577,7 @@ class ArticlesGenerator(CachingGenerator):
         self.authors.sort()
 
         self._update_context(('articles', 'dates', 'tags', 'categories',
-                              'authors', 'related_posts'))
+                              'authors', 'related_posts', 'drafts'))
         self.save_cache()
         self.readers.save_cache()
         signals.article_generator_finalized.send(self)
@@ -643,7 +643,7 @@ class PagesGenerator(CachingGenerator):
         self.hidden_pages, self.hidden_translations = (
             process_translations(hidden_pages))
 
-        self._update_context(('pages', ))
+        self._update_context(('pages', 'hidden_pages'))
         self.context['PAGES'] = self.pages
 
         self.save_cache()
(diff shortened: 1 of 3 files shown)


ingwinlu:
maybe add the same behaviour for drafts as well

jml:
Done. Thanks.

ingwinlu:
you still need to update the docs https://github.com/getpelican/pelican/blob/master/docs/themes.rst

pelican/generators.py
```diff
@@ -643,7 +643,7 @@ def generate_context(self):
         self.hidden_pages, self.hidden_translations = (
             process_translations(hidden_pages))
 
-        self._update_context(('pages', ))
+        self._update_context(('pages', 'hidden_pages'))
         self.context['PAGES'] = self.pages
```
ingwinlu: should probably be removed; it is no where documented that pages is available as `pages` and `PAGES` + it is the (only?) variable I know of that actually is made available as UPPERCASE.

If suggest deprecating `PAGES`, issuing a warning if it is called, and update the default themes to use pages instead of PAGES.

@avaris @justinmayer thoughts?

avaris: "issue a warning if it is called" will be tricky. Otherwise, I agree.

jml: That sounds like a great thing to do in a separate PR.

justinmayer: I agree that should be changed, but in a separate PR as @jml suggested.

justinmayer:
Many thanks for the contribution, Jonathan. Now that the documentation has been added, would you be so kind as to squash the commits prior to merge?

jml:
Thanks, am happy to help. Commits squashed.

jml:
Anything else I need to do?

justinmayer:
Thanks for the enhancement, Jonathan.

Issue #1493, opened 2014-10-09
Performance is Sluggish


mlissner:
Since I started using Pelican, I've been a bit annoyed how long it takes for it to generate my content. Right now it takes Pelican about 15s to generate my site (225 posts), which is just too much for fast iteration while I'm writing.

I dug out a profiler this evening and did a little test:

    python -m cProfile -s time `which pelican` content  -s pelicanconf.py

This gave me some output that indicated that the `slugify` function was called a whopping 300,504 times on my puny blog! 

    Done: Processed 228 article(s), 1 draft(s) and 5 page(s) in 21.45 seconds.
             22876023 function calls (22138959 primitive calls) in 21.782 seconds
    
       Ordered by: internal time

       ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        43614    3.182    0.000    3.182    0.000 {method 'write' of '_io.BufferedWriter' objects}
       300504    2.145    0.000   10.506    0.000 utils.py:227(slugify)
    1399152/1390065    1.771    0.000    1.903    0.000 {method 'sub' of '_sre.SRE_Pattern' objects}
       300504    1.358    0.000    1.880    0.000 __init__.py:22(unidecode)
    1618724/1618686    0.889    0.000    0.903    0.000 {built-in method hasattr}
       303722    0.761    0.000    1.312    0.000 __init__.py:126(unescape)
       303722    0.688    0.000    2.934    0.000 __init__.py:148(striptags)

I'm not an expert on profiling, but if I'm reading this right, it looks like the slowest thing is writing to disk (I have an SSD!) and the second slowest is the `slugify` function, which took over ten of the 21 seconds.

I started picking at `slugify` in hopes of making it faster. One idea from an IRC chatroom was to memoize the function, but when I tried that using `@functools.lru_cache(maxsize=2048)`, it threw:

    RuntimeError: maximum recursion depth exceeded

And I found the `memoize` decorator in `utils.py`, but it threw an error too. 

I guess this is a bug report to say:

1. Slugify is rather, um, sluggish.
1. What's with all the recursion, can we fix it? I'm not sure I understand it properly, but it looks like it will be a problem when blogs using pelican get bigger.
1. Why does my minor blog call slugify 300k times? I do NOT have that many links. That'd be over 1000 per page, practically.
1. Has any effort gone into this and does anybody have ideas for fixing it?

I'm going to move on to other things until I hear back here, but if we can get a conversation going, I'll be happy to try to make this faster.

[...]

PR #1764, opened 2015-06-19
Improve URLWrapper comparison

### LINKED ISSUE #1758: improve URLWrappers comparision
current URLWrappers slugify a lot, resulting in a lot of processing time lost ( #1493 ).

we need to check if we can avoid the always occurring slugify call (which is wrong anyway) and speed up comparisions to other URLWrappers via:

source: @avaris 
```
try: 
  return other.slug
except AttributeError:
  return slugify(blabla)                                       
```

and maybe to strings as well as those comparisons are often done in themes  `{% if cat == "bla" %}`

ingwinlu:
* speed up via reduced slugify calls (only call when needed)
* fix `__repr__` to not contain str, should call repr on name
* add test_urlwrappers and move URLWrappers tests there
  * add new equality test
* cleanup header

additionally:
* Content is now decorated with python_2_unicode_compatible
  instead of treating `__str__` differently (more inline with other
  pelican objects)
* better formatting for test_article_metadata_key_lowercase
  to actually output the conflict instead of a non descriptive
  error

closes #1758
ref #1493

### DIFF
--- a/pelican/urlwrappers.py
+++ b/pelican/urlwrappers.py
@@ -1,7 +1,9 @@
-import os
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
 import functools
 import logging
-
+import os
 import six
 
 from pelican.utils import (slugify, python_2_unicode_compatible)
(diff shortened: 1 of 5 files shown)


pelican/urlwrappers.py
```diff
@@ -52,27 +54,30 @@ def as_dict(self):
...
-    def _key(self):
-        return self.slug
-
     def _normalize_key(self, key):
         subs = self.settings.get('SLUG_SUBSTITUTIONS', ())
         return six.text_type(slugify(key, subs))
 
     def __eq__(self, other):
-        return self._key() == self._normalize_key(other)
+        if isinstance(other, URLWrapper):
```
kylef: Perhaps this would be better as `if isinstance(other, self.__class__):` that way it will verify it's the correct type.

That way these will be different:

```python
Category('test') != Tag('test')
```

ingwinlu: while the point you are trying to make is valid, that differs from how pelican is currently behaving:
```

```
more: it would still evaluate to true as the normalization will return true later, although slower

justinmayer:
@avaris / @kylef: Any comments on Winlu's latest changes?

justinmayer:
Many thanks for the enhancement, Winlu!

PR #1766, opened 2015-06-19
Tox: keep docs output

ingwinlu:
This modifies Tox's docs environment to keep the cache in a temporary environment (to force rebuilds), but keeps the output available under `docs/_build/html` as if you would have run `make html` in the docs directory.


justinmayer:
Thanks, Winlu!

PR #1767, opened 2015-06-19
Remove PAGES; use pages instead

ingwinlu:
* remove PAGES from context as pages is available
* add section to FAQ to provide guidance

as discussed in #1760 this removes `PAGES` and adds a FAQ entry explaining the issue.


justinmayer:
Thanks for taking care of this, Winlu.

PR #1769, opened 2015-06-20
separated argparse to be passed from main to prepare function

sanfx:
To be able to post to a pelican blog one would have to import pelican but their was no straight way to pass arguments, so instead of replicating a command line action via subprocess, I have added a prepare function that accepts either argpase.ArgumentParser or mocked argparse to pass the value to variable like settings etc that are command line flags.

### DIFF
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -386,7 +386,8 @@ def main():
 
 
 def prepare(args):
-    """ preapres settings received from an argparse or mocked argparse from         different application that imports pelican.
+    """ preapres settings received from an argparse or mocked argparse
+        from different application that imports pelican.
 
     	:param args: argparse.ArgumentParser or mocked argparse Class object
 	:type args: argparse.ArgumentParser or mocked argparse
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -382,6 +382,15 @@ def get_instance(args):
 
 def main():
     args = parse_arguments()
+    prepare(args)
+
+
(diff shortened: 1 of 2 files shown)


ingwinlu:
* version bumping is not your responsibility and has no part in this pr
* the function you call prepare actually does not just 'prepare', it runs the whole process

I am also not sure what you are trying to accomplish here

sanfx:
I named the function prepare because you already had run function, I was trying to write an app that automatically do post to blog, but not using subprocess to call pelican as you have it pelican command line , So to prepare function we can pass either third part app argparse or Mocked Argparse to publish post by importing pelican.

setup.py
```diff
@@ -21,7 +21,7 @@
 
 setup(
     name="pelican",
-    version="3.6.1.dev",
+    version="3.7.0.dev",
```
sanfx: you can take care of this version update yourself.

pelican/__init__.py
```diff
@@ -382,6 +382,16 @@ def get_instance(args):
 
 def main():
     args = parse_arguments()
+    prepare(args)
+
+
+def prepare(args):
```
sanfx: so if some imports pelican from command line from his python app , I can just
    import pelican #then
    pelican.prepare() # prepare in turn runs the run function.

ingwinlu:
no, this is terrible; if you want to run from python, import the pelican class, setup your settings there and run its run method, you don't hack functions into the script entry points

Issue #1572, opened 2014-12-30
Pagination breaks in default notmyidea template when last page contains a single item


ehashman:
Encountered this bug when I enabled pagination and the last page contained a single item. When this occurs, the page structure breaks, displaying stray footer elements floating around the page. I am using Pelican v3.5.0; from my investigation below, I've confirmed the bug still exists in master.

**Steps to reproduce:** start a new Pelican website with 2 articles, set `DEFAULT_PAGINATION = 1`, and view the second page (likely index2.html).

**Fix:** In the default notmyidea template, we have three [`loop.length > 1` checks](https://github.com/getpelican/pelican/blob/master/pelican/themes/notmyidea/templates/index.html#L45-L51). If `loop.length == 1` and the item wasn't the first in the loop, these checks will fail; thus, closing tags never get inserted, breaking the page.

If loop.length is non-zero, `if loop.last` should always return true, so I believe we can simply remove the `> 1` checks. I can submit a PR [(see diff)](https://github.com/ehashman/pelican/compare/issue1572), but I'm wondering if I should confine it to this change or if you'd prefer that I first check that other instances of this bug don't exist.

ehashman:
I'm not sure if the fix above actually works (iirc it broke the tests) but here's an example of this bug out in the wild, on Pelican 3.6.0: http://wics.uwaterloo.ca/tags/big-csters/2/

justinmayer:
In my testing earlier this week, this fix did indeed address the problem.

justinmayer:
This should be addressed in 3.6.1.

PR #1770, opened 2015-06-21
Fixes #1068, #1572: template page navigation bugs

### LINKED ISSUE #1068: Pagination shown when no need in notmyidea example theme

ehashman:
Updates the template logic for when page navigation is included in the
generated HTML, fixing:

* Issue #1068: useless pagination controls should not be displayed when a
  single page is generated (i.e. "Page 1/1"). New logic prevents the
  generation of these superfluous page navigation controls. Tests updated
  accordingly.

* Issue #1572: when multiple pages are generated and the last page contains
  only one item, the closing </ol> and </section> tags are not generated,
  resulting in page breakage. We need to check if
  articles_page.has_other_pages(); if it does, a list has been generated per
  line 19 or 25 and the tags must be closed.

### DIFF
--- a/pelican/tests/output/custom/category/bar.html
+++ b/pelican/tests/output/custom/category/bar.html
@@ -51,9 +51,6 @@ YEAH !</p>
 <img alt="alternate text" src="../pictures/Sushi.jpg" style="width: 600px; height: 450px;" />
 </div>
 <p>There are <a href="../oh-yeah.html#disqus_thread">comments</a>.</p>                </article>
-<p class="paginator">
-    Page 1 / 1
-</p>
             </aside><!-- /#featured -->
         <section id="extras" class="body">
                 <div class="blogroll">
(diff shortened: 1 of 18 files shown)


ehashman:
Hm, some tests seem to be failing---they might have been ones that got skipped when I tested on my local system. Looks like they're just expecting pagination when #1068 doesn't want it, so I'll try updating those too and see if that fixes everything.

ingwinlu:
should `simple` have the same behaviour?

ehashman:
I can look into adding that if that's what you want.

justinmayer:
Many thanks, Elana. Great to finally address this issue. Much appreciated.

PR #1774, opened 2015-06-28
Make generating various content optional.

alexras:
In some cases, I'd rather not generate archives, tags, categories or
authors. This patch provides flags for turning their generation off
selectively.


alexras:
Going to open this off another branch. Whoops!

justinmayer:
You might want to hold off re-submitting this, because I'm not sure it is necessary. See: Docs > Settings > URL Settings

Specifically:

    *_SAVE_AS = ''

... where * is what you don't want generated.

PR #1775, opened 2015-06-28
Make generating various content optional.

alexras:
In some cases, I'd rather not generate archives, tags, categories or
authors. This patch provides flags for turning their generation off
selectively.


ingwinlu:
http://pelican.readthedocs.org/en/latest/settings.html#url-settings
```
If you do not want one or more of the default pages to be created (e.g., you are the only author on your site and thus do not need an Authors page), set the corresponding *_SAVE_AS setting to '' to prevent the relevant page from being generated.
```

justinmayer:
@alexras: Given that this can already be accomplished (as Winlu noted here and [I noted in your other PR](https://github.com/getpelican/pelican/pull/1774#issuecomment-116252021)), do you still see a need for the settings you've proposed?

alexras:
Good to know. Thanks! Closing this then.

PR #1618, opened 2015-02-08
No tests in bdist

ionelmc:
Discussion from IRC:

```
18:38 <ionelmc> can you stop including the tests in the installed package?
18:39 <+justmay> Fine with me. Care to submit a PR that addresses that?
18:39 <ionelmc> sure
18:47 <ionelmc> justmay: is `.mailmap` a stray file?
18:47 <ionelmc> look like your sdist is incomplete
18:48 <ionelmc> doesn't include all the docs and test conf
18:50 <ionelmc> justmay: oh snap. i have to move out the tests to fix this :-)
18:51 <ionelmc> because include_package_data is used any included data file overlaying the package is going into the bdist
18:51 <+justmay> ionelmc: .mailmap is there by design. See "Mapping Authors": https://www.kernel.org/pub/software/scm/git/docs/git-shortlog.html
18:52 <ionelmc> mkay
18:52 <ionelmc> justmay: you're not going to have a problem with the tests dir at the same level as pelican package right?
18:53 → e-Flex joined (~e-Flex@h-62-241.a259.priv.bahnhof.se)
18:54 <+justmay> There's no other way to prevent inclusion of tests in the installed package?
18:55 <ionelmc> justmay: there are two horrible ways
18:55 <ionelmc> don't include it in the sdist (highly undesirable)
18:55 <ionelmc> or
18:55 <ionelmc> manually specify package_data 
18:55 <ionelmc> which i can do it correctly for you know, but it will be error prone to maintain
18:56 <ionelmc> s/know/now/
18:56 <ionelmc> i think that's also not desirable
18:56 <ionelmc> that's why i think moving them out is ok
18:57 <ionelmc> i'll fix the test configuration to work that way
18:57 <ionelmc> justmay: agree? :-)
18:59 <+justmay> ionelmc: Quite honestly, I don't have the bandwidth this morning to dig deeply enough into this topic. Would you submit an issue so we (i.e., the community) can discuss this and come to a consensus?
19:00 <ionelmc> justmay: there's already https://github.com/getpelican/pelican/issues/1409 - i seriously doubt a new issue will help in any way
19:01 <winlu> ionelmc: justs prune tests and be done with it
19:01 <ionelmc> justmay: it's either the relocation or manual package_data, make a choice :-)
19:01 <ionelmc> winlu: pruning the tests will remove them from sdist
```

Closes #1609. Closes #1545. Closes #1409.

PR #1778, opened 2015-07-09
Avoid adding tests in bdists (like wheel bdist)

ionelmc:
Supersedes and closes #1618.


justinmayer:
Seems like a sensible solution to me. Many thanks for fixing this, Ionel!

Issue #1787, opened 2015-07-28
truncate_html_words() fails to truncate content when tags span multiple lines


andreacorbellini:
I have an article with the following content:

    <strong>text text text <span                                              
    style="font-family:serif;white-space:nowrap;"> text text text</strong>

and I've set `SUMMARY_MAX_LENGTH` to a non-zero value.

When the article gets truncated by Pelican, here's the markup I get:

    <strong>text text text <span
    style="font-family:serif ...</strong></p>

As you can see, Pelican believes that `style="font-family:serif;white-space:nowrap;"` is text and truncates it without correctly closing the `<span>`, resulting in a broken web page.

The problem occurs because the regular expression `re_words` does not look for newline characters. This can be fixed by using the `DOTALL` option:

    re_words = re.compile(r'&.*?;|<.*?>|(\w[\w-]*)', re.U | re.DOTALL)

[This can also be fixed by using a HTML parser, but I guess you already know that :)]

iKevinY:
Looks like this issue can be closed since #1788 has been merged in. /cc @justinmayer

justinmayer:
This issue has been addressed by @andreacorbellini via the above-mentioned pull request. (Thanks for the nudge, Kevin!)

PR #1788, opened 2015-07-29
Truncate words using a HTML instead of regular expressions

andreacorbellini:
This is a fix for issue #1787.

I've decided to rewrite `truncate_html_words()` in order to parse HTML with a proper parser, instead of regular expressions.

Note that the tests I've added do not pass with the old `truncate_html_words()`.

### DIFF
--- a/pelican/utils.py
+++ b/pelican/utils.py
@@ -24,6 +24,7 @@ from itertools import groupby
 from jinja2 import Markup
 from operator import attrgetter
 from posixpath import join as posix_join
+from six.moves.html_parser import HTMLParser
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


pelican/utils.py
```diff
@@ -25,6 +25,11 @@
 from operator import attrgetter
 from posixpath import join as posix_join
 
+try:
```
ingwinlu: Pelican uses the `six` module where possible:

```from six.moves.html_parser import HTMLParser```

ingwinlu:
except for my one find this looks descend though I would prefer to hear feedback from others as well.

Keep in mind that commits in PR's should be squashed when you are finished working on them.

andreacorbellini:
I've created another branch to handle expressions like `it's` and `caf&eacute;` (= `café`). Both expressions are currently counted as two words (and therefore may be truncated), and both expressions are counted as one word in my new branch. The branch is:

https://github.com/andreacorbellini/pelican/tree/better-word-recognition

I've kept it separate from this PR to make the diff shorter and easier to understand, but if you prefer a single PR please tell me and I'll merge+squash everything.

justinmayer:
Hi Andrea. Many thanks for the improvement. It seems Alexis wrote that code two years prior to the [legendary treatise on parsing HTML with regex](http://stackoverflow.com/questions/1732348/regex-match-open-tags-except-xhtml-self-contained-tags/). :wink: 

No need to include the better-word-recognition branch in this PR; I agree that's better-suited for its own pull request. As Winlu suggested, however, the three commits in this PR form a discrete whole and should probably be squashed. Would you mind taking care of that?

Thanks again for the enhancement!  (^_^)

andreacorbellini:
There you go. Squashed everything :)

justinmayer:
@andreacorbellini: Not sure how you did the rebase+squash, but it seems like there are some spurious unrelated commits included. Think you can sort that out?

andreacorbellini:
Ups, sorry for that. Fixed!

justinmayer:
Thanks for the contribution, Andrea.  (^_^)

PR #1790, opened 2015-07-31
Switch from Coveralls to Codecov

treyhunner:
**Full disclosure**: I work for Codecov

A few reasons pelican should be using Codecov:
- Codecov supports branch coverage (e.g. Codecov [caught this missing else branch](https://codecov.io/github/treyhunner/pelican/pelican/tools/pelican_import.py?ref=e348964d04564d456b2ab76d3cb16cf9408c6b49#l-523) while Coveralls [did not](https://coveralls.io/builds/3183205)).
- Codecov has a browser plugin for display coverage right from GitHub ([example](http://i.imgur.com/YKW8P13.png))
- The reports are easier to navigate. [Here's an example report for pelican](https://codecov.io/github/treyhunner/pelican/pelican?ref=e348964d04564d456b2ab76d3cb16cf9408c6b49).

Branch coverage is the reason I originally started using Codecov.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -59,9 +59,9 @@ Why the name "Pelican"?
 .. |build-status| image:: https://img.shields.io/travis/getpelican/pelican/master.svg
    :target: https://travis-ci.org/getpelican/pelican
    :alt: Travis CI: continuous integration status
-.. |coverage-status| image:: https://img.shields.io/coveralls/getpelican/pelican.svg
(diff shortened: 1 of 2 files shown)


justinmayer:
Many thanks for the contribution, Trey, and apologies for the absurdly long delay in reviewing and responding. Given that we removed Coveralls, I think it's best to retire this pull request for now. Thanks again for the submission! 👍

PR #1793, opened 2015-08-07
Add missing *.markdown files to PyPI sdist

koobs:
The following file is missing from the PyPI source distribution (sdist) due to *.markdown not being referenced in MANIFEST.in:

pelican/tests/content/article_with_markdown_extension.markdown

This missing file appears to cause a number of errors running the test suite. An alternative to this change would be to rename the file to .md so that the file is covered by the existing *.md entry.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,3 +1,3 @@
 include *.rst
-recursive-include pelican *.html *.css *png *.in *.rst *.md *.mkd *.xml *.py
+recursive-include pelican *.html *.css *png *.in *.rst *.markdown *.md *.mkd *.xml *.py
 include LICENSE THANKS docs/changelog.rst


iKevinY:
Looks good to me. I believe the `.markdown` extension is intentional (as it's [technically](http://daringfireball.net/linked/2014/01/08/markdown-extension) an [accepted](https://github.com/github/markup/blob/b865add2e053f8cea3d7f4d9dcba001bdfd78994/lib/github/markups.rb#L1) file extension), so this seems to be the correct way to go about solving the issue. Thoughts, @justinmayer?

justinmayer:
Many thanks to Kubilay for the contribution and to Kevin for reviewing.  (^_^)

koobs:
<3

PR #1795, opened 2015-08-08
Add warning for unknown replacement indicators

### LINKED ISSUE #1794: output warning during unknown content replacement
https://github.com/getpelican/pelican/blob/master/pelican/contents.py#L215

output a warning if what is invalid and no substitutions occur because of it

ingwinlu:
If an unknown replacement indicator {bla} was used, it was ignored
silently. This commit adds a warning when an unmatched indicator occurs
to help identify the issue.

closes #1794


justinmayer:
Thanks, Winlu!

PR #1797, opened 2015-08-10
Add optional override_output to write_feed

TC01:
The write_feed function now takes an override_output parameter that does the same thing as override_output does to write_file. Implemented for custom generators.

This is a very minor change; all the unit tests still pass as no surface behavior was changed.

As for why I wanted/needed this... I was working a plugin that implemented a custom generator (that fetches articles from mailboxes, which is why it had to be implemented as a generator and not a reader). I ran into an interesting potential issue: among other things, author names could potentially collide between articles in ```content/``` and "articles" produced from the mailbox. It was easy enough to fix this by passing ```override_output=True``` to ```write_file```, but since such an option didn't exist for ```write_feed``` it wasn't possible to overwrite the feeds.

I could imagine other plugin developers potentially needing or wanting this functionality too.

### DIFF
--- a/pelican/writers.py
+++ b/pelican/writers.py
@@ -79,7 +79,8 @@ class Writer(object):
         self._written_files.add(filename)
         return open(filename, 'w', encoding=encoding)
 
-    def write_feed(self, elements, context, path=None, feed_type='atom'):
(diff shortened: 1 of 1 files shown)


justinmayer:
Thanks for the enhancement, Ben.  (^_^)

PR #1800, opened 2015-08-17
Added a note in the docs about metadata variablenames being lowercase…

kbotnen:
…d before prosessed

### DIFF
--- a/docs/content.rst
+++ b/docs/content.rst
@@ -68,6 +68,10 @@ Metadata syntax for Markdown posts should follow this pattern::
 
     This is the content of my super blog post.
 
+.. note::
+
+   Remember that all metadata variablenames are lowercased before they are prosessed by Pelican.
+
 Readers for additional formats (such as AsciiDoc_) are available via plugins.
 Refer to `pelican-plugins`_ repository for those.
 


ingwinlu:
i see no value in adding this note, it is a note targeted at developers in the middle of user documentation. then one step further, if you are a dev, why not just look at the code?

additionally:
* the place you inserted it seems strange, why under the markdown metadata 
* having a note kind of makes the 'remember' intro pointless
* metadata keys, not variablenames (metadata is a dict internally)
* metadata is getting lowered as is is processed, not before it is being processed by pelican

kbotnen:
Hi,

The note is not targeted at developers, but to new users lime myself.

The motivation comes from a buglike experienced I had when trying to create a metadata variable that I could use to sort articles.

In my config file i had

<pre>
PAGE_ORDER_BY = 'Sortorder'
</pre>
and in my article.md i had
<pre>
Sortorder: 60
</pre>

which did not work. The reason I used uppercase 'S' in 'Sortorder' is because the example documentation of metadata in a markdown file all use uppercase variablenames.

But when I did the following in my configfile
<pre>
PAGE_ORDER_BY = 'sortorder'
</pre> 
and no change in the metadata in the pagenames.md files, it worked.

If the documentation had been clear on the fact that metadata is downcased I would have seen the cause of the problem after reading the documentation, and not just by accidently trying random stuff.

And as a sidenote, it is already documented, but in a imo odd place:  http://docs.getpelican.com/en/3.6.2/themes.html?highlight=lowercase#article-html

The reason why I think this is an odd place is that it is in the theme section of the documentation, but I was working with the config file and articles while I stumbled upon the issue.

So that was a bit of background information. I could agree or not agree if it should be a note, or just normal sentence. About the technical details about dicts and where its prosessed etc, I stand corrected. Im a user, not a developer.

ingwinlu:
there should be a warning printed when using an invalid string, still need to check if we test if that actually works; the sort by also should .lower strings to prevent that kind of behaviour

ingwinlu:
just had another look at it, we do indeed output a warning if the set sorting attribute does not exist:
`WARNING: There is no "does-not-exist" attribute in the item metadata. Defaulting to slug order.`

justinmayer:
Try as I might, I couldn't find any wording that made a lot of sense to convey this, nor could I find  an appropriate place to put said notice. Given that this hasn't come up since, I think it might be an edge case that doesn't require addressing at this time.

That all said, thank you Kristian for taking the time to submit a pull request. I truly appreciate the intention and care you took to do this. 👍

PR #1804, opened 2015-08-19
Add option to let pelican CLI run a server in regenerate mode

lionel-panhaleux:
This avoids a very ugly "save PID for shutdown" technique and the pelican regenerated server can run single-handedly (not in background).

I did not see any issue on this subject, but can open one if needed.
Commentaries / requests for improvements are welcome.


avaris:
Either the title or your implementation is wrong. The way it's implemented right now, server has no correlation to the regenerate mode. It's run whenever you give the `--serve`/`-S` option, regardless of the `--autoreload`/`-r` option being present or not.

lionel-panhaleux:
It has a relation in the sense that if you do not use the -r option, you won't have an infinite loop and the server won't run for very long ;) I can put the server option inside the -r option block to make this more explicit if it is preferred on your side

avaris:
I don't have a preference actually :). I'm just saying that the title and commit message doesn't match the implementation. Either do what you said, or say what you did. Otherwise it'll be confusing later on.

lionel-panhaleux:
I'll upload a new, more complete pull request in a minute

ingwinlu:
I would prefer you not opening new pr's for the same feature, instead using git {branch, rebase, reset, ...} and the edit button to archive the same effect. See the contribution guidelines (which even pops up when you crate an issue or pr) if you need more help.

This keeps the discussion in one place and does not spam everyone with emails.

PR #1813, opened 2015-09-04
When truncating, stop parsing the document immediately after finding the last word

andreacorbellini:
This commit makes the `_HTMLWordTruncator` faster by stopping it immediately as soon as the last word is found. This should improve performances for blogs that have many articles.

This is the last PR of this series.

### DIFF
--- a/pelican/utils.py
+++ b/pelican/utils.py
@@ -414,6 +414,13 @@ class _HTMLWordTruncator(HTMLParser):
     _singlets = ('br', 'col', 'link', 'base', 'img', 'param', 'area',
                  'hr', 'input')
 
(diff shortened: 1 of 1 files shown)


ingwinlu:
logically looks good.

from a code readability standpoint I would suggest to add the return value as an attribute to the exception (minor details but it is more self describing than `.args[0]`)

justinmayer:
Hi Andrea. Any thoughts on Winlu's suggestion here?

andreacorbellini:
Sorry, I'm busy right now. I will fix it this weekend :)

andreacorbellini:
A bit late, but finally here I am.

Issue #1733, opened 2015-05-26
Wrong iterating keywords, description in simple theme.


dmgl:
Name: pelican
Version: 3.5.0

Set
     
    THEME = 'simple'

in configuration file pelicanconf.py

In template 

    themes/simple/templates/article.html

https://github.com/getpelican/pelican/blob/master/pelican/themes/simple/templates/article.html

````
{% for keyword in article.keywords %}
    <meta name="keywords" content="{{keyword}}" />
{% endfor %}
````

My article have:

````
Keywords: test1, test2
````

i have wrong iteration by chars, not by words.

Output  is:

````
<meta name="keywords" contents="t" />
<meta name="keywords" contents="e" />
<meta name="keywords" contents="s" />
<meta name="keywords" contents="t" />
<meta name="keywords" contents="1" />
<meta name="keywords" contents="," />
<meta name="keywords" contents="t" />
<meta name="keywords" contents="e" />
<meta name="keywords" contents="s" />
<meta name="keywords" contents="t" />
<meta name="keywords" contents="2" />
````

If replace with

````
{% if article.keywords %}
<meta name="keywords" content="{{ [article.keywords]|join(', ') }}" />
{% endif %}
````

problem solved.

Output:

````
<meta name="keywords" content="test1, test2" />
````

The same situation with:

````
  {% for description in article.description %}
    <meta name="description" content="{{description}}" />
  {% endfor %}
````

Вut for tags another behavior

````
  {% for tag in article.tags %}
    <meta name="tags" content="{{tag}}" />
  {% endfor %}
````

Output:

````
<meta name="tags" contents="test1" />
<meta name="tags" contents="test2" />
````

Please, explain such behavior, if it possible, don't understand. Is this problem in jinja template?

ingwinlu:
keywords does not get the same treatment as metadata fields that are expected to be a list (tags, authors,...), so it just reads the line as one string.

this results in jinja iterating over a string, on each char.

dmgl:
Ok, i understand that it's iterating on each char. I thought it was a mistake, wrong behavior. Thanks for your comment.

PR #1814, opened 2015-09-05
Remove article.keywords from simple theme

ingwinlu:
We don't process 'keywords' metadata specially, so it never gets
processed into a list.

This prevents issues like #1733.


justinmayer:
Thanks for the fix, Winlu.

PR #1819, opened 2015-09-11
Add versioning information to debug output

ingwinlu:


justinmayer:
Thanks for this addition, Winlu. I'm sure it will come in handy. (^_^)

PR #1823, opened 2015-09-14
Add index and author replacement indicators

onur:
There is no easy way to make a link to index of website created with pelican. This patch is aiming to fix this issue. With this patch we can create links to our index with `[Link]({index})`.

Patch also adds author replacement indicator. We can create links to author pages with: `[Link]({author}Someone)`.

### DIFF
--- a/pelican/contents.py
+++ b/pelican/contents.py
@@ -21,7 +21,7 @@ from pelican.utils import (SafeDatetime, deprecated_attribute, memoized,
                            slugify, strftime, truncate_html_words)
 
 # Import these so that they're avalaible when you import from pelican.contents.
-from pelican.urlwrappers import (URLWrapper, Author, Category, Tag)  # NOQA
+from pelican.urlwrappers import (Author, Category, Tag, URLWrapper)  # NOQA
 
 logger = logging.getLogger(__name__)
 
@@ -253,6 +253,10 @@ class Content(object):
                 origin = '/'.join((siteurl, Category(path, self.settings).url))
             elif what == 'tag':
                 origin = '/'.join((siteurl, Tag(path, self.settings).url))
+            elif what == 'index':
+                origin = '/'.join((siteurl, self.settings['INDEX_SAVE_AS']))
+            elif what == 'author':
+                origin = '/'.join((siteurl, Author(path, self.settings).url))
             else:
                 logger.warning(
                     "Replacement Indicator '%s' not recognized, "
--- a/pelican/tests/test_urlwrappers.py
+++ b/pelican/tests/test_urlwrappers.py
@@ -2,7 +2,7 @@
 from __future__ import unicode_literals
 
 from pelican.tests.support import unittest
-from pelican.urlwrappers import Category, Tag, URLWrapper
+from pelican.urlwrappers import Author, Category, Tag, URLWrapper
 
 
 class TestURLWrapper(unittest.TestCase):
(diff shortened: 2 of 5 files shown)


ingwinlu:
* index might not be at index.html, there is a reason we have a setting for this
* missing documentation
* missing tests

onuraslan:
Yes you are right, I am gonna fix this

pelican/contents.py
```diff
@@ -21,7 +21,7 @@
                            slugify, strftime, truncate_html_words)
 
 # Import these so that they're avalaible when you import from pelican.contents.
-from pelican.urlwrappers import (URLWrapper, Author, Category, Tag)  # NOQA
+from pelican.urlwrappers import (URLWrapper, Author, Category, Tag, Index)  # NOQA
```
ingwinlu: order should be alphabetical (NOQA suppresses this, sadly necessary to get rid of the unused imports warning I guess)

pelican/urlwrappers.py
```diff
@@ -122,3 +122,7 @@ def __init__(self, name, *args, **kwargs):
 
 class Author(URLWrapper):
     pass
+
+
+class Index(URLWrapper):
+    pass
```
ingwinlu: what?

pelican/tests/test_urlwrappers.py
```diff
@@ -48,5 +52,13 @@ def test_equality(self):
...
 
+        # Author describing the same should be equal
+        author_equal = Author('Test', settings={})
+        self.assertEqual(author, author_equal)
+
+        # Index describing the same should be equal
+        index_equal = Index('Test', settings={})
+        self.assertEqual(index, index_equal)
+
         cat_ascii = Category('指導書', settings={})
```
ingwinlu: those tests don't provide anything regarding the correctness of your code

pelican/urlwrappers.py
onuraslan: `URLWrapper._from_settings` is using `self.__class__.__name__.upper()` and key to generate setting name. In this case we need INDEX_SAVE_AS setting value. That is also reason  we have:
```python
class Category(URLWrapper):
    pass


class Author(URLWrapper):
    pass
```

pelican/tests/test_urlwrappers.py
onuraslan: I agree but I will add more test into test_contenst.py.

onuraslan:
There is an issue with this patch. It doesn't work if `INDEX_SAVE_AS` not defined in conf. Looks like there is no `DEFAULT_CONFIG` for `INDEX_SAVE_AS` setting.

pelican/urlwrappers.py
ingwinlu: I know, the problem is index is (logically) not an URLWrapper (right now), because of the reason you just described.

You can directly pull the save_as setting, which makes more sense.

If we would revisit how we handle `DIRECT_TEMPLATES` we could add them as URLWrappers, but that is not the issue that you are trying to solve here.

ingwinlu:
indeed it is missing (it currently gets constructed in the generator instead of read from a default setting).

you can add it and set it to the default value that is described in the settings.rst file.

ingwinlu:
looking better, still missing some documentation and a squash of the commits when you are done!
Keep it up!

onur:
Squashing is done, not sure about missing docs

ingwinlu:
missed the added docs, oops.

looks good to me, the author and co replacment tests are meh but we are currently not doing anything better for the other replacements as well.

justinmayer:
Thanks for the contribution, Onur.  (^_^)

PR #1825, opened 2015-09-17
Make PagesGenerator more generic

LeResKP:
It's usefull to be able to inherit from PagesGenerator to generate many kind of pages.

### DIFF
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -182,8 +182,17 @@ class Pelican(object):
 
         articles_generator = next(g for g in generators
                                   if isinstance(g, ArticlesGenerator))
-        pages_generator = next(g for g in generators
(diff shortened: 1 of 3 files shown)


avaris:
The place you create the signals is too late. Did you try to connect to those signals you replaced (`page_generator_init` etc.) in a plugin? I seriously doubt that you'll succeed.

justinmayer:
Hi @LeResKP. Any comments on the feedback from @Avaris?

LeResKP:
Sorry didn't take time to reply and to find a good solution. I think @avaris is right and I need to update the code. I think I have to create a metaclass or perhaps you have a better idea?
Also we need to write more tests, It's strange nothing breaks with this commit...

PR #1826, opened 2015-09-20
Changed rebuild to preview in publish to GitHub pages command in fabfile.py

lazycoder-ru:
I think it wrong to build not-publish version before publishing site.


justinmayer:
Thank you for the contribution and apologies for the delay in responding. It seems this change is already present in `tasks.py`.

PR #1827, opened 2015-09-21
Make name of blogroll/social widgets configurable

TC01:
The ```BLOGROLL_WIDGET_NAME``` and ```SOCIAL_WIDGET_NAME``` settings are now respected by notmyidea if they are specified in your config file. They override the default names of "blogroll" and "links" in the notmyidea theme.

I'm sure I'm not the only one who would prefer to call the "Blogroll" widget "Links"; I know that some themes (at the very least pelican-bootstrap3) do this by default, and obviously it's easy enough to change on any theme, but I think it would make more sense if it was a configurable option that themes would respect...

So I figured I would implement this and send a pull request at the main pelican repository for the notmyidea theme, and if it's agreed that this is a useful feature, perhaps we can update some of the themes in pelican-themes too?

If ```BLOGROLL_WIDGET_NAME``` and ```SOCIAL_WIDGET_NAME``` aren't defined, the theme uses the default option-- in this case, "blogroll" and "social".

The change is documented in settings.rst.

### DIFF
--- a/THANKS
+++ b/THANKS
@@ -24,6 +24,7 @@ Andrew Spiers
 Arnaud BOS
 asselinpaul
 Axel Haustant
+Ben Rosser (TC01)
 Benoît HERVIER
 Borgar
 Brandon W Maister
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -724,6 +724,10 @@ Setting name              What does it do?
 ``TWITTER_USERNAME``      Allows for adding a button to articles to encourage
                           others to tweet about them. Add your Twitter username
                           if you want this button to appear.
+``LINKS_WIDGET_NAME``     Allows override of the name of the links widget.
+                          If not specified, defaults to "links".
+``SOCIAL_WIDGET_NAME``    Allows override of the name of the "social" widget.
+                          If not specified, defaults to "social".
 =======================   =======================================================
 
 In addition, you can use the "wide" version of the ``notmyidea`` theme by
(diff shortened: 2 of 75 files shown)


ingwinlu:
`{{ my_variable|default('my_variable is not defined') }}` and you don't need the if's

TC01:
Thanks for the suggestion; fixed, committed, and rebased. :)

justinmayer:
I agree with your motivation, Ben. In keeping with the idea that Pelican can be used for more than blog sites, perhaps we should call the new setting `LINKS_WIDGET_NAME` (instead of `BLOGROLL_WIDGET_NAME`).

As an ancillary (and totally unintended) side benefit, that means the relevant settings table in the docs no longer needs to be widened.  (^_^) 

Apologies for adding my thoughts after you've already made a round of changes.

TC01:
Not a problem. :) I like that idea; would we perhaps also want to rename blogroll to links in notmyidea, then?

justinmayer:
Yup, seems reasonable to me.

justinmayer:
Hey Ben. Any updates on your pull request here?

TC01:
Whoops. Just pushed an update, sorry for the delay!

TC01:
Oops. Looks like the tests need updating too, in order to check for "links" and not "blogroll".

justinmayer:
Thanks, Ben. Would you please [rebase & squash your commits](https://github.com/getpelican/pelican/wiki/Git-Tips#squashing-commits)?

TC01:
Done. :)

justinmayer:
Thanks, Ben!

PR #1832, opened 2015-10-01
Use <time> instead of <abbr>

meisterluk:
The pubdate given is in no way an abbreviation. Hence the semantics of
``<abbr>`` do not justify its usage in this context.

``<time>`` was introduced in HTML5. It's datetime attribute can override the
textContent. Anyway it specifies a date+time of a given event.

http://www.w3.org/TR/html5/text-level-semantics.html#the-time-element

Originally the pubdate attribute was meant to annotate that the
publication date of the closest section unit is specified, but this was
dropped and is not part of the official specification.

https://www.w3.org/wiki/HTML/Elements/time

### DIFF
--- a/pelican/themes/simple/templates/article.html
+++ b/pelican/themes/simple/templates/article.html
@@ -21,13 +21,13 @@
  {{ translations.translations_for(article) }}
   </header>
   <footer class="post-info">
-    <abbr class="published" title="{{ article.date.isoformat() }}">
+    <time class="published" datetime="{{ article.date.isoformat() }}">
       {{ article.locale_date }}
-    </abbr>
+    </time>
 	{% if article.modified %}
-    <abbr class="modified" title="{{ article.modified.isoformat() }}">
+    <time class="modified" datetime="{{ article.modified.isoformat() }}">
       {{ article.locale_modified }}
-    </abbr>
+    </time>
 	{% endif %}
     {% if article.authors %}
     <address class="vcard author">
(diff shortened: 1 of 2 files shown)


meisterluk:
Sorry for the second fix. Leave me a hint if I should squash the messages and offer a new pull request.

meisterluk:
Oh, I just found a small discussion in the original Smashing Magazine blog article where they discussed the design:

http://www.smashingmagazine.com/2009/08/designing-a-html-5-layout-from-scratch/#comment-178185

ingwinlu:
shouldn't those commits trigger a test failure?

meisterluk:
@ingwinlu can you point out which "test_*.py" runs equivalence tests with some expected output? It's difficult for me to go through the whole codebase.

kylef:
@ingwinlu I believe the tests are against notmyidea theme not the simple theme. The fixtures are at https://github.com/getpelican/pelican/tree/master/pelican/tests/output

justinmayer:
Hi @meisterluk. Thanks for the contribution and for offering to squash your commits. Would you please do that so we can get this merged? (^_^)

meisterluk:
Thanks @justinmayer for the feedback. I opened another issue #1858 for the new branch / squashed messages.

ingwinlu:
please don't do that in the future, it spreads out discussion across multiple pr's instead of one.

meisterluk:
True, alright.

PR #1833, opened 2015-10-01
Added Opengraph data

JeremyMorgan:
Added additional opengraph data to default templates

### DIFF
--- a/pelican/tests/output/custom_locale/posts/2010/décembre/02/this-is-a-super-article/index.html
+++ b/pelican/tests/output/custom_locale/posts/2010/décembre/02/this-is-a-super-article/index.html
@@ -1,0 +1,129 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+        <meta charset="utf-8" />
+        <title>This is a super article !</title>
+        <link rel="stylesheet" href="../../../../../theme/css/main.css" />
+        <link href="http://blog.notmyidea.org/feeds/all.atom.xml" type="application/atom+xml" rel="alternate" title="Alexis' log Atom Feed" />
+        <link href="http://blog.notmyidea.org/feeds/all.rss.xml" type="application/rss+xml" rel="alternate" title="Alexis' log RSS Feed" />
+
+        <!--[if IE]>
+            <script src="https://html5shiv.googlecode.com/svn/trunk/html5.js"></script>
+        <![endif]-->
+</head>
+
+<body id="index" class="home">
+<a href="http://github.com/ametaireau/">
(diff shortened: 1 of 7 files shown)


justinmayer:
Thanks for the contribution, Jeremy. Looks like tests are failing. Would you please have a look and see if you can get that resolved?

justinmayer:
Hi @JeremyMorgan. Would you mind updating your pull request such that tests pass?

justinmayer:
@getpelican/reviewers: What do you think? Perhaps the [extended_meta plugin](https://github.com/kplaube/extended_meta) by @kplaube might be a better solution?

liltinkerer:
@justinmayer are the test failing because it compares the output against current theme?
Could you also please take a look at [#2036](https://github.com/getpelican/pelican/pull/2036), that is somewhat similar pull request.

justinmayer:
@liltinkerer: The most common reason for test failures is when folks neglect to read our [How to Contribute](http://docs.getpelican.com/en/latest/contribute.html#contributing-code) documentation. Specifically, any changes that modify how Pelican performs generation must also include updates to the functional test output.

My main question to @getpelican/reviewers is whether OpenGraph data should be provided via core or via plugins like [extended_meta](https://github.com/kplaube/extended_meta) by @kplaube or [pelican-open_graph](https://github.com/whiskyechobravo/pelican-open_graph) by @whiskyechobravo. Anyone have any thoughts on the matter?

avaris:
Well, this modifies the built in themes/templates to include OG data. A plugin won't help here. This is a two phase problem. First, handling/providing the OG data (which can be done via a plugin) and second, including this data in the output (which is theme/template related). So, the question becomes, should the built in themes include that?

Having said that, this PR has multiple issues as it stands. Includes `OGIMAGE` and `TWITTER_HANDLE` settings without any documentation for them. Has unnecessary duplication: e.g. `twitter:title` isn't needed if there is `og:title`.

liltinkerer:
@justinmayer Sorry about that. I guess I was too busy reading how to squash commits. Neglect is a bit harsh.

About current issue: As @avaris mentioned, Twitter already is using some of the OG tags.

This PR gives all the pages all the pages same image and title, which can be harmful since FB and Twitter and pretty good at picking up the best images. Which you probably care the most. And if they don't find OG tags, they default to good old title and meta description.

It would make more sense to read posts' metadata if you want custom titles or images. But then again you're better off writing a theme from scratch if you care about such things.

JeremyMorgan:
Sorry, these messages got buried. I will read the documentation for contributing, and yes this probably should be a plugin of some type rather than a default theme type add on and might be something I can work on soon.

justinmayer:
Many thanks for your contribution, Jeremy. Since this pull request never quite made it to the level we need it to be for inclusion into Pelican master, for now I think it's best to close this PR. If you would like to work on this endeavor further, please submit an issue so we can discuss.

PR #1837, opened 2015-10-04
Fix quote escaping in HTML attributes. Fixes issue #1260.

### LINKED ISSUE #1260: Quote escapes improperly handled when generating HTML articles
Say if I want to include JSON dataset in an HTML element say

```html
<section id="my-section" data-my-data='{"myJSON": "object"}'> ... <\section>
```

the output will be

```html
<section id="my-section" data-my-data="{"myJSON": "object"}"> ... <\section>
```

which is terrible. Note the significant change from single to double quotes.

Under most circumstances this isn't a problem, since reversing the quotes  (double quotes surrounding single quotes) will retain the orders, but that is not a valid JSON-object. It keys and strings *have* to be surrounded by double quotes.

I tried looking through the code base to fix it my self. But I have no idea where the transition happens.

Best regards,

SimonStJG:
Hello, I think this fixes #1260.  runarberg seemed to have already done most of the work in 2014, so I won't claim originality. But anyway, I hope it's useful to you :)

If an the value of an HTML attribute contains a double quote, we now escape with single quotes, otherwise we escape with double quotes.

### DIFF
--- a/pelican/readers.py
+++ b/pelican/readers.py
@@ -18,17 +18,13 @@ from pelican import rstdirectives  # NOQA
 from pelican import signals
 from pelican.cache import FileStampDataCacher
 from pelican.contents import Author, Category, Page, Tag
-from pelican.utils import SafeDatetime, get_date, pelican_open, posixize_path
+from pelican.utils import SafeDatetime, escape_html, get_date, pelican_open, \
+    posixize_path
 
 try:
     from markdown import Markdown
 except ImportError:
     Markdown = False  # NOQA
-
-try:
-    from html import escape
-except ImportError:
-    from cgi import escape
 
 # Metadata processors have no way to discard an unwanted value, so we have
 # them return this value instead to signal that it should be discarded later.
@@ -354,7 +350,7 @@ class HTMLReader(BaseReader):
                 self._in_body = False
                 self._in_top_level = True
             elif self._in_body:
-                self._data_buffer += '</{}>'.format(escape(tag))
+                self._data_buffer += '</{}>'.format(escape_html(tag))
 
         def handle_startendtag(self, tag, attrs):
             if tag == 'meta' and self._in_head:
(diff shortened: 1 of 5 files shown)


SimonStJG:
Okay, I've obviously got this a little bit wrong, since I've broken the build for Python 3.3.  Let me have a look and sort it out, sorry!

ingwinlu:
```
[winlu@virtualArch ~]$ python
Python 3.5.0 (default, Sep 20 2015, 11:28:25) 
[GCC 5.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from html import escape
>>> escape('"test"')
'&quot;test&quot;'
>>> 
[winlu@virtualArch ~]$ python2
Python 2.7.10 (default, Sep  7 2015, 13:51:49) 
[GCC 5.2.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from html import escape
  File "<stdin>", line 1, in <module>
ImportError: No module named html
>>> from cgi import escape
>>> escape('"test"')
'"test"'
>>> 
```

Both escapes take an optional 'quotes' kw that determines if quotes should be escaped as well for use in html attributes.
It defaults to false in cgi, it defaults to true in html

SimonStJG:
Thanks.  I've fixed the problem with the escaping of quotes.

justinmayer:
@ingwinlu: What do you think about Simon's latest changes here? Anything left to do?

ingwinlu:
the pr itself is fine, but I suggest making a wrapper function around the escape call that defaults to the same parameters regardless of python version before problems with it pop up.

SimonStJG:
I've made a wrapper function which defaults to quote=True and replaced all the other occurances of escape with it.

ingwinlu:
moving the escape function to utils makes sense as it is not locally bound to readers in any way

SimonStJG:
That makes sense.  I've moved it.

ingwinlu:
the docstring in `escape_html` should follow the [guideline](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings). Feel free to improve upon this suggested docstring instead:
```
"""Escape '&', '<' and '>' to HTML-safe sequences.

In Python 2 this uses cgi.escape and in Python 3 this uses html.escape. We wrap here to ensure the quote argument has an identical default.
"""
```

Your commit message also has a slight english hiccup. Additionally it should just be a summary in the first line, with additional explanation in other lines. The Message should also contain what you changed instead of just the reason to be more verbose.
Again, feel free to expand on this suggestion:
```
Fix quote escaping in read html attributes.

* Wrap HTML attributes in quotes according to their content
  If it contains a double quote use single quotes, otherwise escape with double quotes
* Add escape_html utility to ensure quote entities are converted identically across Python versions

Fixes #1260
```

Thanks for your contribution.

SimonStJG:
I've fixed the docstring and the commit message.  I wasn't aware of the docstring convention - I'm quite new to python, and as you may have noticed, this is my first attempt at a pull request.

Anyway, thanks for being patient.

justinmayer:
Many thanks to Simon for the contribution and to Winlu for the review and feedback.  (^_^)

PR #1842, opened 2015-10-13
Rename summary references to formatted_fields

ingwinlu:
Since FORMATTED_FIELDS was introduced the variables are not specific to
summary and contain other data as well.


justinmayer:
Thanks, Winlu!

PR #1843, opened 2015-10-13
do not fail tox if coveralls push fails

ingwinlu:
When tox environments are run locally, coveralls exits with an error
code as it is not authenticated to push to coveralls. Adding the - in
front allows tox to ignore a fail on this command and not mark the test
case as failed.


ingwinlu:
closed in favor of #1850

PR #1844, opened 2015-10-13
Replace get_summary with _get_summary

ingwinlu:
When memoizing summary a dummy get_summary function was introduced to
properly work. This has multiple problems:
* The siteurl argument is actually not used in the function, it is only
  used so it properly memoizes.
* It differs from how content is accessed and memoized.

This commit brings summary inline with how content is accessed and
processed. It also removes the old get_summary function with the unused
siteurl argument.

### DIFF
--- a/pelican/contents.py
+++ b/pelican/contents.py
@@ -273,6 +273,9 @@ class Content(object):
 
         return hrefs.sub(replacer, content)
 
+    def get_siteurl(self):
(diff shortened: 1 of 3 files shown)


ingwinlu:
fails on coveralls because it cleans up code that was run in tests, but did not influence test results. way to go.

justinmayer:
Thanks for the fixes, Winlu.  (^_^)

Issue #1845, opened 2015-10-14
Static markdown file


Kokxer:
Hi, I have a static file in `content/static` named `README.md`. I want pelican directly copy it to output root folder, so I have:

```
STATIC_PATHS = [
    'static/README.md',
]
EXTRA_PATH_METADATA = {
    'static/readme.md': {'path': 'README.md'},
}
```

in my `pelicanconf.py` file. Bug when I run `pelican`, I got: 

```
ERROR: Skipping static/README.md: could not find information about 'NameError: title'
```

With `-D`, I got:

```
DEBUG: Read file static/README.md -> Article
ERROR: Skipping static/README.md: could not find information about 'NameError: title'
```

Anyone have ideas how to tell pelican do not render this file?

justinmayer:
Try: Docs > Settings > `IGNORE_FILES`, applying it to your entire `static` folder.

Kokxer:
Thanks for your reply, but `IGNORE_FILES` cause pelican completely ignore my `README.md`, it was not copied or rendered to `output`.

ingwinlu:
`IGNORE_FILES` is global, if you don't want it to be treated as an article, set `ARTICLE_PATHS` to suite your needs or utilize `ARTICLE_EXCLUDES`

justinmayer:
Thanks for the correction, Winlu.

Kokxer:
@ingwinlu  `ARTICLE_EXCLUDES = ['static']` solved my problem, thanks you! But I'm doubting why `STATIC PATHS` isn't here to meet this requirement.

ingwinlu:
because the ArticlesGenerator runs before the StaticGenerator

Kokxer:
@ingwinlu  Thanks! Closing...

PR #1846, opened 2015-10-17
Add FAQ about static files parsed as content

ingwinlu:
Since #1509 got merged this pops up again and again (latest #1845). This
PR addresses the issue by adding a FAQ about it that explains why it is
necessary to add `*_EXCLUDES`.


justinmayer:
Thanks for the clarification, Winlu.

PR #1850, opened 2015-10-20
Make Pelican Python 3.5 compatible

### LINKED ISSUE #1847: truncate_html_words not working with python 3.5
while preparing a pr to add py35 to tox and travis I discovered that `pelican.utils.truncate_html_words` is no longer working with python 3.5. No shorting is taking place.

it is possible a good idea to avoid python 3.5 for now till we have made sure it is working properly.

If you are suffering from this issue please have a look at the docs which explain how to use virtualenv to create a custom virtual environment that contains a python version that is supported (2.7, 3.3, 3.4)

### LINKED ISSUE #1849: python warnings module ignored
when running the test suite warnings from the warnings module should be echod and if possible the test where they are are issued should fail.

this would prevent #1848 and provide a way to detect and upgrade problematic code before it becomes a problem

ingwinlu:
This summarizes all PR's from me that were open regarding logging/testing changes as it does not make sense to split them up.

* add logging for warnings during test suite run only (also error out tests that produce warnings that are originating from pelican's code)
* correct HTMLParser deprecations for Python 3.5
* correct ReaderTest to be properly Python 2 and Python 3 compatible
* add testing environment for Python 3.5 (tox and travis)
  Note that travis is not yet fully supporting Python 3.5, hence the matrix workaround
* do not fail tox environments just because submitting to coveralls fails

closes #1849
closes #1847

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,11 @@ env:
   - TOX_ENV=py27
   - TOX_ENV=py33
   - TOX_ENV=py34
+matrix:
+  include:
+    - python: 3.5
+      env:
+      - TOX_ENV=py35
 addons:
   apt_packages:
     - pandoc
--- a/pelican/log.py
+++ b/pelican/log.py
@@ -172,8 +172,26 @@ class LimitLogger(SafeLogger):
 logging.setLoggerClass(LimitLogger)
 
 
+def supports_color():
+    """
+    Returns True if the running system's terminal supports color,
+    and False otherwise.
+
+    from django.core.management.color
+    """
+    plat = sys.platform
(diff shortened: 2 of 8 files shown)


justinmayer:
Many thanks for all your hard work on this, Winlu.  \o/

loleg:
Looks like there's no more reason to dissuade folks from using Python 3 [in the install doc](https://github.com/getpelican/pelican/blob/master/docs/install.rst)? The only issue I have experienced is that an older version of Pelican (or plugin?) required wsgiref. Deleting it from `requirements.txt` was a very quick fix.

justinmayer:
@loleg: Thanks for the heads-up. Fixed.

PR #1852, opened 2015-10-29
Add '--no-mime-magic --no-preserve' to s3_upload

xenithorb:
`--no-mime-magic` is absolutely critical on POSIX generators with broken libmagic - css files for example will otherwise be uploaded with the incorrect header and be completely broken for the user without knowing why right away.

`--no-preserve` prevents leaking username and permissions, otherwise a header is passed and set on S3 that is the received by the client:

```
x-amz-meta-s3cmd-attrs: uid:1000/gname:xenith/uname:xenith/gid:1000/mode:33188/mtime:1446050473/atime:1446050473/md5:7a747784b1f67ae304c09eacc7210c2c/ctime:1446050585
```


justinmayer:
Thanks for the enhancement, Michael.

PR #1854, opened 2015-11-03
Test if summary is parsed by replacer

ingwinlu:
As mentioned in #1342 the issue is already resolved.

I took the liberty and reworded the commit issued by @smartass101 and only took the changes to the test.

Closes #1342


justinmayer:
Thanks to @ingwinlu and @smartass101!

PR #1858, opened 2015-11-05
Use <time> instead of <abbr>

meisterluk:
The pubdate given is in no way an abbreviation. Hence the semantics of
<abbr> do not justify its usage in this context.

<time> was introduced in HTML5. It's datetime attribute can override the
textContent. Anyway it specifies a date+time of a given event.

http://www.w3.org/TR/html5/text-level-semantics.html#the-time-element

Originally the pubdate attribute was meant to annotate that the
publication date of the closest section unit is specified, but this was
dropped and is not part of the official specification.

https://www.w3.org/wiki/HTML/Elements/time

This is a followup for issue #1832

### DIFF
--- a/pelican/themes/simple/templates/article.html
+++ b/pelican/themes/simple/templates/article.html
@@ -21,13 +21,13 @@
  {{ translations.translations_for(article) }}
   </header>
   <footer class="post-info">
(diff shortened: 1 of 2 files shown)


justinmayer:
Many thanks for the contribution, @meisterluk!

PR #1859, opened 2015-11-13
Generate output directory before 'CD'ing into it

HiPhish:
If the output directory does not exist the 'cd' will fail, but the
script will resume, starting the server in the base directory. Therefore
we first make sure the output directory actually exists.


justinmayer:
Thanks for the contribution, Alejandro.  (^_^)

PR #1862, opened 2015-11-15
Add --fatal=errors|warnings program option

kernc:
Or is there an alternative when one is using continuous integration to build and relies on program exiting cleanly (or not) as an indicator of success?

### DIFF
--- a/pelican/log.py
+++ b/pelican/log.py
@@ -175,7 +175,22 @@ class LimitLogger(SafeLogger):
     def enable_filter(self):
         self.addFilter(LimitLogger.limit_filter)
 
-logging.setLoggerClass(LimitLogger)
+
+class FatalLogger(LimitLogger):
+    warnings_fatal = False
+    errors_fatal = False
+
+    def warning(self, *args, **kwargs):
+        super(FatalLogger, self).warning(*args, **kwargs)
+        if FatalLogger.warnings_fatal:
+            raise RuntimeError('Warning encountered')
+
+    def error(self, *args, **kwargs):
+        super(FatalLogger, self).error(*args, **kwargs)
+        if FatalLogger.errors_fatal:
+            raise RuntimeError('Error encountered')
+
+logging.setLoggerClass(FatalLogger)
 
 
 def supports_color():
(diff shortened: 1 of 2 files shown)


pelican/log.py
```diff
@@ -203,7 +223,10 @@ def get_formatter():
         return TextFormatter()
 
 
-def init(level=None, handler=logging.StreamHandler(), name=None):
+def init(level=None, warnings_fatal=False, errors_fatal=False,
+         handler=logging.StreamHandler(), name=None):
+    FatalLogger.warnings_fatal = warnings_fatal
+    FatalLogger.errors_fatal = errors_fatal or warnings_fatal
```
avaris: Isn't this backwards? `warnings_fatal` should exit if warnings or errors occur, not the other way around.

avaris:
Also, can we not spread `sys.exit` all around the code? Multiple exit points just makes maintaining harder. Just raise an exception and let the code in `__init__.py` handle it.

pelican/log.py
kernc: After reading your email yesterday, I was actually agreeing with your observation right up until this moment when I saw the edit. These predicates be hard! :laughing:

justinmayer:
Hey @kernc! Many thanks for the follow-up changes. It looks like there is one [PEP-8 test failure](https://travis-ci.org/getpelican/pelican/jobs/91681877) in 86b7109 that needs addressing. Would you mind fixing it and amending that commit?

@avaris: Any other changes needed from your perspective?

kernc:
Amended, thanks.

I'm not completely completely sold on `--errors-fatal` switch name, though. Does `--fail-on` or even just `--fatal errors|warnings` sound better and acceptable?

kernc:
Have settled with `--fatal`.

justinmayer:
Nice work, @kernc. Many thanks for your efforts on this feature!

ingwinlu:
This breaks pelican when --fatal is not set.

https://gist.githubusercontent.com/ingwinlu/f39d5b6345f576cd2893/raw/6f6e02278b55fc1cd2b18176665d41a84152c24e/pelican%2520master%2520log,%2520.fatal%2520=%2520None

Examining args shows that obviously fatal defaults to None when it is not set resulting in the error above
```
Namespace(autoreload=False, cache_path=None, delete_outputdir=None, fatal=None, ignore_cache=False, output='/home/winlu/gitrepos/www.heroicdebugging.biz/output', path='/home/winlu/gitrepos/www.heroicdebugging.biz/content', relative_paths=False, selected_paths=None, settings='/home/winlu/gitrepos/www.heroicdebugging.biz/pelicanconf.py', theme=None, verbosity=10)
```

kernc:
Har, har, hurr!  :sweat_smile:

PR #1863, opened 2015-11-16
Sets followlinks=True in utils.py. Fixes #1785

### LINKED ISSUE #1785: Set followlinks=True in utils.py copy
When using themes with symlinked assets, the default method of `utils.copy` does not copy all of the assets. This fixes the issue:

```
for src_dir, subdirs, others in os.walk(source_, followlinks=True):

```

Any objection to making that the default behavior?

fboxwala:


justinmayer:
Thank you for the contribution, Fatema. Would you be so kind as to add one or more tests for this? That would help insure that any future changes that might break symlink functionality will be more easily caught. (Please refer to our [Contributing Guidelines](http://docs.getpelican.com/en/latest/contribute.html) for more details.)

fboxwala:
Would it be enough to just write tests checking to see if the copy utility still works, or should the tests check whether the problem with symlinked assets in #1785 is resolved?

justinmayer:
Checking whether the problem with symlinked assets in #1785 is resolved is probably the most useful and relevant test to accommodate this pull request. Would really appreciate your help with that!

justinmayer:
Yay! Thanks, Elana!

fboxwala:
@justinmayer I added a unittest on symlinks for the copy utility, is this mergeable?

ingwinlu:
no, as you do not test anything related to the pr

justinmayer:
@ingwinlu / @getpelican/reviewers: Perhaps we could suggest better tests so this PR could be updated and merged?

justinmayer:
Actually, this change was included when #2306 was merged.

PR #1866, opened 2015-11-17
notmyidea: support for share_post plugin

amitsaha:
Does this make sense? The idea is if you are using the share_post plugin, this will add the sharing links. For example: http://doingmathwithpython.github.io/trying-out-solutions.html


justinmayer:
Probably best for this kind of thing to not live in Pelican core. Otherwise, we'll have a ton of plugin-related things that aren't relevant to folks who aren't using those plugins.

PR #1874, opened 2015-11-28
Add MD_EXTENSION_CONFIGS setting

atodorov:
used to control Markdown extensions configuration

### DIFF
--- a/docs/content.rst
+++ b/docs/content.rst
@@ -53,7 +53,8 @@ install Markdown``.
 
 Pelican also supports `Markdown Extensions`_, which might have to be installed
 separately if they are not included in the default ``Markdown`` package and can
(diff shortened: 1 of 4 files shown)


ingwinlu:
see https://github.com/getpelican/pelican/pull/1638

atodorov:
@ingwinlu so you're saying that the new interface will be MD_EXTENSIONS which is a dict with extension names and their settings, and this PR is not needed ? Is this correct ? 

Which branch/version of Pelican is this planned for ?

ingwinlu:
The setting will stay MD_EXTENSIONS, it's format will change. Your PR only addresses a small part of the issues, the other one is more mature.

Since it is only absolutely needed when python Markdown 2.7 releases we still have some time to get it in.

PR #1875, opened 2015-12-01
Making `setup.py` `pep8`-compliant

habi:
To test the water with the pelican developers...

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,6 @@
 #!/usr/bin/env python
 from os import walk
-from os.path import join, relpath, dirname
-
+from os.path import join, relpath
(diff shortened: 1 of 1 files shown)


habi:
Would you be interested in re-formatting some code to follow the [PEP8](https://www.python.org/dev/peps/pep-0008/)?

I would do it for [24 Pull requests](http://24pullrequests.com/)...

leotrs:
@habi Apologies for the belated response. This looks good to me.

justinmayer:
I believe this has already been addressed via aa2c84a. If that's not the case, please feel free to comment here and I can re-open.

PR #1880, opened 2015-12-26
Ensure DEFAULT_DATE = 'fs' actually uses modified time

allanman:
change so that DEFAULT_DATE = 'fs' makes pelican actually uses file mtime as stated in the manual (was ctime)


allanman:
usually there's no difference, but i ran into trouble when i had to move my content to a new partition/fs. mtime can be retained, but ctime is set to "now", changing the order of everything

justinmayer:
Thanks to @allanman for the fix and to @iKevinY for the review.  (^_^)

PR #1885, opened 2016-01-03
Add missing Makefile PHONY target stopserver

andreaso:
The Makefile generated by pelican-quickstart isn't listing the stopserver target as a phony target.

This commit/pull-requests adds stopserver as a dependency to the .PHONY target.


justinmayer:
Thanks, Andreas!

Issue #1891, opened 2016-01-15
pelican run fails when --fatal is not set


ingwinlu:
see https://github.com/getpelican/pelican/pull/1862

PR #1893, opened 2016-01-18
Don't fail if --fatal not provided

kernc:
Fixes https://github.com/getpelican/pelican/issues/1891

Sorry.

### DIFF
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -323,7 +323,7 @@ def parse_arguments():
                         help='Comma separated list of selected paths to write')
 
     parser.add_argument('--fatal', metavar='errors|warnings',
-                        choices=('errors', 'warnings'),
+                        choices=('errors', 'warnings'), default='',
                         help=('Exit the program with non-zero status if any '
                               'errors/warnings encountered.'))
 


justinmayer:
@kernc: Looks like there are some test failures here.

ingwinlu:
NO, fix the design, don't hack a solution.

kernc:
Got it, I set the argparser's `default=''`. `ArgumentDefaultsHelpFormatter` now prints `(default: )` which isn't far from truth anyway.

I don't understand why the build is failing, though. Doesn't seem related.

avaris:
Build is failing because Pygments recently released a [new version (2.1)](https://pypi.python.org/pypi/Pygments/2.1) and the output is somehow different than the previous one (2.0.2). So yeah, it's not related to this PR. We probably need to update our test output before any other PR.

ingwinlu:
while this would fix the issue I would rather see a different approach:

introduce log.FATAL = 0,1,2 where 0 is for nothing, 1 is for errors, 2 is for warnings; or similar.

I do not like comparing strings in that use case. If you do not want to do it that way this will do in the meantime.

justinmayer:
Merging this for now.

@kernc: Would you perhaps consider a follow-up pull request that implements @ingwinlu's suggested approach?

kernc:
I was amazed how tolerant you were of this failing. :smiley: 

I didn't quite _get_ @ingwinlu's suggestion; why would one wish to complicate stuff by introducing a lookup table instead of using the (only) two booleans that make sense?

PR #1894, opened 2016-01-19
Default --output to "content" instead of "output"

W7PEA:
The default workflow of the rest of the tools expect the .rst|.md files to be in the content folder, but the pelican-import command was defaulting to placing those files in the output folder which is where pelican puts the generated html. This change creates a workflow which is more obvious for new users.


W7PEA:
If this is approved and merged, I'll update the docs too. http://docs.getpelican.com/en/3.6.3/importer.html#optional-arguments

W7PEA:
Redoing with the doc changes

PR #1896, opened 2016-01-19
--output moved to content folder

W7PEA:
fixed whitespace in my last checkin


W7PEA:
Failures are the same as the previous PR which are due to an unrelated change in the test tool versions.

PR #1899, opened 2016-01-20
Change import output to content

W7PEA:
This PR changes the default --output folder from ./output' to ./content. The ./output folder is where Pelican places the generated HTML, its not where you want the markup files placed. The ./content folder is the more appropriate default as it allows for the two tools to work together based on their defaults.


W7PEA:
Redid my change and resubmitted. The failures all appear to my eye to be the same as the ones in PR #1893 and based on the comments there about a new Pygments version I believe can be ignored. Let me know if that's not the case. Thanks for your collective patience!

justinmayer:
Hi Patrick. Apologies for the delay in reviewing your contribution. Could you elaborate on why you think the default output directory should be `./content`? I personally believe the current practice of keeping source content and generated output separate from one another to be far more sensible.

avaris:
@justinmayer This appears to be about `pelican-import`. Right now it writes the generated `.rst/.md` files to an `output` folder by default. But after import, they will be `content` for `pelican` itself. So this change makes sense (at least in terms of `pelican`s convention). PR just needs a squash and rebase.

justinmayer:
Ah, thank you for that explanation. Makes much more sense now.  (^_^) 

@W7PEA: Would you please [rebase & squash your commits](https://github.com/getpelican/pelican/wiki/Git-Tips#squashing-commits)?

leotrs:
This seems to have been approved by @avaris, and was squashed and rebased by @W7PEA as requested. Looks ready to merge, @justinmayer.

avaris:
Not really squashed. There are 9 commits here.

W7PEA:
Hey all, I'm no longer using Pelican and can't look into this any longer.

justinmayer:
If someone else wants to pick up the torch, please feel free to do so.

PR #1909, opened 2016-02-11
New feature: support for {include} syntax. Fixes #1902.

### LINKED ISSUE #1902: RFE: Make it possible to include content files from other content files
Hi guys,

atodorov:
The new {include} syntax makes it possible to include
frequently used text snippets into your content.

### DIFF
--- a/pelican/contents.py
+++ b/pelican/contents.py
@@ -321,10 +321,19 @@ class Content(object):
         """
         regex = r"""[{|]include[|}](?P<path>[\w./]+)"""
         hrefs = re.compile(regex, re.X)
+        processed_paths = []
+        # In Python 3.x we can use the `nonlocal` declaration, in `replacer()`,
+        # to tell Python we mean to assign to the `source_path` variable from
+        # `_update_includes()`.
+        # In Python 2.x we simply can't assign to `source_path` in `replacer()`.
+        # However, we work around this by not assigning to the variable itself,
+        # but using a mutable container to keep track about the current working
+        # directory while doing the recursion.
+        source_dir = [source_path]
 
         def replacer(m):
             path = m.group('path')
-            path = self._path_replacer(path, source_path)
+            path = self._path_replacer(path, source_dir[0])
             path = posixize_path(
                     os.path.abspath(
                         os.path.join(self.settings['PATH'], path)
@@ -343,15 +352,25 @@ class Content(object):
                 logger.warning("Unable to read `%s`, skipping include.", path)
                 return ''.join(('{include}', m.group('path')))
 
+            # recursion stop
+            if path in processed_paths:
+                raise RuntimeError("Circular inclusion detected for '%s'" % path)
+            processed_paths.append(path)
+
             reader = self.readers.reader_classes[ext](self.settings)
             text, meta = reader.read(path)
 
             # if we recurse into another file to perform more includes
             # self._path_replacer needs to know in which directory
             # it operates otherwise it produces wrong paths
-            source_dir = posixize_path(os.path.dirname(path))
+            source_dir[0] = posixize_path(os.path.dirname(path))
+            current_source_dir = source_dir[0]
 
-            text = self._update_includes(text, source_dir)
+            # recursively replace other includes
+            text = hrefs.sub(replacer, text)
+
+            # restore source dir
+            source_dir[0] = current_source_dir
             return text
 
         return hrefs.sub(replacer, content)
--- a/pelican/contents.py
+++ b/pelican/contents.py
@@ -11,7 +11,7 @@ import sys
 import pytz
 
 import six
-from six.moves.urllib.parse import urlparse, urlunparse
+from six.moves.urllib.parse import unquote, urlparse, urlunparse
 
 from pelican import signals
 from pelican.settings import DEFAULT_CONFIG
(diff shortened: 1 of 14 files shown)


avaris:
Nice idea, but it won't work as expected for non-html content. You're dumping the content of the file in place of the include. For `.rst` or `.md` (or any other parsed content), this means that it'll put raw syntax instead.

e.g. for a markdown file to be included, you'd get:

```
[link](http://www.example.com)
```

rather than

```
<a href="http://www.example.com">link</a>
```

Plus, you have no error checking at all. No graceful fallbacks. Give a wrong path by mistake and things will break down.

atodorov:
@avaris - Good points, thanks for the feedback. 

Yes I dump raw content in place of the include, this will be a problem for inclusions between files of different types. I'll try to render the new content to HTML before including it. How does it sounds ?

On the second point - error checking - will add it. Is there any particular exceptions or methods to handle this. If possible point me to some place in the code which has good error checking and fallbacks.

pelican/contents.py
```diff
@@ -276,12 +288,41 @@ def replacer(m):
...
+        def replacer(m):
+            path = m.group('path')
+            path = self._path_replacer(path, source_path)
+            path = posixize_path(
+                    os.path.abspath(
+                        os.path.join(self.settings['PATH'], path)
+                    )
+                )
+
+            with pelican_open(path) as text:
```
avaris: This will throw an exception if `path` doesn't exist or isn't a file. You could use the same method done in `_update_content`. Show a warning and ignore the replace.

avaris:
Also, if I'm not mistaken, `pelican` will *try* to read your includes as regular content files. Either it'll will succeed and they will be considered for `Page/Article` or it'll fail and spit a bunch of warnings. I don't think any of these outcomes is what you want. They should be ignored, IMO.

atodorov:
@avaris - it took me a while to get back to this feature but I'm back on it. Regarding your last comment:

Yes, Pelican tries to read the included files as regular content. In most of the cases it fails b/c I don't provide a title for example. I'll be happy to ignore them if there is an easy way to do so. Is there such a way?

If not then I propose to have an additional setting, something like `IGNORE_DIRS` which are directories to ignore reading content from and instruct the user to place their include files there. I'm not sure if this won't conflict with `pelican_open `though.

I'll look at your previous comment as well and update this PR when ready.

atodorov:
`IGNORE_FILES` is the setting I needed earlier. Just listing my included content there stops Pelican from trying to render it and produce an error. I've added a section about this in the docs. 

Now looking into the final issue from 2nd comment - how to properly include content between files of various formats.

atodorov:
ping, I've rebased to master. This is ready for code review.

justinmayer:
Fellow @getpelican/reviewers: Would you please take a look a this PR and provide feedback?

pelican/contents.py
```diff
@@ -187,6 +199,30 @@ def get_url_setting(self, key):
...
+        if path.startswith('/'):
+            path = path[1:]
+        else:
+            # relative to the source path of this content
+            path = self.get_relative_source_path(
+                os.path.join(relative_dir, path)
+            )
+
+        if path not in self._context['filenames']:
+            unquoted_path = path.replace('%20', ' ')
```
leotrs: What about using [unquote](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.unquote)?

atodorov:
@getpelican/reviewers can you review?

ingwinlu:
hello everyone,

sadly I am currently extremly busy so I only had time for a quick look:
Is there a way to avoid having the import of readers into content the way
you do it in the PR? I would prefere restructuring code to having to start
to use such kind of imports.

avaris:
I agree with @ingwinlu about circular imports. Instead you can pass in the reader to the `__init__` of `Content` from `Readers.read_file` ([here](https://github.com/getpelican/pelican/blob/master/pelican/readers.py#L570)). 

A check for infinite recursion might be needed as well (if an import imports itself).

Also, no ads please :).

atodorov:
@avaris I've implemented your proposal to pass `Readers()` to `Content.__init__()` and I broke `test_cache.py`. I think we need to create the readers like I do in `test_content.py` and that probably needs to go to `generators.py` or somewhere in `test_cache.py` but ATM I can't figure out where exactly.

avaris:
@atodorov, the more I look at it, the more I think that it's better to resolve the `{include}` at the reader level, before it reaches to the `Content`. I know, this will be a major overhaul, but I think it will be cleaner that way.

atodorov:
> @atodorov, the more I look at it, the more I think that it's better to resolve the {include} at the reader level, before it reaches to the Content. I know, this will be a major overhaul, but I think it will be cleaner that way.

I can give it a try but honestly my Pelican knowledge is a bit rusty ATM so you will have to tell me where to look at or give me some hints as to how you see things being implemented.

justinmayer:
@avaris: Any chance you could point Alexander in the direction you're thinking about this PR?

Lucas-C:
@atodorov do you mind if I take over on this PR to try to finish this feature ?
I'd love to have this !

atodorov:
@Lucas-C go ahead.

@justinmayer, @avaris how do you want this work to be continued? Maybe open another PR with everything here cherry-picked + necessary changes  ?

Lucas-C:
I'm closing this.

Follow-up pull requests :
- https://github.com/getpelican/pelican/pull/2628
- https://github.com/getpelican/pelican/pull/2638

PR #1923, opened 2016-03-03
Use 'ghp-import -p' to push instead of 'git push origin {branch}'

Jangsea:
`ghp-import` has the option `-p` to push the branch to origin/{branch} after committing. It doesn't need to call `git` separately. ( ref: https://pypi.python.org/pypi/ghp-import/0.4.2 )


justinmayer:
Thanks for the optimization, Jangsea.

PR #1926, opened 2016-03-13
More granular control of tags and categories slugs. Fixes #1873

### LINKED ISSUE #1873: RFE: Add slugs for tags for backward compatibility with existing URLs
Hi guys,
I've migrated to Pelican from Octropress and managed to keep all articles URLs the same but had troubles with some tags. If your tag contains a dot (e.g. fedora.planet) then its sluge removes the dot (becomes fedoraplanet) and I don't see any way to keep my existing tag URL.

atodorov:
- add TAG_SUBSTITUTIONS AND CATEGORY_SUBSTITURIONS settings
- make slugify keep non-alphanumeric characters if configured

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,7 +4,14 @@ Release history
 Next release
 ============
 
-- Nothing yet
+* ``SLUG_SUBSTITUTIONS`` now accepts 3-tuple elements, allowing to keep
+  non-alphanum characters. Existing 2-tuple configurations will continue to work
+  without change in behavior. The new 3rd parameter has side effects when there
+  are multiple substitutions defined. Plese see the docs.
+* Tag and category slugs can be controlled with greater precision using the
+  ``TAG_SUBSTITUTIONS`` and ``CATEGORY_SUBSTITUTIONS`` settings. These also
+  allow for keeping non-alphanum characters for backward compatibility with
+  existing URLs.
 
 3.6.3 (2015-08-14)
 ==================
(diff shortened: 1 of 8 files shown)


justinmayer:
@ingwinlu / @avaris: Any comments on this proposed feature?

atodorov:
Hi guys,
I've added another commit to this PR, implementing AUTHOR_SUBSTITUTIONS as well b/c I happen to need that as well. Authors on my blog want to have specific usernames (usually the ones they use on GitHub already). 

Let me know if you want me to squash the two commits.

justinmayer:
Thanks for the enhancement, Alexander!

avaris:
This introduced a bug. 

When you replace a property, the whole thing (`getter` **and** `setter`) is replaced. This means that the `slug` property in subclasses becomes read-only and can't be assigned any more.
See getpelican/pelican-plugins#846

Also, is there a specific reason for not applying `SLUG_SUBSTITUTIONS` to `Author`?

atodorov:
@avaris IIRC `SLUG_SUBSTITUTIONS` are not applied for `Author` to avoid conflicts with the settings in `AUTHOR_SUBSTITUTIONS` but I don't recall the exact reason ATM and I see I've not documented it thoroughly.


Wrt the bug: can it be easily solved by removing the `@property` decorator ?

avaris:
No. If you remove the decorator, it becomes a method. You can't get it like `cat.slug`, it'll be `cat.slug()`. When you replace the property, you need to provide the setter as well. That means duplicating the base class setter in all subclasses. Which isn't ideal.

Alternatively, a better approach is to handle it in the base class (similar to what `_from_settings` does with `self.__class__.__name__`) and I was going to do that but then noticed the special handling of the `Author`. That makes things complicated. Even more so, it looked odd to me since I couldn't see the reason for the different treatment. That's why I asked.

PR #1423, opened 2014-07-29
Add MD_OUTPUT_FORMAT setting

randomllama:
This adds the MD_OUTPUT_FORMAT setting to keep Markdown from generating non-valid code.  See issue #1024.  Defaults to Markdown's current default of xhtml1.

PR #1927, opened 2016-03-14
Replaces MD_EXTENSIONS with MARKDOWN; Fixes #1024 

### LINKED ISSUE #1024: Footnote processing returns non-valid html rel="footnote"
When validating pelican-created HTML5, validation fails on the markup of footnotes.

It would appear that the Markdown processor automatically includes the attribute _rel="footnote"_, which attribute is not included in html specification.

Scheirle:
MARKDOWN_SETTINGS will be directly passed to the Markdown constructor, which is more flexible.

**TODO:**
- [x] Update Documentation
- [x] Fix tests
- [x] Make flake8 happy 

**Refs:**
Fixes #1024 
Replaces #1423
http://pythonhosted.org/Markdown/reference.html#using-markdown-as-a-python-library

### DIFF
--- a/pelican/readers.py
+++ b/pelican/readers.py
@@ -246,8 +246,16 @@ class MarkdownReader(BaseReader):
 
     def __init__(self, *args, **kwargs):
         super(MarkdownReader, self).__init__(*args, **kwargs)
-        self.extensions = self.settings['MD_EXTENSIONS']
-        self.extensions.setdefault('markdown.extensions.meta', {})
+        # make sure 'extension_configs' exists and
+        # and either way 'markdown.extensions.meta' must be in there
+        settings = self.settings['MARKDOWN']
+        settings.setdefault('extension_configs', {})
+        settings['extension_configs'].setdefault(
+            'markdown.extensions.meta', {})
+        settings.setdefault('extensions', [])
+        settings['extensions'].extend(
+            list(settings['extension_configs'].keys()))
+        settings['extensions'] = list(set(settings['extensions']))
         self._source_path = None
 
     def _parse_metadata(self, meta):
@@ -283,8 +291,7 @@ class MarkdownReader(BaseReader):
         """Parse content and metadata of markdown files"""
 
         self._source_path = source_path
-        self._md = Markdown(extensions=self.extensions.keys(),
-                            extension_configs=self.extensions)
+        self._md = Markdown(**self.settings['MARKDOWN'])
         with pelican_open(source_path) as text:
             content = self._md.convert(text)
 
--- a/pelican/settings.py
+++ b/pelican/settings.py
@@ -107,6 +107,7 @@ DEFAULT_CONFIG = {
             'markdown.extensions.extra': {},
             'markdown.extensions.meta': {},
         },
+        'output_format': 'html5',
     },
     'JINJA_EXTENSIONS': [],
     'JINJA_FILTERS': {},
--- a/pelican/settings.py
+++ b/pelican/settings.py
@@ -101,10 +101,12 @@ DEFAULT_CONFIG = {
     'PELICAN_CLASS': 'pelican.Pelican',
     'DEFAULT_DATE_FORMAT': '%a %d %B %Y',
     'DATE_FORMATS': {},
-    'MD_EXTENSIONS': {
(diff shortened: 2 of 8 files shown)


pelican/readers.py
```diff
@@ -246,8 +246,13 @@ class MarkdownReader(BaseReader):
...
         super(MarkdownReader, self).__init__(*args, **kwargs)
-        self.extensions = self.settings['MD_EXTENSIONS']
-        self.extensions.setdefault('markdown.extensions.meta', {})
+        # make sure 'extension_configs' exists and
+        # and either way 'markdown.extensions.meta' must be in there
+        settings = self.settings['MARKDOWN_SETTINGS']
+        settings.setdefault('extension_configs', {})
+        settings['extension_configs'].setdefault(
+            'markdown.extensions.meta', {})
+        settings['extensions'] = list(settings['extension_configs'].keys())
```
randomllama: I found this line confusing.  Since we're passing kwargs to markdown, I was expecting `MARKDOWN_SETTINGS = { 'extensions': [ 'my_custom_extension' ] }` to work in the config.  Perhaps it would be better to append?

```
settings.setdefault('extension', [])
settings['extensions'].append(list(settings['extension_configs'].keys()))
```

..or the like.

Scheirle: You're right, fixed it.
Now it is also possible to load custom extensions e.g.:
```
from markdown.extensions import Extension

class MyExtension(Extension):
    def extendMarkdown(self, md, md_globals):
        print("run MyExtension")

MARKDOWN_SETTINGS = {
    'extensions': ['markdown.extensions.headerid', MyExtension()],
    'extension_configs': {
        'markdown.extensions.extra': {},
        'markdown.extensions.headerid': {'level': 1},
        'markdown.extensions.codehilite': {'css_class': 'highlight', 'linenums': True},
    }
}
```

Scheirle:
Has anyone else some feedback or can this be merged?

leotrs:
@justinmayer

justinmayer:
Hi Bernhard! Very sorry for the long delay in reviewing.

Code looks good to me. Only one question... What about calling the new setting `MARKDOWN`? The current proposed `MARKDOWN_SETTINGS` seems a bit redundant since all the entries in the Pelican settings file are... settings.  (^_^)

Any @getpelican/reviewers care to submit a review of this PR?

Scheirle:
Renamed `MARKDOWN_SETTINGS` to `MARKDOWN`. 

Note to self:
```
git filter-branch -f --tree-filter 'git grep -l "MARKDOWN_SETTINGS" | xargs sed -i "s/MARKDOWN_SETTINGS/MARKDOWN/g"' 515f2fb5b2aee33086b6cd53830c388a1c7a9fa3..HEAD
git filter-branch -f --msg-filter 'sed "s/MARKDOWN_SETTINGS/MARKDOWN/g"' 515f2fb5b2aee33086b6cd53830c388a1c7a9fa3..HEAD
```

pelican/settings.py
```diff
@@ -101,10 +101,12 @@ def load_source(name, path):
...
-    'MD_EXTENSIONS': {
-        'markdown.extensions.codehilite': {'css_class': 'highlight'},
-        'markdown.extensions.extra': {},
-        'markdown.extensions.meta': {},
+    'MARKDOWN': {
+        'extension_configs': {
+            'markdown.extensions.codehilite': {'css_class': 'highlight'},
+            'markdown.extensions.extra': {},
+            'markdown.extensions.meta': {},
+        },
```
Scheirle: Maybe we should add `'output_format' : 'html5'` as default?

Refs
#1024
http://pythonhosted.org/Markdown/reference.html#output_format

Totktonada: Sooner or later that should happens (from Python-Markdown side of from Pelican one). Sooner is better. Only one note: it is breaking change and should be mentioned in the release notes.

justinmayer:
@Scheirle said:

> Maybe we should add 'output_format' : 'html5' as default?

I think we should, yes. I'll mention that in the release blog post when the time comes.

justinmayer:
Thanks for all your hard work on this, Bernhard. Much appreciated!  👍

PR #1942, opened 2016-04-12
Fix description of the search bar's positon

muellermartin:
The search bar in the currently used Read the Docs Sphinx Theme is
located at the top-left instead of top-right.


justinmayer:
Thanks for noticing that artifact from the RtD theme change!

PR #1947, opened 2016-05-04
Fix issue with one-element tuple

magul:
It looks, that there is/was bug introduced in https://github.com/getpelican/pelican/commit/c386e29d about one-element tuple.


justinmayer:
Thanks for the fix, Tomasz!

PR #1948, opened 2016-05-05
Ignore output for development server

### LINKED ISSUE #1888: make devserve starts process in background but connected to terminal
Running `make devserve` using the pelican-generated Makefile starts the integrated web-server, which logs requests to the terminal, but also puts the process in the background, making the shell and the web-server fight for the terminal.

This is IMHO bad UI design, a process should not write to the terminal when it is running in the background. This situation could be prevented by either not forking the web-server process to the background or by writing the web-servers output to a log file instead of the terminal.

muellermartin:
The Makefile starts pelican to continuously generate the website and a
development web server both in the background, but the commands do still
output text to the shell. This behaviour is annoying when the same shell
is used further, therefore the output of both commands get redirected to
/dev/null.
Note: Even the -q, --quiet flag of pelican does not suppress
all output.

This fixes #1888

### DIFF
--- a/pelican/tools/templates/develop_server.sh.in
+++ b/pelican/tools/templates/develop_server.sh.in
@@ -62,11 +62,11 @@ function start_up(){
   local port=$$1
   echo "Starting up Pelican and HTTP server"
   shift
-  $$PELICAN --debug --autoreload -r $$INPUTDIR -o $$OUTPUTDIR -s $$CONFFILE $$PELICANOPTS &
+  $$PELICAN --quiet --autoreload -r $$INPUTDIR -o $$OUTPUTDIR -s $$CONFFILE $$PELICANOPTS &> /dev/null &
   pelican_pid=$$!
   echo $$pelican_pid > $$PELICAN_PID
   mkdir -p $$OUTPUTDIR && cd $$OUTPUTDIR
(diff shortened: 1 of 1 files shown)


leotrs:
It makes sense to me that messages are not printed when the server is run in the background. However, this being the development server, it seems desirable to retain them in some way.

Would it be possible to have a log file as an optional parameter? If there is, we redirect all output there, otherwisee, we can just redirect everything do /dev/null.

muellermartin:
I already thought about using a log file, but the problem here is, that there are actually two programs running (Pelican in autoreload mode and the Pelican web server) which both write to stdout independently.

I don't really know what happens if both write simultaneously into the same log file (maybe the text even gets garbled if the output is not line buffered?).

One solution could be to use seperate log files for each program.

And there is another problem: When do the logfiles get pruned? After a restart? A certain size or age?

leotrs:
@justinmayer ^

leotrs:
@muellermartin I can't decide on this point. Awaiting @justinmayer input.

justinmayer:
The previous development server functionality was replaced via #2351, so it's not clear if this pull request is still relevant. If so, please feel free to re-submit this pull request against current master.

PR #1952, opened 2016-05-09
import: multiple wordpress attachments may point to the same URL, download it only once

catdog2:
This PR also includes a very small fix to support URLs with non-ascii characters which is unrelated but touches the same lines.

### DIFF
--- a/pelican/tools/pelican_import.py
+++ b/pelican/tools/pelican_import.py
@@ -647,7 +647,7 @@ def get_attachments(xml):
     return attachedposts
 
 
-def download_attachments(output_path, urls):
+def download_attachments(output_path, urls, successful_url_cache):
     """Downloads WordPress attachments and returns a list of paths to
     attachments that can be associated with a post (relative path to output
     directory). Files that fail to download, will not be added to posts"""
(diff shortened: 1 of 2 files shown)


pelican/tools/pelican_import.py
```diff
@@ -667,7 +667,12 @@ def download_attachments(output_path, urls):
             os.makedirs(full_path)
         print('downloading {}'.format(filename))
         try:
-            urlretrieve(url, os.path.join(full_path, filename))
+            if(url in successful_url_cache):
+                print('already downloaded: {} ... skipping'.format(filename))
+            else:
+                quote_url = quote(url, safe="%/:=&?~#+!$,;'@()*[]")
+                urlretrieve(quote_url, os.path.join(full_path, filename))
+                successful_url_cache.add(url)
```
leotrs: Does every call to `urlretrieve` generate a successful url?
Is there a way to test if `urlretrieve` was actually successful and only then add it to `successful_url_cache`?

justinmayer:
Hi Peter. Is there any chance you would be willing to rebase your pull request on current master so it can be merged?

catdog2:
Thanks for the reminder, I will try to find some time to revisit this in the near future.

justinmayer:
Hey Peter. If you find the time to revisit this, your contribution would be welcome. 😄

PR #1953, opened 2016-05-09
Introduce timezone option for import from wordpress

catdog2:
Wordpress stores the date of a post inside 2 different fields
(post_date_gmt and post_date). If the new parameter is specified
the value of post_date_gmt is taken instead of post_date and
converted to the specified timezone.

### DIFF
--- a/pelican/tools/pelican_import.py
+++ b/pelican/tools/pelican_import.py
@@ -12,6 +12,8 @@ import time
 
 from codecs import open
 
+import pytz
+
 from six.moves.urllib.error import URLError
 from six.moves.urllib.parse import urlparse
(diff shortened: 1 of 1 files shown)


nstinus:
A related addition #1972 I made allows to default the timezone with the current locale provided tzlocal is installed. It might be of interest here.

leotrs:
1. In what case would the new `date_parse` throw an exception?
2. Please add the exception class instead of a naked `except: ` clause (use `except ExceptionClass: `)
3. Could `date_parse` catch its own exception, instead of wrapping the call with `try`?

justinmayer:
If someone would like to pick this effort up and improve upon it, by all means please feel free to do so.

PR #1476, opened 2014-09-24
Makes DEFAULT_DATE accept string dates; fixes #1464

areyoutoo:
Took a stab at tests and docs. I'm relatively new to Python development, so I apologize in advance if I've missed anything.

PR #1959, opened 2016-05-24
Accept string dates in DEFAULT_DATE. Fixes #1464; fixes #1476

### LINKED ISSUE #1464: Support string formatted dates for DEFAULT_DATE

atodorov:
I've cherry-picked the commits from https://github.com/getpelican/pelican/pull/1476, rebased to latest master and squashed them together, fixed pep8 issues and kept original authorship intact.

Opening new PR b/c I can't amend the original one and b/c there is no movement there. 
@justinmayer you already reviewed this and I believe it is ready to merge.


justinmayer:
Thanks, Alexander. Much appreciated!

PR #1961, opened 2016-05-26
Pin Pygments version to 2.1.3. Fixes #1916

### LINKED ISSUE #1916: Pin Pygments to specific version
Recent updates have caused test failures, so perhaps we should pin this particular dependency.

atodorov:


justinmayer:
Many thanks for implementing this, Alexander.  (^_^)

PR #1964, opened 2016-05-28
Fix typogrifying objects without title

rr-:
This fixes my use case where I use `readers.read_file` from within a plugin to load something that is neither a page nor an article and it makes little sense for it to have a title, but it has metadata and content that should be parsed like the rest of the objects.

### DIFF
--- a/pelican/readers.py
+++ b/pelican/readers.py
@@ -545,6 +545,8 @@ class Readers(FileStampDataCacher):
 
             if content:
                 content = typogrify_wrapper(content)
+
+            if 'title' in metadata:
                 metadata['title'] = typogrify_wrapper(metadata['title'])
 
             if 'summary' in metadata:


iKevinY:
Looks fine to me. Maybe even pull it out of the `if content` block so that all 3 of the conditionals are independent from one another?

rr-:
Fixed

Scheirle:
The [pelican_comment_system](https://github.com/getpelican/pelican-plugins/tree/master/pelican_comment_system) also needs this fix. It uses `readers.read_file` on comments which don't have a title.

I didn't run into this error earlier since I don't use `typogrify`.

---

:+1:

rr-:
Status?

justinmayer:
Apologies for the delay. Thanks to @rr- for the fix and to @iKevinY and @Scheirle for reviewing.  (^_^)

rr-:
Thanks :3

PR #1967, opened 2016-05-30
Add default settings file name to argparse. Fixes #1966

### LINKED ISSUE #1966: Reload settings file when not specified
Explicitly defining a settings file will be reloaded when using the --autoreload option. Using the default will indicate that the settings file is tracked, but will not reload the file when it is changed.

NathanHarrington:
Add a default to the settings file name in the argparse configuration. This will enable the triggering of the reload function when the configuration file is not explicitly set on the command line.

### DIFF
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -277,6 +277,7 @@ def parse_arguments():
                         '"output" in the current path.')
 
     parser.add_argument('-s', '--settings', dest='settings',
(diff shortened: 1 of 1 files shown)


justinmayer:
For future contributions, it would be helpful to add a more descriptive commit message such as:

> Add default settings file name to argparse. Fixes #1966

justinmayer:
Reverted/superseded by improvement in #2005.

PR #1969, opened 2016-06-01
Docker nginx deploy

robren:
# A feature to allow a pelican site to be deployed inside of a container.

I've run the python -m unittest discover set of tests 
    Ran 187 tests in 3.868s
    OK (skipped=21)

This was tested against python 2.7.6 ( happy to test on other versions of course if this feature looks acceptable).
I've run the flake8 command on the two modified files

```
- A new fabric target is introduced 'docker-rebuild'. This will rebuild a
  production version of the users site and deploy this by bind-mounting the
  pelican output directory to the web content  directory within the container.

- The fabric target takes care of container management, cleaning up previous containers
  and instantiating new ones whenever the ' fab docker rebuild' command is issued.

- Unlike other solutions where pelican runs inside the container, this solution
  decouples pelican from the deployment of the blog, it's more in keeping with
  the existing fabric build targets e.g. cf_upload

- The default container based web server is nginx, this can be changed when
  running the pelican_quickstart

- The docker-rebuild target works transparently with docker-machine if
  desired, to allow for very fast rebuild and update of pelican sites running
  within remote containers.

- Docker must be preinstalled on the host machine for this to work
```

Happy to squash the commit too.

Comments welcome.

Rob Rennison

### DIFF
--- a/pelican/tools/pelican_quickstart.py
+++ b/pelican/tools/pelican_quickstart.py
@@ -43,7 +43,10 @@ CONF = {
     'default_pagination': 10,
     'siteurl': '',
     'lang': 'en',
-    'timezone': 'Europe/Paris'
+    'timezone': 'Europe/Paris',
+    'docker_base_image': 'nginx',
+    'docker_target_dir': '/usr/share/nginx/html',
+    'docker_container_name': 'pelican_site_container'
 }
 
 # url for list of valid timezones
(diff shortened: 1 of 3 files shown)


leotrs:
@robren looks good. If the checks passed this could would maybe get further review.

robren:
@leotrs It now passes the checks. I also squashed most of my commits, to provide a cleaner looking log.

Let me know if there's anything else needed.

Cheers

Rob

leotrs:
cc. @justinmayer.

ingwinlu:
I favour closing this pull request due to the following reasons:

* The proposed Dockerfile provides little value as it can be easily be replicated by hand
* The proposed workflow is not very typical for Docker based deployments as it assumes the docker container is run where it is built
* it has been around for multiple years with little interest.

However I would like to also point out that nothing is really wrong with this besides the non pinned docker container base image which would require us to test for regression if the image changes paths for example.

PR #1972, opened 2016-06-04
Localize default lang and timezone in pelican-quickstart

### LINKED ISSUE #1971: Localize defaults in pelican-quickstart
Now, default language and timezone are hard-coded ('en', 'Europe/Paris').
Use system locale to adjust default values.

nstinus:
Fixes #1971.

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -28,8 +28,9 @@ by asking some questions about your site::
     pelican-quickstart
 
 For questions that have default values denoted in brackets, feel free to use
-the Return key to accept those default values. When asked for your URL prefix,
-enter your domain name as indicated (e.g., ``http://example.com``).
+the Return key to accept those default values [#tzlocal_fn]_. When asked for
+your URL prefix, enter your domain name as indicated (e.g.,
+``http://example.com``).
 
 Create an article
 -----------------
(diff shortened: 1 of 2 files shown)


avaris:
You need to somehow inform the user about the new optional dependency (`tzlocal`).

nstinus:
You are right, I thought about it too.
I'll have a look at my options.

nstinus:
Hi,

Either I add tzlocal as a hard requirement, which I don't really like.
Or I use extra_requires
<http://peak.telecommunity.com/DevCenter/setuptools#declaring-extras-optional-features-with-their-own-dependencies>
from
setuptools, which does not seem to be user- but more dev-oriented.

As far as I an tell, there is nothing lke this:
The Markdown module is optional and yet, I don't think it is advertised as
such anywhere.

Nicolas

avaris:
Markdown is stated [here](http://docs.getpelican.com/en/3.6.3/install.html#optional-packages).

If you like, you can create a note at the end of [here](http://docs.getpelican.com/en/3.6.3/quickstart.html#create-a-project).

nstinus:
I didn't search the docs but the source ;)
As suggested, I added a footnote in quickstart's doc.

leotrs:
Looks good to me.

justinmayer:
Thanks for the contribution, Nicolas. Do you think it makes sense to add test(s) for these additions? Also, once done, please [rebase & squash your commits](https://github.com/getpelican/pelican/wiki/Git-Tips#squashing-commits).

justinmayer:
Hi @nstinus. Just checking in regarding my last comment. Have a moment to respond so we can get this merged?

nstinus:
@justinmayer: I don't really see the point of adding tests for that.

justinmayer:
Thanks for the enhancement, Nicolas.

PR #1973, opened 2016-06-04
Allow for atom and rss feeds to use :summary: attributes.

TierraDelFuego:
Cancelled my previous pull request and have followed https://github.com/getpelican/pelican/wiki/Git-Tips.

I have tested the functionality and it is working correctly.

I ran unittests but the output generated was unreadable and unrelated to my changes.


avaris:
First, missing documentation for the new setting.
Second, why is there a catch-all `try/except`?

ingwinlu:
also missing testcases. a 'I tested the functionality' is nice for you. But changes to the codebase in the future might break this feature and without automated tests we would not necessarily find out about it right away.

TierraDelFuego:
Ok, good feedback, thanks for taking the time.

PR #1975, opened 2016-06-09
Check full paths when excluding files

logston:
The `Generator._include_path` method needs to be updated because it
does not filter files with enough specificity. For example, let's say
our Pelican project is devoted to hosting article about talks a
different conferences. If we wanted to filter a talk
at `content/jacksconf-2016/videos/jacks-talk.json`, we could do so
(without using this patch) by including `jacks-talk.json` in the
list IGNORE_FILES. However, if there are multiple `jacks-talk.json`
files (maybe a bunch of different Jacks wanted to talk at a bunch of
different conferences), we would exclude all `jacks-talk.json` files
by including the single `jacks-talk.json` string in IGNORE_FILES. By
checking the full path passed to `_include_path`, we can filter on a
more granular level.

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -114,6 +114,15 @@ Setting name (followed by default value, if any)                                
                                                                                  of these patterns will be ignored by the processor. For example,
                                                                                  the default ``['.#*']`` will ignore emacs lock files, and
                                                                                  ``['__pycache__']`` would ignore Python 3's bytecode caches.
+                                                                                 Longer patterns can be used to filter files and directories with
+                                                                                 more granularity. For example, if multiple files have the same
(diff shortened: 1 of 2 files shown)


avaris:
Isn't the path relative to `content`? Why do you include it in the docs?

Also, this is potentially a breaking change. Say, I had a filter before to ignore all files that start with `t`, like `'t*'`. Now it will also ignore everything in a top-level folder that has a name starting with `t`.

logston:
Sorry for the month delay. Yes, I do see that this could be a breaking change. Thanks for pointing that out  @justinmayer. Should this be merged at a major version instead then? Or should it be merged at all?

Also, I've updated the documentation. Thanks for that catch as well.

PR #1978, opened 2016-06-26
Respect OUTPUT_RETENTION for Fabric "rebuild" task

### LINKED ISSUE #1955: Fabric "rebuild" task ignores OUTPUT_RETENTION
In the generated fabric file, `rebuild` calls `clean` before calling `build`. `clean` wipes out the output directory. `clean` ignores `OUTPUT_RETENTION` because it just calls `shutil.rmtree(DEPLOY_PATH)`. 

The issue is compounded because `gh_pages` calls rebuild. This removes the `.git` directory and causes `gh_pages` to generate an error.

I propose that we change `rebuild` to call the builtin pelican cleaning system. Changing `rebuild` to:

```python
def rebuild():
    """`build` with the delete switch"""
    local('pelican -d -s pelicanconf.py')
```

bnice5000:
Fixes #1955

### DIFF
--- a/pelican/tools/templates/fabfile.py.in
+++ b/pelican/tools/templates/fabfile.py.in
@@ -37,9 +37,8 @@ def build():
     local('pelican -s pelicanconf.py')
 
 def rebuild():
-    """`clean` then `build`"""
-    clean()
-    build()
+    """`build` with the delete switch"""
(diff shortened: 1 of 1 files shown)


justinmayer:
Thanks for the fix, Brian.

Issue #1980, opened 2016-07-10
(feature request) Offsite image links without hardcoded urls -- {imgroot}


mlabbe:
I put all of my Blog images on Amazon servers, but host the blog on my own server.  Images are offsite.

Currently, each time I add a blog image, I have to type a wordy offsite, fully formed URL.  I also have to sync to s3 to see my images even though I haven't published my content.

It would be wonderful if there was a configuration defined value that let me specify a base URL for all image links that *may or may not be relative* to my blog in my markdown.  {imgroot} would be one possibility.

I would assign {imgroot} to be the same as {filename} (possibly the default option) on local builds of my site and then assign it to the s3 http URL on remote builds.  

Thanks for your hard work on Pelican!  I have been using it for 3 years and I really appreciate it!

adeverteuil:
Hi,

I think this plugin can do what you want: https://github.com/getpelican/pelican-plugins/tree/master/interlinks

leotrs:
@mlabbe Please check out @adeverteuil's suggestion and close or extend as necessary. Thanks!

mlabbe:
Thank you for the plugin suggestion!  I deployed this change at frogtoss.com/labs and it seems to work perfectly!

PR #1982, opened 2016-08-06
Add static file options: hard/symlink & only-when-modified

adeverteuil:
Hello,

I am interested in improving `StaticGenerator`. Please let me know if a similar effort is already under way, and also bring your recommendations and ideas. I don't have a timeline but I'd really like to have at least a basic working feature in about 2 weeks given the free time that I have.

The problem is when the site content has several large static files, like videos. When generating the site, all the static files are copied, regardless of modification time. Even `WRITE_SELECTED` doesn't prevent that. For example, the pages and article take 4 seconds to process, but the static files add another almost 30 seconds to it.

I'm thinking of adding a setting `STATIC_CHECK_MODIFIED`, false by default (the current behaviour). I'm currently working on inserting the logic in the `StaticGenerator.generate_context()` method.

Once it works well with mtimes, then maybe I can add a `STATIC_CHECK_MODIFIED_METHOD` settings which would be "mtime" by default, or the name of a hashlib function. But I doubt that computing hashes on two video files (source_path and save_as) would be faster than just copying one over the other.

Then I also want to try symlinks or hardlinks from save_as to source_path. I imagine that would be fast and even space efficient. Do you think of any problems that could arise if the output files are links and not copies?

I also think #1980 is a great idea which could also solve the problem and I might look into it.

Sincerely,
Alexandre de Verteuil

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -230,6 +230,20 @@ Basic settings
    ``PAGE_PATHS``. If you are trying to publish your site's source files,
    consider using the ``OUTPUT_SOURCES`` setting instead.
 
+.. data:: STATIC_CREATE_LINKS = False
+
+   Create links instead of copying files. If the content and output
+   directories are on the same device, then create hard links.  Falls
+   back to symbolic links if the output directory is on a different
+   filesystem. If symlinks are created, don't forget to add the ``-L``
+   or ``--copy-links`` option to rsync when uploading your site.
+
+.. data:: STATIC_CHECK_IF_MODIFIED = False
+
+   If set to ``True``, and ``STATIC_CREATE_LINKS`` is ``False``, compare
+   mtimes of content and output files, and only copy content files that
+   are newer than existing output files.
+
 .. data:: TYPOGRIFY = False
 
    If set to True, several typographical improvements will be incorporated into
--- a/pelican/generators.py
+++ b/pelican/generators.py
@@ -2,6 +2,7 @@
 from __future__ import print_function, unicode_literals
 
 import calendar
+import errno
 import fnmatch
 import logging
 import os
@@ -20,9 +21,8 @@ from pelican import signals
 from pelican.cache import FileStampDataCacher
 from pelican.contents import Article, Draft, Page, Static, is_valid_content
 from pelican.readers import Readers
-from pelican.utils import (DateFormatter, copy, copy_file_metadata, mkdir_p,
-                           posixize_path, process_translations,
-                           python_2_unicode_compatible)
+from pelican.utils import (DateFormatter, copy, mkdir_p, posixize_path,
+                           process_translations, python_2_unicode_compatible)
 
 
 logger = logging.getLogger(__name__)
@@ -682,20 +682,8 @@ class StaticGenerator(Generator):
 
     def __init__(self, *args, **kwargs):
         super(StaticGenerator, self).__init__(*args, **kwargs)
+        self.fallback_to_symlinks = False
         signals.static_generator_init.send(self)
-
-    def _copy_paths(self, paths, source, destination, output_path,
-                    final_path=None):
-        """Copy all the paths from source to destination"""
-        for path in paths:
-            if final_path:
-                copy(os.path.join(source, path),
-                     os.path.join(output_path, destination, final_path),
-                     self.settings['IGNORE_FILES'])
-            else:
-                copy(os.path.join(source, path),
-                     os.path.join(output_path, destination, path),
-                     self.settings['IGNORE_FILES'])
 
     def generate_context(self):
         self.staticfiles = []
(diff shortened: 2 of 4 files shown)


adeverteuil:
Hm, I think something is not right in the continuous integration check for Pythons versions 3.3 and 3.4. They are actually run in Python 2.7, as you can see in the [Details](https://travis-ci.org/getpelican/pelican/builds/150236397) above. In the failing jobs logs, you can also see the python version used in the test's virtualenv is 2.7:

```
$ python --version
Python 2.7.9
```

Please let me know if there is something I can do.

avaris:
No, that's the python used to create virtualenvs.

```
py33 runtests: commands[0] | /home/travis/build/getpelican/pelican/.tox/py33/bin/python --version
Python 3.3.5
```

adeverteuil:
I learned how to use `pyenv` and installed the supported Python versions on my computer, then fixed the failing unit test. The rest of this comment is just for curiosity.

I hit a change of behavior between Python 3.4 and 3.5. I didn't find anything about that in the Python changelogs though, but it looks like a bugfix since the documentation of `codecs.open()` says the `"r"` mode is the default, so it should work the same with or without it explicitly specified. Did anyone know about this discrepency? Here's a small script to reproduce it, and the output.

```
# unexplained.py
import codecs
f = open("f", "wt")
f.write("content")
f.close()

f = codecs.open("f")
print(type(f))
f.close()

f = codecs.open("f", "r")
print(type(f))
f.close()
```

```
alex@meson:pelican$ for v in 3 4 5; do python3.$v < unexplained.py; done                                                                                     
<class '_io.BufferedReader'>
<class '_io.TextIOWrapper'>
<class '_io.BufferedReader'>
<class '_io.TextIOWrapper'>
<class '_io.TextIOWrapper'>     # <--- This is not like the others
<class '_io.TextIOWrapper'>
alex@meson:pelican$ pyenv versions
  system
* 3.3.5 (set by /home/alex/.pyenv/version)
  3.3.6
* 3.4.2 (set by /home/alex/.pyenv/version)
  3.4.5
* 3.5.0 (set by /home/alex/.pyenv/version)
```

In Python 3.5, `codec.open()` without the `"r"` parameter returns a `TextIOWrapper`. To fix the problem I added the `"r"` parameter so the same class of reader is returned in all 3 versions of Python.

Version 3.4.5 works the same as 3.5.0, but 3.3.6 still returns `BufferedReader` without `"r"`.

justinmayer:
Any @getpelican/reviewers have a moment to review this pull request?

docs/settings.rst
```diff
@@ -168,6 +168,15 @@ Setting name (followed by default value, if any)
                                                                                  is also in ``ARTICLE_PATHS`` or ``PAGE_PATHS``. If you are trying
                                                                                  to publish your site's source files, consider using the
                                                                                  ``OUTPUT_SOURCES`` setting instead.
+``STATIC_CREATE_LINKS = False``                                                  Create links instead of copying files. If the content and output
```
almet: I would name it `STATIC_CREATE_STATIC_LINKS` as it explains better the intent.

almet: I would name it `STATIC_CREATE_STATIC_LINKS` as it explains better the intent.

docs/settings.rst
```diff
@@ -168,6 +168,15 @@ Setting name (followed by default value, if any)
...
+                                                                                 directories are on the same device, then create hard links.
+                                                                                 Falls back to symbolic links if the output directory is on a
+                                                                                 different filesystem. If symlinks are created, don't forget to
+                                                                                 add the ``-L`` or ``--copy-links`` option to rsync when uploading
+                                                                                 your site.
+``STATIC_CHECK_MODIFIED = False``                                                If set to ``True``, and ``STATIC_CREATE_LINKS`` is ``False``,
```
almet: `STATIC_CHECK_IF_MODIFIED` ?

almet: `STATIC_CHECK_IF_MODIFIED` ?

pelican/generators.py
```diff
@@ -680,21 +681,9 @@ class StaticGenerator(Generator):
 
     def __init__(self, *args, **kwargs):
         super(StaticGenerator, self).__init__(*args, **kwargs)
+        self.create_symlinks = False
```
almet: is it a hardlink or a symlink?

pelican/generators.py
```diff
@@ -722,13 +712,91 @@ def generate_output(self, writer):
...
+            return self._source_is_newer(staticfile)
+
+    def _source_is_newer(self, staticfile):
+        source_path = os.path.join(self.path, staticfile.source_path)
+        save_as = os.path.join(self.output_path, staticfile.save_as)
+        s_mtime = os.path.getmtime(source_path)
+        try:
+            d_mtime = os.path.getmtime(save_as)
+        except OSError:
+            return True
```
almet: Do you know when this happens? Should we log something?

adeverteuil: Good catch. This is the case where the destination file doesn't exist. The `try` block is redundant; this case is already checked in the `_file_update_required()` method. I removed the `try` and updated the `…_when_dest_does_not_exist` unit test to cover the latter method instead.

docs/settings.rst
adeverteuil: Hm, I think you're thinking of statically linked libraries, no? How about `STATIC_CREATE_HARDLINKS`?

pelican/generators.py
adeverteuil: This is a flag that is set to `True` in `_link_staticfile()` the first time an "Invalid cross-device link" exception is catched. From then on we stop attempting to create hard links and we create soft links instead. Maybe I should rename it `self._cross_device_link_exception_catched`, what do you think?

docs/settings.rst
adeverteuil: Yes, I agree.

docs/settings.rst
adeverteuil: I remember why I didn't call it `STATIC_CREATE_HARDLINKS` in the first place. That's because if the content and output directories are on different filesystems, then the program will fallback to creating symbolic links so I picked a more generic name.

almet: Okay, after a second look, that makes sense to me :)

pelican/generators.py
almet: Name it "can_create_symlinks"?

almet:
Okay cool, we're almost there!

pelican/generators.py
adeverteuil: Well, we **can always** create symlinks, it's just that sometimes we **must**. `fallback_to_symbolic_links`?

adeverteuil:
I had to merge upstream/master into the feature branch to make checks pass again and to resolve conflicts. Should I rebase or squash or is it okay as-is?

justinmayer:
Once code reviews are marked as approved, we usually request that you [rebase & squash your commits](https://github.com/getpelican/pelican/wiki/Git-Tips#squashing-commits).

@almet: Any follow-up comments on this pull request?

adeverteuil:
Ok, I understand how to rebase and squash my commits. Is it okay to force push the changes on Github after doing so? Can or should I assume nobody cloned that feature branch from my fork?

I'm 90% sure I understand what I have to do but this is my first contribution to a public project and it goes against everything I read about changing version history once it has been pushed to a public repo.

adeverteuil:
Hi, any comments about the changes? There's no rush, but I'm looking forward to some free time during Christmas holidays, so it would be good timing to finish this in the coming weeks. Please also take a moment to address my question in the last comment.

I hope you are well. Thank you for your time and such a great project. Still use it for my blog and it's fun.

almet:
Everything seems good here! Squashing is the way to go, don't be afraid (especially because you're in a branch). This is done in order to keep the history easier to read.

Thanks for your time and additions :)

justinmayer:
Hi Alexandre. As Alexis suggested, please [rebase & squash your commits](https://github.com/getpelican/pelican/wiki/Git-Tips#squashing-commits). Thanks again for all your work on this!  (^_^)

adeverteuil:
Hi @justinmayer, @almet, here you go: rebased and squashed properly. I hope the users will like it!

justinmayer:
Many thanks, Alexandre! Once we ship 3.7.1 (a minor bug-fix release), I'll add this to the 3.8 milestone and merge.

justinmayer:
Many thanks to @adeverteuil for the excellent contribution and to everyone who helped review it.  (^_^)

cpitclaudel:
Would there be anything wrong doing the same thing for theme files?

avaris:
I can't think of a reason.

PR #1989, opened 2016-08-08
Use the summary as the feed's summary

### LINKED ISSUE #1886: Atom feed should output both <summary> and <content>
The Atom feed seems to only spit out a `<summary>` tag.  Currently I have two options: populate it with summaries (which means the feed won't contain the full articles in any form, which is really annoying) or populate it with full articles (which is arguably against spec, and also means feed readers can't show a brief summary, which is also really annoying).

Unfortunately, from a brief glance through feedgenerator, it doesn't seem to support the `<content>` tag at all (!).

eevee:
This fixes #1886.  I tried this on my own blog and the results are definitely more pleasing: `<summary>` contains a summary, `<content>` contains the full content.

As mentioned in that issue, this _does_ change the behavior for RSS.  RSS's `<description>` is intended as a summary, and it has no equivalent to `<content>`, so with this change, RSS feeds won't contain the full content any more.

I can add an option to force putting the full content in the summary/description tag, if this is a problem.  Alternatively, just use Atom.  ;)

(For what it's worth, I checked my new feed in feedly, and it appears to understand `<content>` and still show the full article.  I don't know how other readers behave.)


iKevinY:
Looks good to me; can we get this merged in, @justinmayer?

ingwinlu:
lgtm

justinmayer:
I just ran `python -m unittest discover` and am seeing three failing tests on current master (d6ac93a), with at least one failure being seemingly related to `<updated>` tags being expected but absent in the output. I'm seeing similar failures when running on the code from this PR. This is with FeedGenerator 1.9 on Mac OS 10.11.6 and Python 2.7.11. Anyone else seeing this and/or have any insight into what's causing them?

Just thought it would be good to resolve this before proceeding.  (^_^)

iKevinY:
> Anyone else seeing this and/or have any insight into what's causing them?

Hmm, I'm also seeing test failures for `master`: four when running `python -m unittest discover` and three when running `nose`. (feedgenerator 1.7, Python 2.7.10, OS X 10.11.6)

eevee:
Merged with master.

The `test_error_on_warning` test is now failing on my machine, but it's also failing on master, and Travis seems happy.  Odd.

iKevinY:
I noticed failing tests on my machine too, but I updated my dependencies and everything seems to pass now (when running `python -m unittest discover` at least). I think it might have to do with the fact that Pygments is pinned to 2.1.3 now?

avaris:
About `test_error_on_warning`, see: #2007 

Test runner in `unittest` is annoyingly overriding warning filters, so you need the `-Wd` flag to stop that.

Other test runners doesn't have that issue. Hence the happy travis (which uses `nose` instead).

iKevinY:
Interestingly, I _do_ get failures with `nose`, but not with the base `python -m unittest discover`.

justinmayer:
Merged via eac8fa2abd91823f3ffbdc679147504ee9489618. Many thanks to @eevee for working on this and to all the folks who helped review.

Issue #1568, opened 2014-12-26
DEFAULT_ORPHANS inconsistent with documentation


kernc:
The documentation for `DEFAULT_ORPHANS` setting says:
> The minimum number of articles allowed on the last page. Use this when you don’t want the last page to only contain a handful of articles.

However, with
```
DEFAULT_PAGINATION = 1
DEFAULT_ORPHANS = 1
```
and with **three articles** paginated, one appears on the first page and two on the second. The same setup with `DEFAULT_PAGINATION = 2` produces a single, unpaginated page.

So it seems that `(DEFAULT_ORPHANS + 1)` is the _minimum number of articles allowed on the last page_.

leotrs:
@kernc can you still replicate this?

kernc:
Seems so. With pelican git, given three articles in a paginated category,
```text
DEFAULT_PAGINATION = 1
DEFAULT_ORPHANS = 1
```
results in two pages (1+2). Expected are three pages (1+1+1).

Likewise,
```text
DEFAULT_PAGINATION = 2
DEFAULT_ORPHANS = 1
```
results in a single, unpaginated page. Expected are two pages (2+1).

Both of these behaviors are inconsistent with docs for `DEFAULT_ORPHANS`.

Setting in both above cases `DEFAULT_ORPHANS = 0` results in expected behavior (as noted above). This seems to indicate `DEFAULT_ORPHANS` is really interpreted as `DEFAULT_ORPHANS + 1`.

leotrs:
Sounds like a change in documentation could solve this issue. PRs are welcome.

kernc:
That would be the easy and probably invalid fix. Got something better in https://github.com/getpelican/pelican/pull/1992, I hope. :smiley:

kernc:
Resolution: invalid https://github.com/getpelican/pelican/pull/1992#issuecomment-239975458.
Thanks.

PR #1992, opened 2016-08-14
Fix edge condition with DEFAULT_PAGINATION and DEFAULT_ORPHANS

kernc:
Fixes https://github.com/getpelican/pelican/issues/1568

### DIFF
--- a/pelican/paginator.py
+++ b/pelican/paginator.py
@@ -35,7 +35,7 @@ class Paginator(object):
         "Returns a Page object for the given 1-based page number."
         bottom = (number - 1) * self.per_page
         top = bottom + self.per_page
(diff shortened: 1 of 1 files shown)


leotrs:
Please make sure the tests pass to move forward with review.

kernc:
This is unsolvable.

If I have 5 posts and DEFAULT_PAGINATION=2 and DEFAULT_ORPHANS=2, I can either have one page render with three posts (and thus somewhat not honor DEFAULT_PAGINATION), or have it with a single post (and thus not honor DEFAULT_ORPHANS).

For what I care, these edge cases ... I don't.

Thanks! :smiley:

PR #1996, opened 2016-08-19
Fabric file not compatible with python 3

bnice5000:
Python 3 changes the name of SocketServer to socketserver. This fixes the naming error.

### DIFF
--- a/pelican/tools/templates/fabfile.py.in
+++ b/pelican/tools/templates/fabfile.py.in
@@ -3,7 +3,11 @@ import fabric.contrib.project as project
 import os
 import shutil
 import sys
-import SocketServer
+
+if sys.version_info.major == 2:
+    import SocketServer
+elif sys.version_info.major == 3:
+    import socketserver as SocketServer
 
 from pelican.server import ComplexHTTPRequestHandler
 


avaris:
Umm, `fabric` doesn't support python 3?

bnice5000:
Yes, but there is a very popular fork of fabric that does support python 3.

https://pypi.python.org/pypi/Fabric3
https://github.com/mathiasertl/fabric/

Totktonada:
@justinmayer If you want more extrinsic opinions, this is mine.

Why not to improve Python 3 compatibility if there are no any cons? It would be good addition to serve() implementation from #1200. It works for me on Python 3.5 when being injected into `pelican-quickstart` generated configuration.

However I would stick to Python 3 names instead of Python 2 like `six` package do. Next, it’s possible to just use `from six.moves import socketserver`. It seems that pelican already depends on `six`.

And there are some thoughts little out of the topic, but related. It seems that fabfile reuses code from `server.py` from pelican, but duplicates its code under `if __name__ == '__main__'`. It's possible to wrap this code into `serve` function in `server.py`, import it and just use. Would it be better, any thoughts?

iKevinY:
I think #2184 resolved this issue.

justinmayer:
Indeed. Thanks, Kevin.

PR #1999, opened 2016-08-29
Allow setting Jinja environment arguments from settings

rjames86:
Allow the ability to set custom Jinja environment arguments from within the settings. One consideration is that this setting could also be used in favor of `"JINJA_EXTENSIONS"` by just setting the key in the dictionary.

One potential downside is that someone could override the arguments already being set [here](https://github.com/getpelican/pelican/blob/master/pelican/generators.py#L64). We could whitelist/blacklist arguments that need to stay as the defaults.

### DIFF
--- a/pelican/settings.py
+++ b/pelican/settings.py
@@ -109,8 +109,12 @@ DEFAULT_CONFIG = {
         },
         'output_format': 'html5',
     },
-    'JINJA_EXTENSIONS': [],
     'JINJA_FILTERS': {},
+    'JINJA_ENVIRONMENT': {
+        'trim_blocks': True,
+        'lstrip_blocks': True,
+        'extensions': [],
+    },
     'LOG_FILTER': [],
     'LOCALE': [''],  # defaults to user locale
     'DEFAULT_PAGINATION': False,
@@ -162,6 +166,12 @@ def read_settings(path=None, override=None):
                            'PLUGIN_PATHS, moving it to the new setting name.')
             local_settings['PLUGIN_PATHS'] = local_settings['PLUGIN_PATH']
             del local_settings['PLUGIN_PATH']
+        if 'JINJA_EXTENSIONS' in local_settings:
+            logger.warning('JINJA_EXTENSIONS setting has been deprecated, '
+                           'moving it to JINJA_ENVIRONMENT setting.')
+            local_settings['JINJA_ENVIRONMENT']['extensions'] = \
+                local_settings['JINJA_EXTENSIONS']
+            del local_settings['JINJA_EXTENSIONS']
         if isinstance(local_settings['PLUGIN_PATHS'], six.string_types):
             logger.warning("Defining PLUGIN_PATHS setting as string "
                            "has been deprecated (should be a list)")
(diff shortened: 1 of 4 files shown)


justinmayer:
Thanks for the contribution, Ryan. It looks like you may have forgotten to run `flake8` on the files you changed, which ensures that PyCodeStyle guidelines are followed. This is something our test suite checks for, as you can see by the failing check in this pull request.

@avaris / @ingwinlu / @leotrs: Any thoughts or comments?

avaris:
If we are going this way, we might as well deprecate `JINJA_EXTENSIONS` and set the predetermined values (`trim_blocks`, `lstrip_blocks`) as defaults in this setting.

rjames86:
@avaris Sounds good. I figured that would be the best way forward, but wanted to check to be sure. I'll get a change out for that

rjames86:
One question: I've removed the `JINJA_EXTENSIONS` setting and updated the docs. Are there any other places that I need to update for deprecated settings variables?

avaris:
I think the wording on that warning might be confusing. To me, it sounds like only the setting name is changed, but it's a completely new setting with a different context. Maybe something like:

```
JINJA_EXTENSIONS setting has been deprecated, moving it to JINJA_ENVIRONMENT setting.
```

justinmayer:
@avaris: Any comments on Ryan's latest changes? Anything else here that you think needs to be addressed?

pelican/settings.py
```diff
@@ -212,6 +222,18 @@ def get_settings_from_file(path, default_settings=DEFAULT_CONFIG):
     return get_settings_from_module(module, default_settings=default_settings)
 
 
+def get_jinja_environment(settings):
+    """Sets the environment for Jinja"""
+
+    jinja_env = settings.get('JINJA_ENVIRONMENT', {})
```
avaris: Slight issue. `.get` here doesn't do anything if the default is returned. You'll get an empty dict which is then populated but nothing will happen to the `settings`. If anything, `.setdefault` should be used instead.

Since any settings will use `DEFAULT_CONFIG` as base, there should be a `'JINJA_ENVIRONMENT'` key regardless, but still :).

justinmayer:
@rjames86: Could you take another pass based on the feedback from @avaris?

rjames86:
@justinmayer yep will do. was planning on doing it this weekend

pelican/settings.py
```diff
@@ -212,6 +222,21 @@ def get_settings_from_file(path, default_settings=DEFAULT_CONFIG):
     return get_settings_from_module(module, default_settings=default_settings)
 
 
+def get_jinja_environment(settings):
+    """Sets the environment for Jinja"""
+
+    jinja_env = settings['JINJA_ENVIRONMENT']
```
avaris: This is unnecessary and the cause of test failures. `.setdefault` will return the value if it exists, else sets the value and returns it.

rjames86: Whoops. Sorry, that's my fault. That wasn't supposed to be there

avaris:
This is good to go once you [squash](https://github.com/getpelican/pelican/wiki/Git-Tips#squashing-commits) it :).

rjames86:
@avaris just wanted to bubble this review up. Everything should be good to go here. I've updated the settings.rst since there was a big refactor

justinmayer:
Many thanks to @rjames86 for the contribution and to @avaris for the review.  (^_^)

PR #2001, opened 2016-08-30
Allow to ignore warnings by id

Totktonada:
Useful when a message contains %s placeholders. Inspired by #1594.

That's my first contribution in Pelican, so code style / naming in the patch can be not so good. Any remarks are very welcome!

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -107,7 +107,7 @@ Setting name (followed by default value, if any)                                
                                                                                  and the message to be ignored.
                                                                                  For example: ``[(logging.WARN, 'TAG_SAVE_AS is set to False')]``
 ``LOG_FILTER_IDS = []``                                                          A list of containing IDs of messages to be ignored.
-                                                                                 For example: ``['warn.img.alt.empty']``
+                                                                                 For example: ``['warn.readers.empty-alt-attribute']``
 ``READERS = {}``                                                                 A dictionary of file extensions / Reader classes for Pelican to
                                                                                  process or ignore. For example, to avoid processing .html files,
                                                                                  set: ``READERS = {'html': None}``. To add a custom reader for the
@@ -780,7 +780,7 @@ It's hard to ignore a group of messages containing dynamic substrings (e.g.
 paths to files). Such messages can be ignored using the ``LOG_FILTER_IDS``
 setting.
 
-For example: ``['warn.img.alt.empty']``
+For example: ``['warn.readers.empty-alt-attribute']``
 
 .. _reading_only_modified_content:
 
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -106,6 +106,8 @@ Setting name (followed by default value, if any)                                
 ``LOG_FILTER = []``                                                              A list of tuples containing the logging level (up to ``warning``)
                                                                                  and the message to be ignored.
                                                                                  For example: ``[(logging.WARN, 'TAG_SAVE_AS is set to False')]``
+``LOG_FILTER_IDS = []``                                                          A list of containing IDs of messages to be ignored.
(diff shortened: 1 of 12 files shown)


Totktonada:
Ouch, sorry for the splitting it into two commits. Little mess in my local copy :(

In my case it allows to don't suppress all warnings by `-q` option and ignore the specific one. It useful for me when debugging new website features (e.g. when writing new python-markdown plugins).

Constantly get the same warnings isn't option (even when its are limited to 5) because it obfuscate others in regular usage and visually hide warnings that really matter, when something going bad.

Totktonada:
It looks like your contribution guides suggest to squash such series of commits to one and force-push it back to feature branch. Should I really rewrite the history? I'm not sure.

pelican/log.py
```diff
@@ -100,6 +101,11 @@ def filter(self, record):
         if record.levelno > logging.WARN:
             return True
 
+        # filter record if user ignored it by id
+        id = record.__dict__.get('id', None)
+        if id in self._ignore_ids:
+            return False
+
```
iKevinY: If you want to do a variable assignment here, it might be better to use `_id` as a name to avoid shadowing Python's `id` built-in. Otherwise, you could of course just do this (the second argument to `.get()` defaults to `None` if one isn't explicitly given):

```python
if record.__dict__.get('id') in self._ignore_ids:
    return False
```

iKevinY:
@Totktonada Squashing and force-pushing is fine (and I would even say encouraged for small fix-up commits). Since a pull request operates on your own fork of a repository, rewriting history is fine because you're the only one pulling and pushing.

docs/settings.rst
```diff
@@ -774,6 +776,11 @@ be filtered out.
 
 For example: ``[(logging.WARN, 'TAG_SAVE_AS is set to False')]``
 
+It's hard to ignore a group of messages containing dynamic substrings (e.g.
+paths to files). Such messages can be ignored using ``LOG_FILTER_IDS`` setting.
```
iKevinY: I think this would read better with a "the" before "`LOG_FILTER_IDS`".

iKevinY:
Thanks for the contribution to Pelican, @Totktonada! Your PR look good to me (+1 for updating the documentation as well). Thoughts, @justinmayer @avaris @ingwinlu?

Totktonada:
@iKevinY, thanks for review! I fixed the code and the docs according to your remarks and squashed all fixup commits.

avaris:
As far as I can see, this only allows to suppress one type of warning (`empty alt attr on img`), but you documented it as if it's a more general feature than it really is.

Totktonada:
@avaris, yes. I forgot to mention this, sorry. Are there list of warnings that user likely want to suppress? I think there are no need to add IDs to all messages (and I doubt whether I can do it w/o good Pelican code understanding).

Totktonada:
I assigned IDs to all debug/info/warn messages, 76 total.

There are messages with one ID in the following cases:

* The messages form one multiline message.
* The messages looks as arriving in a pretty similar situations (at least I failed to distinguish).

I tried to assign meaningful IDs, but I have no strong understanding of the code, so I'm not sure it's good in all cases. I'm give command to obtain all IDs in the commit message, so it can be easily checked (I hope).

There are problems in some cases with too long IDs. So I stripped out '-deprecated' suffix where long settings names were used. I'm think a setting name is more important.

I'm marked #1594 (where initial discussion was started) as fixed in this commit.

Hope it increases odds of this feature to come to the next release :)

Totktonada:
Any thoughts?

Totktonada:
As discussed in IRC with @avaris, the approach originally proposed in #1594 is more clear. I'm closed this PR and transform the issue #1594 to a PR.

PR #1594, opened 2016-09-12
LOG_FILTER cannot be used with '%s'ed messages

Totktonada:
Currently LOG_FILTER doesn’t allow ignoring messages with '%s' placeholders, because `ignore_key` variable in `filter` method of `log` module use 'getMessage()' method of `LogRecord` class, but not `msg` field.

Example which isn’t work:

```
LOG_FILTER = [
    (logging.WARN, 'Empty alt attribute for image %s in %s')
]
```

The problem is that now no way to suppress such warnings. I guess, filtering by `msg` field doesn’t break anything, but allow to more complex filtering.

Or maybe better intruduce another varible, say LOG_FILTER_GROUP, with such functionality.

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -777,6 +777,16 @@ be filtered out.
 
 For example: ``[(logging.WARN, 'TAG_SAVE_AS is set to False')]``
 
+It is possible to filter out messages by a template. Check out source code to
+obtain a template.
+
+For example: ``[(logging.WARN, 'Empty alt attribute for image %s in %s')]``
+
+**Warning:** Silencing messages by templates is a dangerous feature. It is
+possible to unintentionally filter out multiple message types with the same
+template (including messages from future Pelican versions). Proceed with
+caution.
+
 .. _reading_only_modified_content:
 
 
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -787,6 +787,8 @@ possible to unintentionally filter out multiple message types with the same
 template (including messages from future Pelican versions). Proceed with
 caution.
 
+Note: This option does nothing ``--debug`` is passed.
+
 .. _reading_only_modified_content:
 
 
--- a/pelican/log.py
+++ b/pelican/log.py
@@ -112,12 +112,12 @@ class LimitFilter(logging.Filter):
         else:
             self._raised_messages.add(message_key)
 
-        # ignore LOG_FILTER records
-        # use .msg and not .getMessage() for string formatting to allow
-        # filtering by templates
-        ignore_key = (record.levelno, record.msg)
-        if ignore_key in self._ignore:
-            return False
+        # ignore LOG_FILTER records by templates when "debug" isn't enabled
+        logger_level = logging.getLogger().getEffectiveLevel()
+        if logger_level > logging.DEBUG:
+            ignore_key = (record.levelno, record.msg)
+            if ignore_key in self._ignore:
+                return False
 
         # check if we went over threshold
         if group:
--- a/pelican/log.py
+++ b/pelican/log.py
@@ -92,6 +92,7 @@ class LimitFilter(logging.Filter):
     """
 
     _ignore = set()
+    _raised_messages = set()
     _threshold = 5
     _group_count = defaultdict(int)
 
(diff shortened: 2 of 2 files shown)


Totktonada:
Travis CI seems failed after recent feedgenerator version bump. As I saw <update/> tag added in Atom feeds that point me to getpelican/feedgenerator@87ff59c2. I guess reference output for the tests should be regenerated.

Totktonada:
Rebased at top of current master and force-pushed to force Travis CI to rebuild it.

pelican/log.py
```diff
@@ -92,10 +92,15 @@ class LimitFilter(logging.Filter):
...
     _ignore = set()
+    _raised_messages = set()
     _threshold = 5
     _group_count = defaultdict(int)
 
     def filter(self, record):
+        # don't filter log messages when "debug" is enabled
+        logger_level = logging.getLogger().getEffectiveLevel()
+        if logger_level == logging.DEBUG:
+            return True
```
Scheirle: In debug mode now all messages get through even the ones which previously got filter because of a threshold.
(Or did I miss something?)

---

Also I don't understand the intention behind this change.
I see myself only using this feature **in** debug mode to reduce the logging flood.

Totktonada: You right. We discussed in IRC a "collision" situation when several messages have one template. I proposed lightweight "protection" with enabling all silenced (LOG_FILTER) messages when `--debug` passed to help user digging into a problem when it arised. My thought was threshold-based filtering can hide potential problems as well as template-based filtering.

I'm not sure what is better, so will give back original behaviour for threshold-based filtering.

Krastanov: > Also I don't understand the intention behind this change.
> I see myself only using this feature in debug mode to reduce the logging flood.

I would actually want to use this filtering in production, not debug mode. There is at least one warning type that I consider wrong (the one about empty `alt` attributes for images, as there are good reasons to have empty `alt`s), and I would like to have a way to disable it. It is a string-interpolated warning, hence the current LOG_FILTER does not work on them, but this PR would solve that problem.

Totktonada: I’m fixed last commit and now it doesn’t interfer with threshold-based filtering. LOG_FILTER works only in production mode and disabled in debug one. It seems that now PR fit both your requirements.

docs/settings.rst
```diff
@@ -774,6 +774,16 @@ be filtered out.
 
 For example: ``[(logging.WARN, 'TAG_SAVE_AS is set to False')]``
 
+It's possible to filter out messages by a template. Check out source code to obtain a template.
+
+For example: ``[(logging.WARN, 'Empty alt attribute for image %s in %s')]``
+
+**Warning:** Silence messages by templates is dangerous feature. There are odds
+to unintantional filter out messages of different types with the same template
+(includes future messages from newer Pelican versions). Proceed with caution.
```
avaris: When I said "suppress future messages", I didn't mean any *new* warnings with same template (although quite unlikely, that is still a possibility). But I mean the same warning for future runs. Let's say you did the above and silenced `alt` warnings. You'll *never* see that warning again. It might be OK for now, since you made sure every warned image is supposed to have an empty alt. But that warning might be relevant in the future for new articles/pages.

avaris: When I said "suppress future messages", I didn't mean any *new* warnings with same template (although quite unlikely, that is still a possibility). But I mean the same warning for future runs. Let's say you did the above and silenced `alt` warnings. You'll *never* see that warning again. It might be OK for now, since you made sure every warned image is supposed to have an empty alt. But that warning might be relevant in the future for new articles/pages.

avaris: When I said "suppress future messages", I didn't mean any *new* warnings with same template (although quite unlikely, that is still a possibility). But I mean the same warning for future runs. Let's say you did the above and silenced `alt` warnings. You'll *never* see that warning again. It might be OK for now, since you made sure every warned image is supposed to have an empty alt. But that warning might be relevant in the future for new articles/pages.

justinmayer:
@Totktonada: Any thoughts on Avaris's comments?

@Krastanov / @Scheirle: Any comments on the latest changes?

Totktonada:
@justinmayer I just wrote about it in IRC. In short: I don’t quite understand the Avaris’s request and don’t know what can I do to make messages’ filtering more intuitive.

Now if user requested a message to be silenced, it silenced except when `--debug` passed.

docs/settings.rst
```diff
@@ -774,6 +774,16 @@ be filtered out.
...
+
+Note: This option do nothing when ``--debug`` passed.
+
```
Krastanov: Some grammar suggestions:

**Warning:** Silencing messages by templates is a dangerous feature. It is possible to unintentionally filter out multiple message types with the same template
(including messages from future Pelican versions). Proceed with caution.

Krastanov: Note: This option does nothing ``--debug`` is passed.

Krastanov:
@justinmayer , I am not a pelican developer, but if this is passing tests (it is) I would like to see it merged as I would like to use it.

Totktonada:
@Krastanov, thanks! I updated the documentation with your suggestions and rebased the commits at top of current master.

justinmayer:
Thanks to Alexander for the contribution and to everyone who helped review it.  (^_^)

PR #2002, opened 2016-09-14
workaround for python issue28141; copystat fails on Android

jerryasher:
I've been running pelican on Android in python 3.5 and under termux. It was failing whenever it called shutil.copy2 which called shutil.copystat which tries to change the last access and modification time of the destination files. 

You can see that bug documented here:
http://bugs.python.org/issue28141

The changes in this commit wrap the calls to copy2 in an exception handler.

I tested this against pelican  3.6.3 (and I do have pelican working on my Android tablet.)


jerryasher:
Sorry, just realized the change in this pull request to generators.py is wrong. I'm going to close it and try again.

PR #2003, opened 2016-09-14
pelican fails on Android when copying files into the output directory.

jerryasher:
I've been running pelican on Android in python 3.5 and under termux. It was crashing with a permissions error whenever it called shutil.copy2 which called shutil.copystat which tries to change the last access and modification time of the destination files.

You can see that bug documented here:
http://bugs.python.org/issue28141

The changes in this commit wrap the calls to copy2 in an exception handler.

I tested this against pelican 3.6.3 (and I do have pelican working on my Android tablet.)


jerryasher:
I think that makes sense, I'm out of town for a bit, and it will take a few days to make that change.

jerryasher:
I made the change, but I am not where I can test it.

But it makes me wonder as a github noob, what is the proper protocol? Close this pull request and make a new one? Or something different?

iKevinY:
@jerryasher You'll probably want to keep working on the same pull request rather than opening a new one. You can use [`git squash`](https://github.com/getpelican/pelican/wiki/Git-Tips#squashing-commits) in order to clean up the commit history.

jerryasher:
things to learn!

Seriously, I appreciate the help.

jerryasher:
having done a bit of studying, it seems the pull request updates itself, and it has. And thanks for explaining the issue with thebroken builds, I wondered what I had done *this* time.

justinmayer:
Thanks for the contribution, Jerry. I know Git/GitHub can be daunting at first, which is precisely why we've taken such pains to document as much as we can. To that end, if you wouldn't mind, please thoroughly re-read our [Contributing Code](http://docs.getpelican.com/en/stable/contribute.html#contributing-code) section, the first bullet in which mentions the importance of creating a new Git branch specific to your change (as opposed to making your commits in the master branch, which is what you've done in this pull request).

There is other information in that section that is critical to making maintainers' lives sane, so your diligence in this respect is most greatly appreciated.  (^_^)

jerryasher:
Thanks Justin, that's a helpful and enlightening document. I'll do that when I get back to a real computer.

jerryasher:
So I squashed the commit, but it is still in the master branch as I am unsure (that I have no clue) how to put the commit into a different branch and still remain within this pull request.

I'm more than happy to close/delete this pull request and create a new one in a new branch and/or "do the right thing" to place this commit in a new branch if someone can advise me on how to do that.

iKevinY:
Fortunately, GitHub added the ability to [switch the base branch](https://github.com/blog/2224-change-the-base-branch-of-a-pull-request) of a pull request, so you won't need to open a new one. Assuming you're currently on the `master` branch in your fork, you should be able to just do `git branch <new-branch-name>`, push that new branch to GitHub, then switch the branch of this PR.

Also, I've realized that `PermissionError` only exists in Python 3.3+. To support Python 2, you'll need to catch the [more general `OSError`](http://stackoverflow.com/a/18199529) instead.

Scheirle:
> Fortunately, GitHub added the ability to switch the base branch of a pull request, so you won't need to open a new one.

@iKevinY You can change the base (=target) branch but not the source branch. So this doesn't work here. 

I recommend finishing this PR with the master branch, there is nothing wrong with it.
`master` is only a name and has no special properties (in most cases). 
The always create a new branch policy is just there to avoid git pitfalls for new users.
(You can't use your master branch now for additional unrelated changes, because it is already occupied by this PR and you don't want to mix those unrelated changes.)

iKevinY:
> You can change the base (=target) branch but not the source branch. So this doesn't work here.

Whoops, you're totally right — got the two mixed up.

jerryasher:
So this currently stands as follows:

+ I have squashed the commit
+ I have informally tested (ie, did a make html in my pelican blog) this fix on Android using both Python 2.7 and Python 3.5 and when those were successful, running make serve and observing the results
+ ran the unittests on Windows, and they pass
+ ran the unittests on Android and they fail (complete log here: http://pastebin.com/9tsAxe2z)

In particular, there is a specific unittest that fails that shows to me, that my fix of wrapping a try/except works for my needs, but is likely wrong in how it may break pelican semantics that I as a noob don't understand.

The particular failure is here:

```
ERROR: test_watchers (pelican.tests.test_utils.TestUtils)
File "pelican/tests/test_utils.py", line 280, in test_watchers
os.utime(path, (t, t))
OSError: [Errno 1] Operation not permitted: 'pelican/tests/content/article_with_metadata.rst'

```

That's dying when test_watchers calls os.utime() which as I noted earlier is returning errors on Android. 

So I am not sure what the right thing to do is. Advice?

iKevinY:
I suspect that this failing test isn't related to the changes that you made in your pull request. It seems to me that, on Android, the test runner doesn't have the permissions necessary to change the accessed/modified times of a file. Could you try running the tests against Pelican's current `master` to see if you see the same error?

(Incidentally, this is generally why it's good practice to create separate branches for your PRs — it allows you to keep your fork's `master` synced with the upstream repo's `master`, and also gives you the ability to have several distinct pull requests open from a single fork.)

jerryasher:
Yeah, I realized later on these are similar errors based on utime, but different in that this unit test is testing for completely different functionality.

However, the failure using the getpelican's current master is the same.

iKevinY:
> However, the failure using the getpelican's current master is the same.

Thanks for checking that. I'd say don't worry about those failing tests for this PR at least; perhaps in a future PR we could get all tests passing on Android. As far as I know, Travis CI doesn't have an Android environment, so we don't really have a way to automate these regression tests.

pelican/generators.py
```diff
@@ -727,8 +727,13 @@ def generate_output(self, writer):
...
             save_as = os.path.join(self.output_path, sc.save_as)
             mkdir_p(os.path.dirname(save_as))
-            shutil.copy2(source_path, save_as)
             logger.info('Copying %s to %s', sc.source_path, sc.save_as)
+            try:
+                # workaround for Android python copy2 bug ([issue28141])
+                shutil.copy2(source_path, save_as)
+            except OSError as e:
+                logger.warning("A problem occurred copying file %s to %s; %s",
+                             source_path, save_as, e)
```
iKevinY: `flake8` is complaining here because this line is ["under-indented"](https://travis-ci.org/getpelican/pelican/jobs/161134330#L186). IIRC, it wants the indent to line up with the opening paranthesis on the previous line (ie. add 2 spaces to the indentation).

pelican/generators.py
```diff
@@ -727,8 +727,13 @@ def generate_output(self, writer):
             source_path = os.path.join(self.path, sc.source_path)
             save_as = os.path.join(self.output_path, sc.save_as)
             mkdir_p(os.path.dirname(save_as))
-            shutil.copy2(source_path, save_as)
             logger.info('Copying %s to %s', sc.source_path, sc.save_as)
+            try:
+                # workaround for Android python copy2 bug ([issue28141])
```
iKevinY: I think it might be clearer to link to the issue page here, as `[issue28141]` feels a little ambiguous.

jerryasher:
Sigh. I think I broke my version of the repo beyond my ability to fix it tonight when trying to rebase.

The end result, the two files that are changed look good and clean and the repo passes the unittests. But rebasing fails.

I did this to myself having introduced merge conflicts between my android tablet and my laptop, and then trying to rewrite history a bit, and then rewriting history some more. And some more.

If anyone can figure out what I need to do to clean up my "17 commits!" that would be great, otherwise, my natural inclination given my current git abilities is to kill this PR and start over.

iKevinY:
Yeah, Git can be overwhelming at times. You could be perform a `git reset --hard getpelican/master` (substitute `getpelican` for whatever the name of the main repo's remote happens to be), redo the appropriate changes, then force-push that new commit. That would essentially be starting over, but would keep this PR intact.

jerryasher:
Thanks to everyone for the help, but I am declaring git bankruptcy on this pull request.

I learned a lot, but lost in the end...

PR #2004, opened 2016-09-16
Fixes unit test test_period_in_timeperiod_archive (TestArticlesGenerator)

Scheirle:
Python 2.7:
On systems with a none english default locale (e.g. de_DE.utf8) this test failed because the
period is a tuple of year and month name and the month name depends on the
locale.

Python 3:
The test passed. (Without this patch)
Shouldn't it fail with python 3 too?

Refs:
https://github.com/getpelican/pelican/blob/09fb6a1393fc26f22fea21f8e96da058675a2cdb/pelican/tests/test_generators.py#L331

---

On python 3.5.2 test_error_on_warning (pelican.tests.test_testsuite.TestSuiteTest) fails here, but on travis it passes. Do you know what could be the cause for this?

### DIFF
--- a/pelican/tests/test_generators.py
+++ b/pelican/tests/test_generators.py
@@ -297,6 +297,8 @@ class TestArticlesGenerator(unittest.TestCase):
         Test that the context of a generated period_archive is passed
         'period' : a tuple of year, month, day according to the time period
         """
+        old_locale = locale.setlocale(locale.LC_ALL)
+        locale.setlocale(locale.LC_ALL, str('C'))
         settings = get_settings(filenames={})
 
         settings['YEAR_ARCHIVE_SAVE_AS'] = 'posts/{date:%Y}/index.html'
@@ -357,6 +359,7 @@ class TestArticlesGenerator(unittest.TestCase):
                                  generator.get_template("period_archives"),
                                  settings,
                                  blog=True, dates=dates)
+        locale.setlocale(locale.LC_ALL, old_locale)
 
     def test_nonexistent_template(self):
         """Attempt to load a non-existent template"""


avaris:
> On python 3.5.2 test_error_on_warning (pelican.tests.test_testsuite.TestSuiteTest) fails here, but on travis it passes. Do you know what could be the cause for this?

How did you run the test? If you do `python -m unittest discover`, that will fail. I don't remember the specifics but python 3.5 changed the way warnings are handled in unittest. 

You need to do `python -Wd -m unittest discover`.

`nose` somehow manages it by itself, so it doesn't show up in that.

Scheirle:
> You need to do python -Wd -m unittest discover

Thanks, now `test_error_on_warning` passes here too.

justinmayer:
@avaris: Any further comments on this pull request? (Feel free to use the review feature to approve or request changes.)

avaris:
Yeah, sorry, I didn't look at the PR throughly first time. I focused on the second part :).

@Scheirle I'm a little wary about setting locale for the whole class. Can't you just use `LOCALE` setting in the method or just fix locale in that?

Scheirle:
> I'm a little wary about setting locale for the whole class. Can't you just use LOCALE setting in the method or just fix locale in that?

This would probably work.

I just choose to do it for the whole class since many other pelican `TestCase` classes do it this way too, even if not all child tests need it.
And secondly it would avoid future locale dependent bugs.
Also generally it makes sense to make the tests as much platform independent as possible, even if it is not yet required. 

Anyway I'm fine with changing this to a local fix. Just wanted to tell you my reasons.

avaris:
> And secondly it would avoid future locale dependent bugs.

My concern is that, this might *hide* bugs. This class in particular does a lot of the more general tests and `pelican` operates on system locale if not defined (as in this case). This was a misconfigured test, but some other failure in future might be an actual bug.

Scheirle:
> My concern is that, this might hide bugs.

Theoretically you are right and it could hide bugs, but that would require that the tests get actually executed with different locales. AFAIK travis does only test one default locale. Therefor it is 
much more likely that a user finds such a bug and not the tests.

---

Updated the PR to only change the locale locally.

Using `settings['LOCALE'] = ['en_US.UTF-8']` did not work since `pelican.settings.configure_settings` does not get called to set the locale.
Also this approach would not reset the locale.

But setting the locale manually works.

justinmayer:
Thanks to @Scheirle for the fix and to @avaris for the review.

PR #2006, opened 2016-09-16
Fixes an issue in test_testsuite.py that causes errors in python 3.5.

avaris:
Also removes the condition in python 3.3, since it's no longer an issue.


avaris:
Just ignore this. Bad memory issues by me.

PR #2008, opened 2016-09-17
Fixes #1868: Add warning when markdown is not installed but user has markdown files

### LINKED ISSUE #1868: Warn when encountering `.md` file without Markdown installed
Hello,

Currently, pelican will skip <code>.md</code> files if the "markdown" python packge is not installed. This result in a <code>WARNING: No valid files found in content."</code> which is **extremely** confusing and has made me lost 30min of debugging.

A better behavior would be to print a warning because there isn't a situation where a user would add a <code>.md</code> file to his content/ while not wanting it to be parsed.

Kind regards,

ehashman:
I am going a little bit further than the suggested fix in the issue discussion, because I think the suggested hint is misleading in certain contexts: for instance, when the user _does_ have the markdown package installed, but put their content in a place Pelican can't find. IMHO, misleading hints are worse than no hints at all.

This fix has the additional benefit of warning the user about files that are silently not getting processed. For instance, if a user currently has one HTML file in their content and the rest are Markdown, they do not receive a "No valid files found in content" warning, and only a single file gets processed. I personally find this confusing.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -69,10 +69,12 @@ I want to use Markdown, but I got an error.
 
 If you try to generate Markdown content without first installing the Markdown
 library, may see a message that says ``No valid files found in content``.
-Markdown is not a hard dependency for Pelican, so if you have content in
+Markdown is not a required dependency for Pelican, so if you have content in
 Markdown format, you will need to explicitly install the Markdown library.
-You can do so by typing the following command, prepending ``sudo`` if
-permissions require it::
+Pelican will attempt to check this for you, so you should also see an error
+message that says ``It appears you do not have the "markdown" package
+installed``.  You can do so by typing the following command, prepending
+``sudo`` if permissions require it::
 
     pip install markdown
 
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -19,7 +19,7 @@ from pelican import signals
 from pelican.generators import (ArticlesGenerator, PagesGenerator,
                                 SourceFileGenerator, StaticGenerator,
                                 TemplatePagesGenerator)
-from pelican.readers import Readers
+from pelican.readers import Markdown, MarkdownReader, Readers
 from pelican.settings import read_settings
 from pelican.utils import (clean_output_dir, file_watcher,
                            folder_watcher, maybe_pluralize)
@@ -400,6 +400,21 @@ def main():
                                             [''],
                                             pelican.ignore_files),
                     'settings': file_watcher(args.settings)}
+
+        # check to see if user is attempting to process markdown without the
+        # dependency installed
+        markdown_files = folder_watcher(
+            pelican.path,
+            MarkdownReader.file_extensions,
+            pelican.ignore_files
+        )
+
+        if Markdown is False and next(markdown_files) is not None:
+            logger.warning(
+                'It appears you have Markdown content but do not have the '
+                '"markdown" package installed, which is required for '
+                'processing Markdown-based content.'
+            )
 
         old_static = settings.get("STATIC_PATHS", [])
         for static_path in old_static:


ingwinlu:
i don't like the idea of adding a folder watcher just for this use case

ingwinlu:
instead we could enable the MarkdownReader even when Markdown is not available. Then we could modifiy the MarkdownReader.read function to test for it to generate a warning log with the file name and that markdown is not installed.

iKevinY:
I agree with @ingwinlu here; adding a folder-watcher seems like overkill.

ehashman:
I'm not really sure where the idea of "overkill" is coming from. First, we're already walking the filetree, multiple times in the watched case. Furthermore, we're only calling this new generator a single time on startup, so it's not like this adds any significant amount of overhead. I chose to implement this in this manner in order to maximize the reuse of existing logic, instead of rolling my own filetree walk.

If the concern is performance, the difference is imperceptible to an end user:

```
(issue1868)$ touch content/{1..5000}.md  # we now have 5000 content files to stat
(issue1868)$ time make html
pelican ~/src/test-pelican/content -o ~/src/test-pelican/output -s ~/src/test-pelican/pelicanconf.py 
WARNING: It appears you have Markdown content but do not have the "markdown" package installed, which is required for processing Markdown-based content.
Done: Processed 1 article, 0 drafts, 0 pages and 0 hidden pages in 0.19 seconds.

real	0m0.422s
user	0m0.369s
sys	0m0.054s
(issue1868)$ deactivate 
$ . ~/virtualenvs/pypi-pelican/bin/activate
(pypi-pelican)$ time make html
pelican ~/src/test-pelican/content -o ~/src/test-pelican/output -s ~/src/test-pelican/pelicanconf.py 
Done: Processed 1 article, 0 drafts, 0 pages and 0 hidden pages in 0.19 seconds.

real	0m0.403s
user	0m0.352s
sys	0m0.048s
```

That being said, we can minimize some of the overhead for users that already have markdown installed if we check that first, so I went ahead and switched the order of the if statement.

I feel like it is more hacky to enable the Markdown reader when markdown is not available and throw multiple warnings while attempting to read files we already know we can't process. That is both a poorer user experience and seems like a lot of additional overhead.

To be honest, I'm disappointed that a quick and easy fix directly requested by a number of users is being nitpicked over technical details---these are bikeshed arguments amongst developers and ultimately the end user will never see or care about them.

I'll leave it to you as to whether you want to merge, close, or modify this contribution to suit your needs.

iKevinY:
> Furthermore, we're only calling this new generator a single time on startup

Ah, I misinterpreted your changes on my first read-through of the PR, so thank you for the clarification. This seems like a good fix for #1868. It's always possible that someone without `markdown` installed will have Markdown files without intending for them to be read as content files, but I believe those users could simply use `LOG_FILTER` to filter out those messages.

avaris:
In risk of being more nit-picky, here is my problem with the fix:

- You're scanning the whole `content` path without even taking `ARTICLE_{PATHS/EXCLUDES}` and `PAGE_{PATHS/EXCLUDES}` settings into account. Which means that, you'll warn the user that s/he has markdown content, even if the file was not intended as a content. In that case, warning will be very misleading.

-  You're skipping `autoreload` completely. Assume that you started `autoreload` with empty content and then added a `.md` file without having `markdown` installed. User will not see this warning.

iKevinY:
Good points, @avaris. I figured that it would be annoying to see 'Markdown is uninstalled' warnings every time a file change was triggered, but looking closer at the `autoreload` code, this can be handled more gracefully than I initially thought was possible.

ingwinlu:
I never mentioned overkill. I never said that the fix is terrible. I only mentioned that i don't like it.

The suggested solution patches the error case nicely BUT pelican is full of such code snippets that handle a single case. It is only normal that a project that spans over such a variety of contributors evolves that way.

So hence my suggestion: don't just patch in something for one case, instead rework readers to throw nice error messages when they can't read files.

I hope I could bring out my point better now. It was just an suggestion on how I personally think this case can be handled better and just that. So no code or anything that could be compared to the work that has been invested here already.

justinmayer:
Thanks for the enhancement, Elana. Do you have any comments on the feedback here? Would you be willing to make some modifications based on said feedback?

justinmayer:
Hi @ehashman. Just thought I'd follow up and see what you think about the input regarding your pull request. Would you be so kind as to share your thoughts?

justinmayer:
I appreciate the intent here, but since we weren't able to get any traction, I think it's best to close this for now. If someone would like to work on this, by all means please comment here.

PR #2013, opened 2016-09-19
Add AsciiDoc example

marco-m:
Following discussion for issue 488 of pelican-plugins (https://github.com/getpelican/pelican-plugins/issues/488)


Scheirle:
As the AsciiDoc reader is a plugin in my opinion the documentation for it should be with the plugin and not the pelican core. For example in its readme file.

justinmayer:
I must agree with @Scheirle here.

marco-m:
Fair enough. Will open a PR against the plugins repo.

marco-m:
Replaced by https://github.com/getpelican/pelican-plugins/pull/783

PR #2016, opened 2016-09-28
Python's shutil.copy2 fails on Android

jerryasher:
Python's shutil.copy2 fails on Android when copying a file's meta data (perm bits, access times) onto certain filesystems (presumably FAT). This is documented as python [issue28141](https://bugs.python.org/issue28141)

These commits workaround that bug by

+ creating a new function copy_file_metadata in utils.py
+ wrapping calls to copy2 in a try/except clause that logs any errors that occur but keep execution going
+ changing the calls to shutil.copy2 to calls to the new function

### DIFF
--- a/pelican/generators.py
+++ b/pelican/generators.py
@@ -5,7 +5,6 @@ import calendar
 import fnmatch
 import logging
 import os
-import shutil
 from codecs import open
 from collections import defaultdict
 from functools import partial
@@ -21,8 +20,9 @@ from pelican import signals
 from pelican.cache import FileStampDataCacher
 from pelican.contents import Article, Draft, Page, Static, is_valid_content
 from pelican.readers import Readers
-from pelican.utils import (DateFormatter, copy, mkdir_p, posixize_path,
-                           process_translations, python_2_unicode_compatible)
+from pelican.utils import (DateFormatter, copy, copy_file_metadata, mkdir_p,
+                           posixize_path, process_translations,
+                           python_2_unicode_compatible)
 
 
 logger = logging.getLogger(__name__)
@@ -727,8 +727,8 @@ class StaticGenerator(Generator):
             source_path = os.path.join(self.path, sc.source_path)
             save_as = os.path.join(self.output_path, sc.save_as)
             mkdir_p(os.path.dirname(save_as))
-            shutil.copy2(source_path, save_as)
             logger.info('Copying %s to %s', sc.source_path, sc.save_as)
+            copy_file_metadata(source_path, save_as)
 
 
 class SourceFileGenerator(Generator):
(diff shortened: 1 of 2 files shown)


jerryasher:
Thanks, I'll attend to those.

I appreciate the output of flake8, I need to run that manually. I use emacs
and have had to turn elpy-mode off, because something about elpy keeps
rewriting a lot of the whitespace of one of the two files, and I haven't
figured out yet how to turn the rewriting off.

jerryasher:
@avaris I see what you're saying, the new little worker, copy_file_metadata is actually used twice within utils's copy, so maybe it makes more sense to keep that utility then duplicate the try/except.

Hey, can you explain what `def walk_error` does within the `def copy` function? Is that just a vestigial piece of code or does it have some function?

avaris:
Looks like `walk_error` was introduced in 0a48371. I don't know why it's there, I don't see it being used anywhere. Maybe it was part of the overhaul at some point and refactored out, but left with the stray definition. In any case, it should be safe to remove it.

jerryasher:
Hi @justinmayer,  I see the label "awaiting user feedback" was added to this pull request, but I guess I do not know who "the user" is. Are you waiting on me?

If so, are you waiting on me to comment on @avaris' comment?

I did move the function copy_file_metadata up to below the copy function. I did not remove walk_error, though I am happy to do so, as it is extraneous to what this change request requires. 

Tonight I

1. updated my master branch with the latest getpelican master by using `git pull`
1. updated my copy_file_metadata branch by merging my master branch into it.
1. pushing both of my branches onto my repository here.

This places two commits into this pull request. The first are my changes, the next is presumably the recent changes to getpelican/pelican master.

I am not sure if I needed to update my branch at all, or if I did it the right way. What do you think?

justinmayer:
Yes, I was awaiting your response to the latest comments from @Avaris. Now that you have made some follow-up changes, there are two things left to do:

First, as you can see by the indicator at the bottom of this pull request, there is a failing check. In this case, it is compliance with PyCodeStyle, formerly known as PEP 8, that needs to be addressed ([see test output](https://travis-ci.org/getpelican/pelican/jobs/165759469#L187)). You can run `flake8` locally in order to display the failing check(s), and then you can manually fix the displayed problem(s):
```
cd pelican
pip install -r requirements/developer.pip
flake8 pelican/generators.py pelican/utils.py
```
Second, once `flake8` checks pass, you should [rebase & squash your commits](https://github.com/getpelican/pelican/wiki/Git-Tips#squashing-commits) into a single commit.

jerryasher:
Okay, try this one!

justinmayer:
Appreciate for your work on this, Jerry. And thanks to @iKevinY and @avaris for reviewing.

Issue #1963, opened 2016-05-28
save_as and URL can't be arrays, but it's badly reported


JulienPalard:
I had a problem with a file containing two "save_as" due to a bad `sed` (my fault):

```
$ grep -c ^save_as: content/the_problem.md 
2
```

The problem is an exception raised from the `os.path.join` called in [writers.py's write_file](https://github.com/getpelican/pelican/blob/master/pelican/writers.py#L157)

But I don't think catching the exception anywhere can be the right way to do, I'd rather prefer detecting the error upfront, like in the reader, but it look like a checker is already implemented, in `contents.py`, named `check_properties` and it seems the ideal place to drop something like:

```
def check_properties(self):
    """Test mandatory properties are set."""
    string_like_only_properties = {'save_as', 'url'}
    for prop in string_like_only_properties:
        if hasattr(self, 'save_as') and isinstance(self.save_as, list):
            raise ValueError("property save_as should not be a list in {}"
                             " (probably declared twice).".format(self))
    for prop in self.mandatory_properties:
        if not hasattr(self, prop):
            raise NameError(prop)
```

Giving:

> CRITICAL: ValueError: property save_as should not be a list in /home/mdk/pelican_test/content/the_problem.md (probably declared twice).

Instead of:

> CRITICAL: TypeError: join() argument must be str or bytes, not 'list'

But, before writing a PR, I'd like to know if there's other properties that have restrictions, I think of date being a parsable date, what else ? Why not checking if some properties are not empty ? Valid paths maybe ? I noted that starting a URL override with a `/` generate a link in the form `http://0.0.0.0/`, don't know if it's a real bug (I'm new to pelican) but if so, we can check it here too.

leotrs:
+1 on checking missing properties. I think the only two necessary ones are Title and Date.
-1 on checking correct paths, I think pelican reports problematic paths well now.

JulienPalard:
I'm not talking about checking paths, but checking that variables that should contains string (that will use string methods later) are not declared twice (and parsed as lists).

In other words: Forbid (and give verbose error message) lists where strings are intended, typically the "save_as" property, but others as well.

leotrs:
So, perform type/sanity checks for some/all pelican settings? Sounds good to me.

avaris:
[...]

PR #2017, opened 2016-09-30
Explicitly disallow duplications of URL and save_as

JulienPalard:
Closes https://github.com/getpelican/pelican/issues/1963

I preferred an explicit array of disallowed duplicates, far more readable that a `lambda x, y: x` meaning "duplicates are not allowed".


justinmayer:
Would appreciate some help from @getpelican/reviewers. Would you please have a look at this pull request and submit a review?

avaris:
If you're going to define a whole new dict for duplication constraints, don't use `METADATA_PROCESSORS` for that too (i.e. the `or name in METADATA_PROCESSORS` check). You're dividing the information and making it confusing.

Also, since this new dict is just for markdown, it'd be better to have it as a class attr for `MarkdownReader` rather than a module global.

JulienPalard:
@avaris I kept the check on `METADATA_PROCESSORS` for two reasons:

 - For compatibility, if someone is accustomed that an entry in `METADATA_PRCESSORS` can't have duplicates and did not see the new dict
 - From a logical point of view, as it looks like no `METADATA_PROCESSORS` are allowing duplicates entries,  better check them too.We may completly switch the check to the new dict only with the introduction of `METADATA_PROCESSORS` allowing duplicates fields, which may make sense, but I'm not sure

avaris:
The only reason `METADATA_PROCESSORS` was used there is because it looked like it had the relevant information. Apparently it's not true if we have metadata that doesn't need to be processed but at the same time can not have multiple definitions.

If you want to keep the "`METADATA_PROCESSORS` are automatically non-duplicate" relation, then why would you include them in your new dict? Since you're already checking them with that `or` condition. The dict can as well be just:

```python
DUPLICATES_DEFINITIONS_ALLOWED = {
    'url': False,
    'save_as': False
}
```

(P.S.: `url` should be lowercase in your PR too)

Either listing everything in there is pointless, or additional check with `METADATA_PROCESSORS`.

JulienPalard:
Confusion does not emerges from the item being forbidden of being a list twice, but in why it's forbidden in to process a list. So in this point our ides converges, and I remove my test on disallowing lists to be processed.

pelican/readers.py
```diff
@@ -31,6 +31,20 @@
...
+
+DUPLICATES_DEFINITIONS_ALLOWED = {
+    'tags': False,
+    'date': False,
+    'modified': False,
+    'status': False,
+    'category': False,
+    'author': False,
+    'save_as': False,
+    'URL': False,
```
avaris: Just a small issue. `URL` should be lowercase:

```
'url': False
```

justinmayer:
@avaris and any other @getpelican/reviewers: Any other thoughts on the current state of this pull request?

pelican/readers.py
```diff
@@ -264,7 +278,7 @@ def _parse_metadata(self, meta):
                 self._md.reset()
                 formatted = self._md.convert(formatted_values)
                 output[name] = self.process_metadata(name, formatted)
-            elif name in METADATA_PROCESSORS:
+            elif not DUPLICATES_DEFINITIONS_ALLOWED.get(name, True):
```
iKevinY: If all of the values in `DUPLICATES_DEFINITIONS_ALLOWED` are `False`, wouldn't it be cleaner to just refactor that dictionary into a set, and check for set membership here?

justinmayer:
@JulienPalard: Any response (or changes) from you related to the feedback from @iKevinY ?

pelican/readers.py
JulienPalard: I don't think so, explicit is better.

Having an attribute not in the dict would clearly say "we still did not think of it".

Having an attribute set to True would clearly mean "We though of it, and we explicitly allow for duplicates, in the future, do not spend time thinking if it's OK or not".

Yes I imagine this dict growing over time to explicitly tell if something is allowed to appear multiple times or not.

JulienPalard:
@justinmayer replied last week on the review directly.

pelican/readers.py
```diff
@@ -31,6 +31,20 @@
 # This means that _filter_discardable_metadata() must be called on processed
 # metadata dicts before use, to remove the items with the special value.
 _DISCARD = object()
+
+DUPLICATES_DEFINITIONS_ALLOWED = {
```
iKevinY: At the moment, the purpose of `DUPLICATES_DEFINITIONS_ALLOWED` being a dictionary of all `False` values might be a bit puzzling. It would be valuable to add a comment here that explaining that (and perhaps also noting that `True` values might be added in the future), as that would improve future maintainability.

justinmayer:
Many thanks to @JulienPalard for the contribution and to @iKevinY and @avaris for reviewing. 🎊

Issue #1092, opened 2013-09-24
Inherit from arbitrary theme


memeplex:
It would be useful to have the ability to inherit from themes other than simple. Currently there is a workaround to get that, for example for nmnlist:

```
THEME = 'mytheme'
EXTRA_TEMPLATES_PATHS = [ themes_path, themes_path + '/nmnlist/templates' ]
STATIC_PATHS = ['images', themes_path + '/nmnlist/static' ]
```

Then you extend, say, mytheme/templates/base.html from nmnlist/templates/base.html. 

But that's a bit of a hack. It doesn't seem too difficult to add an option that lets the user configure a base theme instead of "simple". In generators.py:

```
        simple_loader = FileSystemLoader(os.path.join(theme_path,
                                         "themes", "simple", "templates"))
```

could be

```
        simple_loader = FileSystemLoader(os.path.join(theme_path,
                  "themes", self.settings.get('BASE_THEME', 'simple'), "templates"))
```

leotrs:
@mitchtbaum This seems like a good addition. What's the state of this?

fuzzy76:
I was just looking for something like this. Being able to make my own base theme that in turn inherited from something like bootstrap3 would be awesome.

digitalrounin:
@justinmayer,

Why not search templates in `EXTRA_TEMPLATES_PATHS` first, before processing the `template/` directory for the current theme?

Keeping the _simple_ theme as the default, baked in safety net just seems like a good idea.  Let all theme developers rely on _simple_ always being there making it a global default.

Processing templates in `EXTRA_TEMPLATES_PATHS` first allows for multiple levels of overriding as well making there system more flexible.  We can configure things like:

```
THEME = 'themes/pelican-bootstrap3'
EXTRA_TEMPLATES_PATHS = (
    'theme_overrides/pelican-bootstrap3/templates'),
    'templates')
STATIC_PATHS = (
    'static',
    'theme_overrides/pelican-bootstrap3/static'))
```

Throwing the theme name (`pelican-bootstrap3`) in a variable, in the configuration file would make swapping in and out of themes really easy.

I am open to other ideas as well.  This just seems like the simplest solution.

digitalrounin:
Let me start by saying that I hear "add Multi-theme support" and I start thinking of theme mixins, which really excites me.

[...]

PR #2020, opened 2016-10-06
Override theme templates. Refs #1092

digitalrounin:
Updated Jinga2 loader so that paths specified in `EXTRA_TEMPLATES_PATHS`
are checked for a template before checking in the theme's `template/`
directory.  This allows a user to override any theme template file.

### DIFF
--- a/pelican/generators.py
+++ b/pelican/generators.py
@@ -52,9 +52,9 @@ class Generator(object):
         # templates cache
         self._templates = {}
         self._templates_path = []
+        self._templates_path += self.settings['EXTRA_TEMPLATES_PATHS']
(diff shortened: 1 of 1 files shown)


ingwinlu:
Can you add a test to verify the changed behaviour?

digitalrounin:
@ingwinlu: I was hoping to get buy in on getpelican/pelican#1092 before investing too much time on implementing tests.  I should have mentioned that earlier.  Don't want to spend too much time on this if another solution is proposed.

But I would be happy to add tests if people are happy with my proposed solution.

digitalrounin:
Closing this out in favor of what I am proposing in feature request getpelican/pelican#2021.

PR #2022, opened 2016-10-09
Prevent server from dying when query strings present

rjames86:
When calling a url that contains a query string, `ComplexHTTPRequestHandler.do_GET` can't find the file since it doesn't exist.

```WARNING:root:Unable to find`/pages/track-my-progress`or variations.```

Removing the query string if present.

### DIFF
--- a/pelican/server.py
+++ b/pelican/server.py
@@ -18,6 +18,10 @@ class ComplexHTTPRequestHandler(srvmod.SimpleHTTPRequestHandler):
     SUFFIXES = ['', '.html', '/index.html']
 
     def do_GET(self):
+        # cut off a query string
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by rjames86** [comment]
@jmayer just wanted to bring this one up since it's a tiny update

**Comment by avaris** [comment]
You're assuming there will be a single `?` in the url, which is not necessarily true ([ref](https://tools.ietf.org/html/rfc3986#section-3.4)). If there is multiple, this will break. You should do this instead:

```python
self.path.split('?', 1)
```

rjames86:
Great catch. Fixing now

justinmayer:
@avaris: Any follow-up comments on this PR?

Issue #1946, opened 2016-04-21
Extension matching in READERS setting


machow:
The docs mention the READERS config may be set to a dictionary to skip processing files. For example, 

```
READERS = {'Rmd': None}
```

In this case, the behavior I expected was that Pelican would not try to process `somefile.Rmd`. However, I'm still getting the errors, "Pelican does not know how to parse %s', u'somefile.Rmd'".

Looking at the [this bit of source](https://github.com/getpelican/pelican/blob/58904af92c845309008989115df6d9de40533b2d/pelican/readers.py#L454-L461) for the Readers class, the problem seems to be that the "None Type" readers from the config will be added to self.reader_classes, but not self.readers. As far as I can tell, reader_classes isn't used beyond that point.

Maybe I missed something, though--because this behavior seems to show up in the unit tests! (I just glanced at it, so could be wrong).

**Edit**
I should also mention that for some reason Pelican is trying to parse .Rmd files in my articles folder. Same with if I change the extension to .RRRRRmd, but not if I change it to .Rme, so I wonder if there is some *md matching going on somewhere.

avaris:
Your second observation is the correct one. [Extensions are matched with `endswith`](https://github.com/getpelican/pelican/blob/master/pelican/generators.py#L115), and they don't include `.`. As long as there is a reader for `md` files (and by default there is), pelican will try to read any `*md` file as well. So, it's a bug.

leotrs:
@avaris would it be enough to do 
```
ext = os.path.splitext(file)[1][1:]
if extensions is False or ext in extensions:
    return True
```
?

If so, I can prepare the PR.

justinmayer:
Addressed via 6cadbdf.

PR #2023, opened 2016-10-13
Fix extension-matching bug in generators.py

iKevinY:
Addresses #1946. I also updated the `_include_path`'s docstring since it slightly threw me off.


justinmayer:
Would any other @getpelican/reviewers care to review this pull request?

justinmayer:
Many thanks to @iKevinY for the fix and to @avaris for the review.

Issue #2031, opened 2016-10-18
{filename} issue when {{ SITEURL }} and {{ STATIC_URL }} are different absolute URLS


vforvalerio87:
I'm trying to build a blog where {{ SITEURL }} and {{ STATIC_URL }} differ (they are actually on two different domains); when I build my blog everything that is linked through {filename} ends up with {{ SITEURL }}{{ STATIC_URL }} as a prefix in its url.

The issue seems to be in pelican/contents.py at line 245:

``` python
origin = '/'.join((siteurl, linked_content.url))
```

I just left "linked_content.url" and it works just fine. This is it now:

``` python
origin = inked_content.url
```

Am I missing something?

PR #2034, opened 2016-10-19
{filename} issue when {{ SITEURL }} and {{ STATIC_URL }} are different absolute URLS

vforvalerio87:
See issue #2031 
this fixes malformed URLs for linked content by {filename}


avaris:
... and breaks every static link if you don't have full urls in `STATIC_URL`. Which is the default case (`{path}`), by the way.

justinmayer:
Rendered moot by #2196

PR #2040, opened 2016-10-29
Fixes Exception name related bugs

avaris:
Fixes two bugs that was introduced by #1743:

 - First was the unnecessary exception name output when skipping
content files if a required metadata wasn't available.
It was
`ERROR: Skipping ./demo.rst: could not find information about 'NameError: date'`
and now should be
`ERROR: Skipping ./demo.rst: could not find information about 'date'`

 - Second was a more serious issue. Improper string formatting in the logger
resulted in implicit decoding and would break for non-ascii error messages.

### DIFF
--- a/pelican/contents.py
+++ b/pelican/contents.py
@@ -473,5 +473,6 @@ def is_valid_content(content, f):
         return True
     except NameError as e:
         logger.error(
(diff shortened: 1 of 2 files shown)


justinmayer:
Many thanks for the fixes, Deniz!

PR #2046, opened 2016-11-04
Add period in line 136 in content.rst

ejdzi:
I added a missing period in documentation, section "File metadata", line 136 in file "content.rst".


justinmayer:
Thanks for the submission. In this case, I think leaving off the period improves readability, which is likely why it was omitted in the first place.

PR #2047, opened 2016-11-05
Fixes #1759 -- gracefully handle RST parse errors.

### LINKED ISSUE #1759: sys.exit called on rst parse errors
The docutils.core.Publisher backend is called in a such a way that it doesn't raise an Exception, but calls sys.exit instead. This makes it harder to use pelican as library.

How to reproduce this?

Using an rst file with invalid syntax, run:

```
from pelican import Pelican
from pelican.settings import read_settings

p = Pelican(read_settings('pelicanconf.py'))
try:
    p.run()
except Exception as e:
   print "Caught exception: ", e

print "Continue after failure"
```

Expected behavior:

```
$ python test.py
....../file.rst:29: (ERROR/3) Unknown directive type "fdsaf".

<details>

Caught exception: SomeException
Continue after failure
```

Actual behavior:

```
$ python test.py
....../file.rst:29: (ERROR/3) Unknown directive type "fdsaf".

<details>

$ (silently exits)
```

This is caused because of  `enable_exit_status=True` which is passed to docutils.core.Publisher.publish since #1224 was merged.

Unfortunately, I don't know enough about the internal structure of Pelican to suggest where the docutils Exception should be changed into a sys.exit, but I'm inclined to say this should be close to the entry point. This allows people who call Pelican as a library to catch an Exception as-is, instead of resorting to the workaround

```
try:
    p.run()
except SystemExit as e:
   print "Caught sys.exit."
```

omiday:
- Revert the change in #1224 where 'exit_status_level' is hard coded inside 
  RstReader  "extra_params" attribute in order to trigger an "exit 1": Docutils 
  *will* issue a SystemExit when this and 'enable_status_exit' are set which 
  has the unwanted effect of breaking user experience by forcing a quit when 
  Pelican is in regenerate mode.
- Do not set 'enable_status_exit' to True by default in 'publish' which is 
  another revert of #1224: Docutils code does check whether the parameter is 
  set in order to decide whether SystemExit should be called or not. We give 
  users more flexibility by allowing the parameter to be set in Pelican config.  
  Also, we are defaulting to False to avoid SystemExit for the reasons 
  explained above.
- When SystemExit is disabled raise an exception that takes advantage of 
  Pelican's "--debug" command line argument to display the traceback.
- For integration with CI tools (as referenced in #1224) users can pass the 
  "--fatal" command line argument that -- when SystemExit is enabled by 
  configuring the 2 parameters discussed above -- will cause an "exit 1".

### DIFF
--- a/pelican/readers.py
+++ b/pelican/readers.py
@@ -210,7 +210,6 @@ class RstReader(BaseReader):
         extra_params = {'initial_header_level': '2',
                         'syntax_highlight': 'short',
                         'input_encoding': 'utf-8',
-                        'exit_status_level': 2,
                         'embed_stylesheet': False}
         user_params = self.settings.get('DOCUTILS_SETTINGS')
         if user_params:
@@ -223,7 +222,25 @@ class RstReader(BaseReader):
         pub.writer.translator_class = PelicanHTMLTranslator
         pub.process_programmatic_settings(None, extra_params, None)
         pub.set_source(source_path=source_path)
-        pub.publish(enable_exit_status=True)
+        status = user_params.get('enable_exit_status', False)
+        try:
+            pub.publish(enable_exit_status=status)
+        except SystemExit:
+            logger.error(
+                '%s "%s" %s',
+                'Error processing',
+                source_path,
+                '\nChanges have been ignored!',
+                exc_info=self.settings.get('DEBUG', False))
+            raise
+        except:
+            logger.error(
+                '%s "%s" %s',
+                'Exception while processing',
+                source_path,
+                '\nUse "--debug" to display a stack trace.',
+                exc_info=self.settings.get('DEBUG', False))
+            raise
         return pub
 
     def read(self, source_path):


omiday:
**Note**: The PR looks for `enable_exit_status` inside `DOCUTILS_SETTINGS`. So we either need to update the docs or add a new config item. The only reason I like it to stay under DOCUTILS_SETTINGS is because `publish()` accepts it as a parameter along with the other settings Pelican is passing as `DOCUTILS_SETTINGS`. Looking forward to feedback from the more experienced ones...

justinmayer:
Would any @getpelican/reviewers care to submit a review on this pull request?

justinmayer:
Following up on my request to @getpelican/reviewers... Anyone care to review this pull request?

justinmayer:
Could really use some help from @getpelican/reviewers regarding this PR. Would someone please take a look? Thank you! 😁

avaris:
@mosra FYI, your change *shouldn't* do the right thing. At least not what your error message suggests. `pelican` will immediately quit on parse error. No further processing, no output... Basically nothing will be generated.

avaris:
BTW, you don't need to specify the command-line *and* config. `--fatal` is enough to get a non-zero exit code with this change. Config is necessary if you, for some odd reason, liked the "terminate **only** on RST errors" option and wanted it back.

No other reader had this. Normal operation is to raise exceptions which are then logged and displayed to the user (or if needed quit with `--fatal`). This is to get RST in line with others.

justinmayer:
@avaris/@mosra: Many thanks for taking a look at this and providing feedback.

@omiday: Would you be so kind as to rebase on current master and resolve the merge conflict?

avaris:
Can you paste the log with `pelican -D --fatal errors`? (I can't test now)

If you're getting an error log but pelican doesn't terminate with non-zero exit code, then there is something wrong with `--fatal`.

mosra:
@avaris Awesome, thank you for digging this out of the docs (I tried myself some time ago but failed). This looks *way* cleaner than everything else. If @omiday won't pick this up, I assume just taking your diff and integrating it would do? Unless you want to do it yourself, I could submit these changes (together with a test) in a week or so as a fresh PR.

avaris:
Sure, if you're willing to add some test cases you can submit a PR :)

mosra:
@justinmayer PR with the suggestion from @avaris opened in #2438, together with a test case. Sorry for the delay, again.

justinmayer:
Addressed via #2438.

PR #2050, opened 2016-11-15
Make Markdown extensions order non-arbitrary

### LINKED ISSUE #2044: Unstable markdown extensions order
I tested Pelican with #1927 merged and found one breaking difference with Pelican 3.6.3.

In Pelican 3.6.3 extensions is list, which passed to Markdown as is (with 'meta' added) and all extensions initialized in order. Now it can be dictionary `extension_configs` or list `extensions`. But in any way it will passed through `list(set(settings['extensions']))`. So extensions initialized in arbitrary order and a new order for each Pelican execution. Looks like such behaviour starts even before, with #1638.

I have extensions that depend on some internal parsers and ones from 'abbr' and 'meta' extensions. I can add my parser before or after some internal one, but now I cannot add my parser after one from 'meta' or 'abbr'.

I suggest to removing repeated extensions w/o changing its order or don’t do it at all (just carefully add keys from 'extension_configs' to end of 'extension'). The second way looks better for me.

@Scheirle, do you have time to fix it? I can make PR if it’s needed. Hope it should be fixed before 3.7 will released.

Scheirle:
Changes:
- Don't use `set` as it is unordered
- Add `markdown.extensions.meta` directly to `extensions` if needed and not to `extension_configs` since we do not set a custom config.

Fixes #2044

@Totktonada please verify that this solves #2044 for you.

### DIFF
--- a/pelican/readers.py
+++ b/pelican/readers.py
@@ -246,16 +246,14 @@ class MarkdownReader(BaseReader):
 
     def __init__(self, *args, **kwargs):
         super(MarkdownReader, self).__init__(*args, **kwargs)
-        # make sure 'extension_configs' exists and
-        # and either way 'markdown.extensions.meta' must be in there
         settings = self.settings['MARKDOWN']
         settings.setdefault('extension_configs', {})
-        settings['extension_configs'].setdefault(
-            'markdown.extensions.meta', {})
         settings.setdefault('extensions', [])
-        settings['extensions'].extend(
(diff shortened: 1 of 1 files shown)


Totktonada:
@Scheirle I don’t observe any problems on my setup with this change applied. Thank you!

justinmayer:
Thanks to @Totktonada for reporting the issue and to @Scheirle for addressing it so quickly! (^_^)

PR #2052, opened 2016-11-20
Fix translation metadata support

Rogdham:
It seems that the translation metadata support is broken since the beginning (see #830 for initial implementation).

The tests I wrote at that time were passing due to chance: in the [following](https://github.com/Rogdham/pelican/blob/eb640b09e6d63c8b6c1e0a07aaabf0c657f922ff/pelican/tests/test_utils.py#L108):
 ```python
 articles = [fr_article1, en_article1, fr_article2, en_article2, fr_article3, en_article3]
 ```
…if the `en` articles came before the `fr` ones, the test would not pass.

So I rewrote the test so that both orders are being tested for.

***

Example to trigger the bug:
 - Set default lang to `en`;
 - Create an article in `fr`;
 - Create a translation of that article in `en`, and use the `translation` metadata to tell that it is a translation.

In that case, the original article is set to the one in `en`, but it should be the one in `fr`.

### DIFF
--- a/pelican/utils.py
+++ b/pelican/utils.py
@@ -627,41 +627,50 @@ def process_translations(content_list, order_by=None):
 
     for slug, items in grouped_by_slugs:
         items = list(items)
-        # items with `translation` metadata will be used as translations...
-        default_lang_items = list(filter(
-            lambda i:
-                i.metadata.get('translation', 'false').lower() == 'false',
-            items))
-        # ...unless all items with that slug are translations
-        if not default_lang_items:
-            default_lang_items = items
 
         # display warnings if several items have the same lang
         for lang, lang_items in groupby(items, attrgetter('lang')):
             lang_items = list(lang_items)
             len_ = len(lang_items)
             if len_ > 1:
-                logger.warning('There are %s variants of "%s" with lang %s',
-                               len_, slug, lang)
+                logger.warning(
+                    'There are %s items with slug "%s" with lang %s',
+                    len_, slug, lang)
                 for x in lang_items:
                     logger.warning('\t%s', x.source_path)
 
-        # find items with default language
-        default_lang_items = list(filter(
-            attrgetter('in_default_lang'),
-            default_lang_items))
+        # items with `translation` metadata will be used as translations...
+        candidate_items = list(filter(
+            lambda i:
+                i.metadata.get('translation', 'false').lower() == 'false',
+            items))
+        # ...unless all items with that slug are translations
+        if not candidate_items:
+            logger.warning('All items with slug "%s" are translations', slug)
+            candidate_items = items
 
-        # if there is no article with default language, take an other one
-        if not default_lang_items:
-            default_lang_items = items[:1]
+        # find items with default language
+        original_items = list(filter(
+            attrgetter('in_default_lang'),
+            candidate_items))
+
+        # if there is no article with default language, go back one step
+        if not original_items:
+            original_items = candidate_items
+
+        # display warning if there are several original items
+        if len(original_items) > 1:
+            logger.warning(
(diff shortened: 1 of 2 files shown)


pelican/utils.py
```diff
@@ -646,22 +638,37 @@ def process_translations(content_list, order_by=None):
...
                     logger.warning('\t%s', x.source_path)
 
+        # items with `translation` metadata will be used as translations...
+        candidate_items = list(filter(
+            lambda i:
+                i.metadata.get('translation', 'false').lower() == 'false',
+            items))
+        # ...unless all items with that slug are translations
+        if not candidate_items:
+            logger.warning('All times of "%s" are translations', slug)
```
avaris: "All **items**"? 
Also, I think it's better to mention that we're displaying `slug`. So something like this maybe:
```python
logger.warning('All items with slug "%s" are translations', slug)
```

pelican/utils.py
```diff
@@ -646,22 +638,37 @@ def process_translations(content_list, order_by=None):
...
-        # if there is no article with default language, take an other one
-        if not default_lang_items:
-            default_lang_items = items[:1]
+        # display warning if there are several base items
+        if len(original_items) > 1:
+            logger.warning('There are %s original (not translated) items "%s"',
+                           len(original_items), slug)
 
         if not slug:
             logger.warning(
```
avaris: Not entirely this PR's context but noticed it while reviewing. 

Isn't this warning a bit misleading and inadequate? Technically, **every** item in this group has empty `slug`, but we're showing only one. Like this, it'd take multiple runs to get warning for all of them to fix.

pelican/utils.py
Rogdham: Updated my commit to reflect that

pelican/utils.py
Rogdham: This happens in one of the folliwng cases:
- for that slug, there are several versions with all have the default lang and all not marked as translations
- for that slug, there are several versions where none have the default lang and all not marked as translations

So we have one warning saying that there are several originals items for that slug. If there are more than 2, I agree that the warning will still be there if they only change one.

I'm not sure why you wrote that the slug is empty?

Rogdham: Also, in that case, the other log message [line 647](https://github.com/getpelican/pelican/pull/2052/files#diff-fb2a57e0cb7b201d7a478128b53c1c5eL647) should have already shown all sources that have that slug, so I think we are good.

avaris: Newline in log messages is perfectly fine since a while back (#1412) I added a decent way to display them :). However it only handles newlines in raw, unformatted messages (first argument). I might look into that to format the final message instead so it'll be possible to use newlines in the arguments as well.

But for now, you can do something like this:

```python
if not slug:
    logger.warning(
        'There are %s items with empty slug' + '\n%s'*len(items),
        len(items),
        *(x.source_path for x in items))
```

Which should output something like:

```
WARNING: There are 2 items with empty slug
  | /tmp/foo/a.md
  | /tmp/foo/b.md
```

Similarly for others.

As for the changes, a separate commit in this PR is fine I think.

Rogdham: Should be in my next commit :)

Now outputs the following:
```
WARNING: There are 2 items with empty slug: 
  | /tmp/foo/b.md
  | /tmp/foo/a.md
WARNING: There are 2 items with slug "" with lang en: 
  | /tmp/foo/b.md
  | /tmp/foo/a.md
WARNING: There are 2 original (not translated) items with slug "": 
  | /tmp/foo/b.md
  | /tmp/foo/a.md
```

justinmayer:
Thanks to @Rogdham for the fix and to @avaris for the review.  (^_^)

PR #1904, opened 2016-11-23
Markdown Summary meadata is parsed for additional metadata

abackstrom:
The Markdown parser forces the "meta" extension for parsing metadata headers. This parser object is reused when formatting header values for any header listed in the `FORMATTED_FIELDS` setting.

Because header values like Summary are parsed for metadata, any metadata-like values are discarded from the final output. Example:

```
Summary: Test: output
```

The header value of `Test: output` will be parsed as metadata, recognized as a header, and discarded from the final formatted Markdown output, resulting in an empty summary.


abackstrom:
@justinmayer Here's an attempt at fixing this. Before processing metadata values, I temporarily remove the "meta" Markdown extension from the processor instance. This is hacky but it's the most minimal of the changes I attempted.

I had back luck creating a dedicated Markdown processor for metadata values (a bunch of tests failed, it seemed to double all the Metatdata values somehow). A less automatic alternative might be a custom "Meta" extension whose functionality can be disabled and enabled as needed.

pelican/readers.py
```diff
@@ -268,7 +268,11 @@ def _parse_metadata(self, meta):
                 formatted_values = "\n".join(value)
                 # reset the markdown instance to clear any state
                 self._md.reset()
+                # process field as markdown, but *don't* extract md metadata
+                preprocessors = self._md.preprocessors
+                self._md.preprocessors.pop('meta', None)
                 formatted = self._md.convert(formatted_values)
+                self._md.preprocessors = preprocessors
```
avaris: This won't do what you intend it to do. I assume you wanted to 'cache' the Markdown preprocessors instance containing `meta` and reverse it afterwards. But you're mutating `self._md.preprocessors`, so the 'cached' `preprocessors` will be the same object regardless (won't contain `meta`).

In any case, I don't think this is needed. A new Markdown instance is created for every read and the one in `self._md` isn't reused after this.

abackstrom: Thanks for the feedback, @avaris. I could change this up a couple different ways:

1. Remove the meta preprocessor from `self._md` after the initial `self._md.convert` call
2. Create an alternate Markdown parser without the meta extension and use that for processing fields (using `copy.deepcopy` on `self.settings['MARKDOWN']`)

@justinmayer, do you have a preference?

avaris: Actually, how you handle it is fine. What I was trying to say that, two lines below don't do anything:

```python
preprocessors = self._md.preprocessors
#...
self._md.preprocessors = preprocessors
```

Just remove them :).

abackstrom:
Pushed an updated version that removes the meta preprocessor at the start of `MarkdownReader._parse_metadata`.

irui4:
Hello,
is there any news on this PR? I am encountering a problem that mentioned here ; I rely on FORMATTED_FIELDS to apply smartypants processing on certain metadata fields, and it's currently breaking any post/page with a formatted field that begins with another word + colon.

irui4:
It seems that python-markdown offers a method on Registry objects to unload an extension (in this case, the `meta` extension), so adding `self._md.preprocessors.deregister("meta")` should be cleaner.
I'm still looking into it, but if I clean things up and make a new PR later on, would it be of interest? It does fix unexpected/unwanted behavior.

Thanks!

justinmayer:
This functionality is now in master. Many thanks to Annika for helping make this happen.

PR #2056, opened 2016-11-30
Update content'rst with ARTICLE_EXCLUDES information

docent-net:
There's lack in documentation for situations when user puts rst/md/html files inside directory defined in STATIC_PATHS. Normally those files will be parsed by ArticlesGenerator and return error since those files are not meant to be parsed. Thus we should expose information about ARTICLE_EXCLUDES in the documentation for those situations.

See [stackoverflow](http://stackoverflow.com/questions/37104625/how-can-i-copy-html-files-without-pelican-parsing-it/40889552#40889552) for reference.

### DIFF
--- a/docs/content.rst
+++ b/docs/content.rst
@@ -315,6 +315,14 @@ If a static file is linked multiple times, the relocating feature of
 After the first link, Pelican will treat ``{attach}`` like ``{filename}``.
 This avoids breaking the already-processed links.
 
+Bare in mind that if you'd like to put Markdown, HTML or reStructuredText files
+into directories configured in ``STATIC_PATHS`` you should also add those
+directories to ``ARTICLE_EXCLUDES`` list. Otherwise Pelican's article
+generator will try to parse those files and return error::
+
+    STATIC_PATHS = ['images', 'presentations']
+    ARTICLE_EXCLUDES = ['presentations']
+
 **Be careful when linking to a file from multiple documents:**
 Since the first link to a file finalizes its location and Pelican does
 not define the order in which documents are processed, using ``{attach}`` on a


avaris:
You need that **if** any of the items in `STATIC_PATHS` is included in `ARTICLE_PATHS` or `PAGE_PATHS` as well. Then you need the relevant paths added to `ARTICLE_EXCLUDES` or `PAGE_EXCLUDES`.

OTOH, default for `ARTICLE_PATHS` is the whole `content` folder (`['']`). So yes, for the default `ARTICLE_PATHS` you're correct, but it's not always the case.

docent-net:
I suggested documentation update as I think that it's the easiest solution. My problem was that I didn't receive any information in the error message, that my files from STATIC_PATHS were parsed because I didn't include them in ARTICLE_EXCLUDES, thus I had to dig it from the source code.

After reading documentation http://docs.getpelican.com/en/latest/content.html I was sure, that STATIC_PATHS is enough and that Pelican will not try to parse anything inside of this dir (that's why it's "static"?).

However in the http://docs.getpelican.com/en/latest/settings.html where ARTICLE_EXCLUDES is mentioned it's still not enough:

> ARTICLE_EXCLUDES
> A list of directories to exclude when looking for articles in addition to PAGE_PATHS.

I'm just lacking here having ARTICLE_EXCLUDES more exposed to a user - and especially it's connection to STATIC_PATHS.

avaris:
> After reading documentation http://docs.getpelican.com/en/latest/content.html I was sure, that STATIC_PATHS is enough and that Pelican will not try to parse anything inside of this dir (that's why it's "static"?).

Not really. If you look at the [mixed content](http://docs.getpelican.com/en/latest/content.html#mixed-content-in-the-same-directory) section, it explicitly says that a path can contain both static and content files and pelican will process content files, then treat the rest as static:

> Starting with Pelican 3.5, static files can safely share a source directory with page source files, without exposing the page sources in the generated site. Any such directory must be added to both `STATIC_PATHS` and `PAGE_PATHS` (or `STATIC_PATHS` and `ARTICLE_PATHS`). Pelican will identify and process the page source files normally, and copy the remaining files as if they lived in a separate directory reserved for static files.

This allows you to put your associated static files along with the content instead of a global static folder.

`ARTICLE_PATHS` and `PAGE_PATHS` are auto-excluded from each other because they are same type of content and should be mutually exclusive. `STATIC_PATHS`, not so. They may or may not be mutually exclusive. It depends on the case.

By the way, I'm not against more documentation :). I'm just trying to clear out the situation so that you can word the additional bits better and more correctly.

justinmayer:
Many thanks for your contribution, Maciej. Since this pull request never quite made it to the level we need it to be for inclusion into Pelican master, for now I think it's best to close it. If you would like to work on this endeavor further, please comment here so we can discuss.

PR #2058, opened 2016-12-10
Fix generator caching

avaris:
With generator level caching, cached files weren't added to the
collections. Only new files were processed. This fixes the issue.


ingwinlu:
is it possible to add a testcase for this problem?

avaris:
It should be and that would be ideal. However I didn't have much time (trying to catch a plane :)), and didn't want to hold up the release. I'll see what I can do, I might be able to push some tests with it tonight.

justinmayer:
Many thanks to @avaris and @ingwinlu for their work on this!

PR #2063, opened 2016-12-13
fix split on NoneType bug

Tmacshamgod:
AttributeError: 'NoneType' object has no attribute 'split'
locale -a, for the items that don't have a '_', show the error message abrove


Vincent-C:
I think the added condition should be
 ```if locale.getlocale()[0] is not None```
instead of:
 ```if locale.getlocale() is not None```

justinmayer:
Thanks to @Tmacshamgod for the fix and to @Vincent-C the review. This should already be fixed in master, via: https://github.com/getpelican/pelican/commit/84920e8fdf754b642b9a137cad38c38f9417bde5. Please feel free to comment here if you think further work is needed.

PR #2067, opened 2016-12-19
Set setup.py open file encoding to UTF-8

farseerfc:
Package building environments usually differ with a working environment
where locale settings are tricky. This change enforce using UTF-8 to
open files in setup.py in order to avoid possible encoding errors when
reading README and changelog files when they contain non-ASCII chars.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 from os import walk
 from os.path import join, relpath
+from io import open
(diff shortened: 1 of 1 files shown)


justinmayer:
Many thanks for the fix, Jiachen. We didn't notice this until after #2082 was merged. Nonetheless, we really appreciate your help! (^_^)

PR #2071, opened 2016-12-23
pagination.html to show all the page numbers instead of current/total

Anivarth:
Changes made to the simple template/pagination.html so that it will show all the page numbers instead of the current page and total number of pages Ex: <-1 2 3 4 5 6 -> instead of <-2/5 ->

This code supports upto 99999 pages. I am new to pelican, any suggestions are welcome. Thank you.

### DIFF
--- a/pelican/themes/simple/templates/pagination.html
+++ b/pelican/themes/simple/templates/pagination.html
@@ -1,11 +1,36 @@
 {% if DEFAULT_PAGINATION %}
 <p class="paginator">
-    {% if articles_page.has_previous() %}
-        <a href="{{ SITEURL }}/{{ articles_previous_page.url }}">&laquo;</a>
-    {% endif %}
-    Page {{ articles_page.number }} / {{ articles_paginator.num_pages }}
-    {% if articles_page.has_next() %}
-        <a href="{{ SITEURL }}/{{ articles_next_page.url }}">&raquo;</a>
-    {% endif %}
+	    {% if articles_page.has_previous() %}
+		    <a href="{{ SITEURL }}/{{ articles_previous_page.url }}">&laquo;</a>
+	    {% endif %}
+	    {% for page in range(articles_paginator.num_pages) -%}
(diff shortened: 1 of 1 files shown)


justinmayer:
Thank you for your contribution, @Anivarth.

Can I get some input from @getpelican/reviewers regarding the perceived value this pull request would provide to the project?

pelican/themes/simple/templates/pagination.html
```diff
@@ -1,11 +1,36 @@
...
+			{% if page+1 == articles_page.number %}
+			   	<a href="#!">{{ page+1 }}</a>
+			{% else %}
+			    {% if page+1 == 1 %}
+			    	<a href="{{ SITEURL }}/{{ url_for_this_page }}.html">{{ page+1 }}</a>
+			    {% else %}
+			    	<a href="{{ SITEURL }}/{{ url_for_this_page }}{{ page+1 }}.html">{{ page+1 }}</a>
+			    {% endif %}
+		    {% endif %}
+	    {%- endfor %}
```
iKevinY, line 29: Regardless of whether we think the level of granularity of pagination should be increased or not, I'm a bit hesitant on this method of constructing the other page URLs. From my initial reading of it, it's trying to reconstruct the URLs of other pages based on the current page (omitting the current page number)? The hard-coded `.html` in the final URLs makes me believe that change assumes some structure to the site that may or may not actually be true.

avaris: I agree. Getting a page (and its `url`) is pretty straightforward:
```jinja2
articles_paginator.page(page_no).url
```

ingwinlu:
Can we close this as there is apparently no interest in following it up?

justinmayer:
@Anivarth: If you would like to re-compose this endeavor as a plugin, I think that would probably be the best approach.

Thanks to @iKevinY, @avaris, and @ingwinlu for reviewing.

Issue #2021, opened 2016-10-08
Add configuration option that allows overriding theme template files


pedrohdz:
It would be useful to allow overriding theme template files without having to fork an existing theme.  There are times where just a single file needs to be replaced, `article.html` as an example, and everything else can stay the same.  Even overriding multiple files would be helpful.

Having to fork an existing theme seems a little excessive and having to do a 3-way merge to keep up with upstream changed to an existing theme increases friction unnecessarily.

Adding a new variable, like `THEME_TEMPLATE_OVERRIDES`, to the configuration file should do the trick.

``` python
THEME_TEMPLATE_OVERRIDES = (
    'templat_overrides/pelican-bootstrap3/templates',
    'templates')
```

`EXTRA_TEMPLATES_PATHS` already does something **similar**, but the issues with it are:
1. It seems to be targeted towards managing `DIRECT_TEMPLATES`.
2. It is processed/scanned **after** the theme templates.  Theme templates take precedence so there is no chance of overriding using `EXTRA_TEMPLATES_PATHS`.
3. Moving `EXTRA_TEMPLATES_PATHS` up might have some un intended consequences.
4. The name `EXTRA_TEMPLATES_PATHS` does not match up with what we are trying to do.

The change in `pelican/generators.py` would look something like:

```
        self._templates_path = []
        self._templates_path += self.settings['THEME_TEMPLATE_OVERRIDES']
        self._templates_path.append(os.path.expanduser(
            os.path.join(self.theme, 'templates')))
        self._templates_path += self.settings['EXTRA_TEMPLATES_PATHS']
```

This would be a quicker solution than what is being discussed in getpelican/pelican#1092.  It is not the same as _Inheriting from an arbitrary theme_, but is a decent stopgap IMHO.

I will try and get a PR in for this later next week.  I'm hoping there is interest.

Cheers...

digitalrounin:
I will gladly change this into a plugin if that is the best place for this type a change.

justinmayer:
Ostensibly addressed via linked PR.

PR #2026, opened 2016-10-16
Add THEME_TEMPLATE_OVERRIDES. Refs #2021

digitalrounin:
Allow for overriding individual templates from the theme by configuring
the Jinja2 `Environment` loader to search for templates in the
`THEME_TEMPLATE_OVERRIDES` path before the theme's `templates/`
directory.

Related to getpelican/pelican#2021

PR #2072, opened 2016-12-25
Add THEME_TEMPLATE_OVERRIDES. Refs #2021

digitalrounin:
Allow for overriding individual templates from the theme by configuring
the Jinja2 `Environment` loader to search for templates in the
`THEME_TEMPLATE_OVERRIDES` path before the theme's `templates/`
directory.

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -706,15 +706,12 @@ Template pages
    tags and category index pages). If the tag and category collections are not
    needed, set ``DIRECT_TEMPLATES = ['index', 'archives']``
 
+   ``DIRECT_TEMPLATES`` are searched for over paths maintained in
+   ``THEME_TEMPLATES_OVERRIDES``.
+
 .. data:: PAGINATED_DIRECT_TEMPLATES = ['index']
 
    Provides the direct templates that should be paginated.
-
-.. data:: EXTRA_TEMPLATES_PATHS = []
-
-   A list of paths you want Jinja2 to search for templates.  Can be used to
-   separate templates from the theme.  Example: projects, resume, profile ...
-   These templates need to use ``DIRECT_TEMPLATES`` setting.
 
 
 Metadata
@@ -1010,6 +1007,21 @@ However, here are the settings that are related to themes.
    theme has other static paths, you can put them here. If files or directories
    with the same names are included in the paths defined in this settings, they
    will be progressively overwritten.
+
+.. data:: THEME_TEMPLATES_OVERRIDES = []
+
+   A list of paths you want Jinja2 to search for templates before searching the
+   theme's ``templates/`` directory.  Allows for overriding individual theme
+   template files without having to fork an existing theme.  Jinja2 searches in
+   the following order: files in ``THEME_TEMPLATES_OVERRIDES`` first, then the 
+   theme's ``templates/``.
+
+   You can also extend templates from the theme using the ``{% extends %}``
+   directive utilizing the ``!theme`` prefix as shown in the following example:
+
+   .. parsed-literal::
+
+      {% extends '!theme/article.html' %}
 
 .. data:: CSS_FILE = 'main.css'
 
--- a/pelican/generators.py
+++ b/pelican/generators.py
@@ -51,20 +51,25 @@ class Generator(object):
 
         # templates cache
         self._templates = {}
-        self._templates_path = []
-        self._templates_path.append(os.path.expanduser(
-            os.path.join(self.theme, 'templates')))
-        self._templates_path += self.settings['EXTRA_TEMPLATES_PATHS']
+        self._templates_path = list(self.settings['THEME_TEMPLATES_OVERRIDES'])
 
-        theme_path = os.path.dirname(os.path.abspath(__file__))
+        theme_templates_path = os.path.expanduser(
+            os.path.join(self.theme, 'templates'))
+        self._templates_path.append(theme_templates_path)
+        theme_loader = FileSystemLoader(theme_templates_path)
 
-        simple_loader = FileSystemLoader(os.path.join(theme_path,
-                                         "themes", "simple", "templates"))
+        simple_theme_path = os.path.dirname(os.path.abspath(__file__))
+        simple_loader = FileSystemLoader(
+            os.path.join(simple_theme_path, "themes", "simple", "templates"))
+
         self.env = Environment(
             loader=ChoiceLoader([
                 FileSystemLoader(self._templates_path),
                 simple_loader,  # implicit inheritance
-                PrefixLoader({'!simple': simple_loader})  # explicit one
+                PrefixLoader({
+                    '!simple': simple_loader,
+                    '!theme': theme_loader
+                })  # explicit ones
             ]),
             **self.settings['JINJA_ENVIRONMENT']
         )
--- a/pelican/settings.py
+++ b/pelican/settings.py
@@ -99,7 +99,7 @@ DEFAULT_CONFIG = {
     'RELATIVE_URLS': False,
     'DEFAULT_LANG': 'en',
     'DIRECT_TEMPLATES': ['index', 'tags', 'categories', 'authors', 'archives'],
-    'EXTRA_TEMPLATES_PATHS': [],
+    'THEME_TEMPLATES_OVERRIDES': [],
     'PAGINATED_DIRECT_TEMPLATES': ['index'],
     'PELICAN_CLASS': 'pelican.Pelican',
     'DEFAULT_DATE_FORMAT': '%a %d %B %Y',
@@ -376,12 +376,26 @@ def configure_settings(settings):
             settings[new_key] = [settings[old_key]]   # also make a list
             del settings[old_key]
 
+    # Deprecated warning of EXTRA_TEMPLATES_PATHS
+    if 'EXTRA_TEMPLATES_PATHS' in settings:
+        logger.warning('EXTRA_TEMPLATES_PATHS is deprecated use '
+                       'THEME_TEMPLATES_OVERRIDES instead.')
+        if ('THEME_TEMPLATES_OVERRIDES' in settings and
+                settings['THEME_TEMPLATES_OVERRIDES']):
+            raise Exception(
+                'Setting both EXTRA_TEMPLATES_PATHS and '
+                'THEME_TEMPLATES_OVERRIDES is not permitted. Please move to '
+                'only setting THEME_TEMPLATES_OVERRIDES.')
+        settings['THEME_TEMPLATES_OVERRIDES'] = \
+            settings['EXTRA_TEMPLATES_PATHS']
+        del settings['EXTRA_TEMPLATES_PATHS']
+
     # Save people from accidentally setting a string rather than a list
     path_keys = (
         'ARTICLE_EXCLUDES',
         'DEFAULT_METADATA',
         'DIRECT_TEMPLATES',
-        'EXTRA_TEMPLATES_PATHS',
+        'THEME_TEMPLATES_OVERRIDES',
         'FILES_TO_COPY',
         'IGNORE_FILES',
         'PAGINATED_DIRECT_TEMPLATES',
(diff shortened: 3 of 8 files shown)


digitalrounin:
Continuation of getpelican/pelican#2026.  I slipped up and closed that PR accidentally.  Ties in with getpelican/pelican-themes#466.

avaris:
Doesn't this make `EXTRA_TEMPLATES_PATHS` obsolete?

Also, IMO, the docs for this should go to the `Themes` subsection, rather than `Template pages`.

digitalrounin:
@avaris,
> Doesn't this make EXTRA_TEMPLATES_PATHS obsolete?

Might deprecate `EXTRA_TEMPLATES_PATHS`.  I did not want to eliminate, or modify `EXTRA_TEMPLATES_PATHS` since it might mess up existing users.  Should we axe `EXTRA_TEMPLATES_PATHS` in this change or give people time to transition?  Maybe there's an argument for keeping both?

> Also, IMO, the docs for this should go to the Themes subsection, rather than Template pages.

Maybe?  I figured since it is one replacing or enhancing `EXTRA_TEMPLATES_PATHS` it should be put in the same section.  The *TEMPLATE* in `THEME_TEMPLATE_OVERRIDES was another reason for putting it in the _Template pages_ section.


I'm amenable on either of these.

avaris:
Deprecating `EXTRA_TEMPLATES_PATHS` here with a warning should be fine. It can be moved to the new `THEME_TEMPLATE_OVERRIDES` to be backwards compatible. Only issue I see is that, `EXTRA_TEMPLATES_PATHS` wasn't overriding the existing templates, whereas this one will. However, since that wasn't a feature before, I don't see any reason for people to have 'overriding' templates in `EXTRA_TEMPLATES_PATHS` since they won't do anything and will be ignored anyway.

The reason I say, doc should go to `Themes` section is the primary use of this setting. It's to modify an existing theme with adding or replacing templates. `Template pages` is more about `TEMPLATE_PAGES` setting which are one-off templates that produce single output. They are more like `Page` contents. `EXTRA_TEMPLATES_PATHS` was put on there, I think, because `DIRECT_TEMPLATES` was there <sup>[*]</sup>. Although, it's also misleading. It says "These templates need to use `DIRECT_TEMPLATES` setting.", which isn't exactly true. You should be able to use any template in that setting like other theme templates (e.g. custom template for a content using `template` metadata).

[*]: Why is `DIRECT_TEMPLATES` there? It's in a grey area. *Not exactly* a theme setting, at the same time not really a `TEMPLATE_PAGES` but similar enough. However, conventionally, it's tied to a theme more than it's tied to `TEMPLATE_PAGES`

digitalrounin:
Noted...  I should be able to get to updating the PR next week...

Thanks!

digitalrounin:
All set @avaris.

almet:
This looks all good to me. I'm in favor of merging it. Any other thoughts?

ssteinerx:
@almet Is the discussion of [child themes]( https://github.com/getpelican/pelican/issues/1092) separate from this, or does this PR obsolete that discussion?  In my mind a simple override, while useful, is not the same as being to inherit from and `extend` a parent theme.

digitalrounin:
@ssteinerx,
I totally agree, template overrides is not the same as being to inherit and `extend` from an existing parent theme.  Though `THEME_TEMPLATE_OVERRIDES` and inheritance are not mutually exclusive either.  Some things to note:

- This works as a stop gap until #1092 gets figured out.  That request has dragged on for 3+ years and some of the original implementation details are being rehashed.
- It fixes what I view as a hobbled `EXTRA_TEMPLATES_PATHS` while maintaining backwards compatibility.  There might have been a reason to implement `EXTRA_TEMPLATES_PATHS` the way it was, but I question it.  (I could be wrong here :-) )
- Moves us away from having to fork themes and have to deal with merging.
- Even if we implement #1092 tomorrow, it still feels like overkill when all that it is needed is replacing a template file or two.
- `THEME_TEMPLATE_OVERRIDES` and inheritance can coexist, when #1092 is implemented.  If anything `THEME_TEMPLATE_OVERRIDES` can leverage inheritance.

`THEME_TEMPLATE_OVERRIDES` was inspired by how [Hugo theme customization](https://gohugo.io/themes/customizing/) works.  I tested out Hugo when I was experimenting with various static site generators.  Hugo came a close second, but _Jinja_ templating swayed me towards _Pelican_.

ssteinerx:
@digitalrounin I agree, and vote to merge this to allow overriding parts of a template completely while we workout/bike shed #1092.

almet:
Lets summon @avaris on this one :-) any feedback ?

docs/settings.rst
```diff
@@ -696,9 +696,11 @@ Template pages
 
 .. data:: EXTRA_TEMPLATES_PATHS = []
 
-   A list of paths you want Jinja2 to search for templates.  Can be used to
-   separate templates from the theme.  Example: projects, resume, profile ...
-   These templates need to use ``DIRECT_TEMPLATES`` setting.
+   ``EXTRA_TEMPLATES_PATHS`` is being depricated, use
```
ingwinlu: is there no sphinx/rst directive to better emphasize deprecation?

pelican/generators.py
```diff
@@ -52,6 +52,7 @@ def __init__(self, context, settings, path, theme, output_path,
         # templates cache
         self._templates = {}
         self._templates_path = []
+        self._templates_path += self.settings['THEME_TEMPLATE_OVERRIDES']
```
ingwinlu:
```
self._templates_path = []
self._templates_path += self.settings['THEME_TEMPLATE_OVERRIDES']
```
is equal to
```
self._templates_path = self.settings['THEME_TEMPLATE_OVERRIDES']
```

pelican/settings.py
```diff
@@ -97,6 +97,7 @@ def load_source(name, path):
     'RELATIVE_URLS': False,
     'DEFAULT_LANG': 'en',
     'DIRECT_TEMPLATES': ['index', 'tags', 'categories', 'authors', 'archives'],
+    'THEME_TEMPLATE_OVERRIDES': [],
```
ingwinlu: we use theme_template**s** internally, might want to change it for consistency

ssteinerx:
My reading is that `EXTRA_TEMPLATES_PATHS` files have to be used by `DIRECT_TEMPLATES` while `THEME_TEMPLATE_OVERRIDES` is intended to allow overriding *any* template with that from another non-theme directory.

I find this really confusing but I also never used `EXTRA_TEMPLATES_PATHS` because I didn't really get the point of it.

I thought the point of this PR was to add a setting to allow overriding templates not used in `DIRECT_TEMPLATES` and, so, is a separate setting.

avaris:
Current doc for `EXTRA_TEMPLATES_PATHS` is a bit misleading. Templates from `EXTRA_TEMPLATES_PATHS` can be treated like any other template. You can use it with `DIRECT_TEMPLATES`, as the docs suggest, but you can also use them for custom templates in Articles or Pages with `template` metadata.

The only difference between `EXTRA_TEMPLATES_PATHS` and `THEME_TEMPLATE_OVERRIDES` (as proposed in this PR) is the position in the lookup chain. `EXTRA_TEMPLATES_PATHS` was after the theme, so any conflicting templates were resolved from the theme (i.e. no way to override). `THEME_TEMPLATE_OVERRIDES` comes before the theme, so it has priority in case of conflict.

ssteinerx:
If I'm reading this right, you can set the theme, then use EXTRA_TEMPLATES_PATHS instead of the  the proposed PR and it's just a documentation issue.

avaris:
No... `EXTRA_TEMPLATES_PATH` can't override theme templates (which is the whole purpose of this PR). Current lookup chain is like this:

 - templates coming from the `THEME`
 - `EXTRA_TEMPLATES_PATHS`
 - `simple` built-in theme

Jinja will look through these in this order and return the first matching template it finds. So, with `EXTRA_TEMPLATES_PATHS` you **can't** provide your alternate `article.html` template. Because the one in theme will be found earlier and used always.

This PR adds the new setting to the top of the chain, so if the template that's being searched is there, it'll be used (instead of the one from the theme).

avaris:
By the way, it might be worth to include the current theme as a `PrefixLoader` (similar to `simple`, say with `!theme` prefix) with this PR. Especially with the small variations. Because with the way the PR is right now, you pretty much need to write the override from scratch (you can't access the theme template because it's being overridden). If the prefixed one is added, you can use the theme version as a base and override relevant blocks:

```jinja
{% extends '!theme/article.html' %}
{% block someblock %}
My override
{% endblock %}
```

ssteinerx:
@avaris So, if you put your parent theme in EXTRA_TEMPLATES_PATHS, and your "child template" as the regular theme, you accomplish the same thing with no code changes?

avaris:
A theme is more than just the templates. You'd need to duplicate the static part (CSS, JS, etc) of the "parent theme" in your "child theme" which is no better than copying the theme, modifying whatever you want and using it as `THEME`.

ssteinerx:
@avaris Yes, I suspected that the other chunks 'wouldn't come along for the ride.'  Since I hadn't looked at `EXTRA_TEMPLATES_PATHS` carefully due to the documented limitation of its use to direct templates I was hoping that this behaviour might have been overlooked, too :wink:.

ssteinerx:
I wonder how traumatic it would be to just add inheritance of the CSS/JS etc. from `EXTRA_TEMPLATES_PATHS` and updating the documentation to reflect the previous undocumented behaviour and the new behaviour at the same time.

My guess is that `EXTRA_TEMPLATES_PATHS` was probably used very little due to its documented restriction and, since the theme will have its own CSS and JS for anyone using themes in the regular way, this shouldn't cause any problems at all.

avaris:
Real theme inheritance is required if the child theme contains its own static content (e.g. customized main CSS). It could still be possible with certain settings config, but would certainly be "hacky". For template-only modifications, this PR should be sufficient.

digitalrounin:
@ingwinlu - Good stuff...  Keep it coming...  I just submitted a new round of changes.  Let me know if I missed, or misinterpreted, anything.

@avaris - Good idea on `!theme`.  I implemented it.  Please check it out.  I'm stumped on how to implement unit tests for it though.  Any ideas?

@ssteinerx - Actually, anyone that wants to add custom pages via `DIRECT_TEMPLATES` would likely be utilizing `EXTRA_TEMPLATES_PATHS`.

If someone whats to go in and add their own static content (images, JS, CSS) `STATIC_PATHS` should be used and then update the corresponding template file, typically through forking a theme, but soon via `THEME_TEMPLATES_OVERRIDES` as well.  There's also `CUSTOM_CSS` depending on the theme you are using.

---

In my mind, _Pelican_ users can wear one of two hats:

1. _Content Creator_ -  Blog writers and whatnot.
2. _Theme Maintainer/Creator_ - Name says it all.

A person can where one, or both.  I view settings such as `THEME_TEMPLATES_OVERRIDES`, `STATIC_PATHS`, etc. as way of allowing _Content Creators_ to modify the look and feel of their site without being forced into becoming _Theme Maintainers_.  Avoiding the whole theme forking thing.

`THEME_TEMPLATES_OVERRIDES` is not meant to be used theme inheritance.

almet:
I love where this is going. Simplicity matters :-)

digitalrounin:
@almet, @ingwinlu, @avaris, @ssteinerx,
Any chance of wrapping this up?  Anything else need changing?

avaris:
@digitalrounin for testing it, you can create an override, create appropriate settings, create a generator and use its [`get_template`](https://github.com/getpelican/pelican/blob/master/pelican/generators.py#L83) for your override, then render it and check.

digitalrounin:
Finally got some free time to work on this!

@avaris,
Tests for `!theme` prefix added.  Added tests for `!simple` and bad prefixes as well.

@almet,
Any chance we can finally merge this?  :-)

justinmayer:
Any last thoughts from @getpelican/reviewers? Your input would be greatly appreciated!

Juanlu001:
Any chance to move this forward?

justinmayer:
@avaris: Any chance you could take another look and see if anything else is required before merging?

justinmayer:
Many thanks to @digitalrounin for the contribution and to everyone who helped review it!

digitalrounin:
Thanks Everyone!

PR #2074, opened 2016-12-26
Add python 3.6 in tox.ini

farseerfc:
Python 3.6 is now released, this change added python3.6 in tox.ini

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = py{27,33,34,35},docs,flake8
+envlist = py{27,33,34,35,36},docs,flake8
 
 [testenv]
 basepython =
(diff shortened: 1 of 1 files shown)


andriisoldatenko:
@farseerfc What about update in all places?
E.g. https://github.com/getpelican/pelican/blob/master/.travis.yml

farseerfc:
Sorry I didn't add travis config. I don't know when travis will support python 3.6

ingwinlu:
Hi, travis already supports it in their own weird way that does not play that nice with our setup. 

See #2094 for a complete update.

farseerfc:
@ingwinlu hi, thanks for telling me!

PR #2075, opened 2016-12-26
Add HTML meta generator tag

liltinkerer:
Added meta generator HTML tag to default themes.
Good for Pelican since numerous sites measure the popularity of web technologies searching for meta generator tag. Such as [BuilWith](https://builtwith.com/) etc.

### DIFF
--- a/pelican/tests/output/basic/a-markdown-powered-article.html
+++ b/pelican/tests/output/basic/a-markdown-powered-article.html
@@ -2,6 +2,7 @@
 <html lang="en">
 <head>
         <meta charset="utf-8" />
+        <meta name="generator" content="Pelican" />
         <title>A markdown powered article</title>
         <link rel="stylesheet" href="/theme/css/main.css" />
         <link href="/feeds/all.atom.xml" type="application/atom+xml" rel="alternate" title="A Pelican Blog Atom Feed" />
--- a/pelican/tests/output/basic/archives.html
+++ b/pelican/tests/output/basic/archives.html
@@ -2,6 +2,7 @@
 <html lang="en">
 <head>
         <meta charset="utf-8" />
+        <meta name="generator" content="Pelican" />
         <title>A Pelican Blog</title>
         <link rel="stylesheet" href="/theme/css/main.css" />
         <link href="/feeds/all.atom.xml" type="application/atom+xml" rel="alternate" title="A Pelican Blog Atom Feed" />
--- a/pelican/tests/output/basic/article-1.html
+++ b/pelican/tests/output/basic/article-1.html
@@ -2,6 +2,7 @@
 <html lang="en">
 <head>
         <meta charset="utf-8" />
+        <meta name="generator" content="Pelican" />
         <title>Article 1</title>
         <link rel="stylesheet" href="/theme/css/main.css" />
         <link href="/feeds/all.atom.xml" type="application/atom+xml" rel="alternate" title="A Pelican Blog Atom Feed" />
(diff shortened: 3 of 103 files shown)


iKevinY:
How commonplace are these generator `<meta>` tags? Personally, they feel reminiscent of technologies like WordPress and friends. That being said, I took a look at Jekyll, thinking that they wouldn't have this tag, but it does exist in their docs, so maybe they're more popular than I thought? I guess regardless of their popularity, it wouldn't necessarily _hurt_ to include them in the default themes.

liltinkerer:
@iKevinY Hugo has it as well, in addition to Wordpress, Jekyll and many others.

iKevinY:
@liltinkerer Ah, I had no idea!

The PR itself looks good; do any other @getpelican/reviewers have opinions on adding this tag?

anthonyfok:
How about adding the version too, so that it is obvious which version of Pelican was used to generate the website?  :-)

jorgesumle:
The example from the HTML 5.1 shows the version of the program too, which I also think is the best thing to do:

> <meta name=generator content="Frontweaver 8.2">

But how do we get the version number? Of course, displaying only the generator name (Pelican) is better than nothing.

repus:
Mailman@list.ibiblio.org

ssteinerx:
@anthonyfok Not sure how relevant for Pelican this is, but one of the things we do for our WordPress sites is to *remove* the tag that indicates the version with which it was created because it makes for an even easier target for known exploits in specific versions of WordPress.  

Most times, when they go prodding for version info in other ways (there are many), the firewall catches them first.  If they know right where to go, that's less likely.

liltinkerer:
@jorgesumle There's currently no easy way to get the version number. If @avaris thinks it's necessary, I can give it a try.

@ssteinerx I don't think it's relevant to static site generator, there are no holes to exploit :)

ssteinerx:
@liltinkerer I know there's not much to exploit but I prefer to let them fumble around trying to get in so they can get caught and blocked at the firewall so they're not consuming server resources ever again.

jorgesumle:
In case you don't know how to get the version number, here is what I do to add the `<meta name="generator" content="Pelican x.x.x">`. First add the following to your `pelicanconf.py` file:

```
import subprocess
PELICAN_VERSION = subprocess.check_output(['pelican', '--version'], universal_newlines=True).rstrip()
```
You can then use `PELICAN_VERSION` in your theme:

```
<meta name="generator" content="Pelican {{ PELICAN_VERSION }}">
```

avaris:
@liltinkerer Eh, what? I wasn't even in this conversation :). Just for the record, I don't really care either way.

FWIW, `theme` will probably matter a lot more than the `pelican` version in terms of holes for the site security.

Also, for getting the version, this is a bit easier:
``` python
from pelican import __version__ as PELICAN_VERSION
```

justinmayer:
How would the version number affect our functional test output? Maybe I haven't had enough coffee yet, but if we add a version number to the `meta` field, won't this result in situations where pull requests have to be updated with new test output simply because a new version of Pelican was released in the interim?

justinmayer:
Thank you for the contribution, @liltinkerer. As I said in my last comment from a couple of years ago, I suspect the addition of a version number and its effect on functional test output might be more trouble than the version number's presence is worth. So for now, I've added the `generator` meta tag via c8bf3ec5a470363b7284e1a18146a5062ec52a6f and 2d590bb8c183ba22af4b7ed3241fd4f80069fa73.

PR #2083, opened 2017-01-04
Add current date when publishing to GitHub Pages

jagmoreira:
The current gh_pages Fabric task publishes the build website with the default "Update documentation" commit message. That is not very informative. This change adds the current date upon publishing so we can better track the build history.


justinmayer:
Thanks for the contribution and apologies for the delay in reviewing it. Pelican's former support for Fabric has been replaced by Invoke. Please feel free to re-submit this pull request, if applicable, based on Invoke's task.py.

PR #2085, opened 2017-01-13
Add new signal: feed_generated

Scheirle:
This signal gets emitted before a feed gets written to disk.
Therefore it allows plugins to do arbitrary changes to the feed.

Use case:
The pelican_comment_system could use this signal to force empty feeds to not update^1 on every site rebuild. Would fix: https://github.com/getpelican/pelican-plugins/issues/780

^1:  The feed update element get set to the current date/time for empty feeds and therefore changes on every rebuild.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,7 +4,7 @@ Release history
 Next release
 ============
 
-- Nothing yet
+* New signal: ``feed_generated``
 
 3.7.1 (2017-01-10)
 ==================
--- a/docs/plugins.rst
+++ b/docs/plugins.rst
@@ -118,6 +118,8 @@ static_generator_init               static_generator               invoked in th
 static_generator_finalized          static_generator               invoked at the end of StaticGenerator.generate_context
 content_object_init                 content_object                 invoked at the end of Content.__init__
 content_written                     path, context                  invoked each time a content file is written.
+feed_generated                      context, feed                  invoked each time a feed gets generated. Can be used to modify a feed
+                                                                   object before it gets written.
 feed_written                        path, context, feed            invoked each time a feed file is written.
 =================================   ============================   ===========================================================================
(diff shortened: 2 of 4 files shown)


justinmayer:
Thanks, @Scheirle. Looks good to me. Would you mind adding a bullet point to the changelog for this?

Any other comments from @getpelican/reviewers?

leotrs:
One thing that I found tricky while learning about signals for my plugins was how/when their arguments are expected to be used. The description in the documentation is short and succinct, but I wonder if there's space somewhere to have a simple example too.

Otherwise, LGTM.

avaris:
I have no problem with a new signal, but given the problem, same can be achieved by overwriting feed with the modified value using the `feed_written` signal. Only downside is the fact that you're writing almost the same thing twice.

Scheirle:
> Would you mind adding a bullet point to the changelog for this?

Done

> One thing that I found tricky while learning about signals for my plugins was how/when their arguments are expected to be used. The description in the documentation is short and succinct, but I wonder if there's space somewhere to have a simple example too.

I agree that if you're not (yet) familiar with the pelican source code, it is not that easy to understand these signals.
Maybe for each signal we should add a link to a (easy/small) plugin so developers could look up how to use them in a real word plugin.

> [...] same can be achieved by overwriting feed with the modified value using the `feed_written` signal. Only downside is the fact that you're writing almost the same thing twice.

My first approach[1] was quite similar, but since the pelican_comment_system adds a additional feed for every article, this would lead to a lot of unnecessary work.

[1] https://github.com/Scheirle/pelican-plugins/commit/52cef39981cf6f90bd245c8ba9b1aea070fc613b

justinmayer:
Many thanks to @Scheirle for the enhancement and to everyone who helped review it. 👍

PR #2086, opened 2017-01-14
Use better headers in tips docs

Sjord:
Instead of "Tip 1", "Tip 2" under "Extra tips", use headers describing
the content.

Also, remove the "hint" pointer as it is not really a hint and the
document consists of tips anyway.


justinmayer:
Thanks to @Sjord for the enhancements and to @ingwinlu and @iKevinY for the review. 😄

PR #2096, opened 2017-01-24
Fix typo; update note and tip formatting

rackonnoiter:
Fixed one small typo. The formatting changes are nitpicks, but I do think they improve the aesthetics.


ingwinlu:
I like the changes to the 'tips' in #2086 better. Maybe you can split up into 2 commits so we can pick?

PR #2117, opened 2017-03-11
Optional items in FILENAME_METADATA expression.

timwienk:
To keep my articles somewhat organised, I like to add the date to the filename of the article and extract it using `FILENAME_METADATA`. However, for pages I don't quite want to do that. I used to use a simple plugin to set the date based on mtime, but since `DEFAULT_DATE = 'fs'` uses mtime now (since December 2016) instead of ctime, this option is great for this.

However, things go wrong when making the date capture group optional in the `FILENAME_METADATA` expression. Previously I monkey patched the `pelican.readers.parse_path_metadata` function, but since the `DEFAULT_DATE = 'fs'` option changed, I figured it's a good idea to try and get this part of the date configuration sorted within Pelican itself as well.

## Case:

files:
- `content/index.md`
- `content/articles/2017-03-11.interesting-article.md`

pelicanconf.py:
```python
FILENAME_METADATA = '(?:(?P<date>\d{4}-\d{2}-\d{2})\.)?(?P<slug>.*)\.md'
```

## Error:

The regex match result for the "date" group is `None` for `content/index.md`. When Pelican attempts to process the date, `None.replace()` throws an `AttributeError: 'NoneType' object has no attribute 'replace'`.

## Solution:

In addition to checking `if k not in metadata` when applying the matched filename metadata, also check `if v is not None`.

Please see the commit in the pull request for the actual change + added tests.

### DIFF
--- a/pelican/readers.py
+++ b/pelican/readers.py
@@ -672,7 +672,7 @@ def parse_path_metadata(source_path, settings=None, process=None):
                     # .items() for py3k compat.
                     for k, v in match.groupdict().items():
                         k = k.lower()  # metadata must be lowercase
-                        if k not in metadata:
+                        if v is not None and k not in metadata:
                             if process:
                                 v = process(k, v)
                             metadata[k] = v
--- a/pelican/tests/test_readers.py
+++ b/pelican/tests/test_readers.py
@@ -166,6 +166,25 @@ class RstReaderTest(ReaderTest):
         }
         self.assertDictHasSubset(page.metadata, expected)
 
+    def test_article_with_optional_filename_metadata(self):
+        page = self.read_file(
+            path='2012-11-29_rst_w_filename_meta#foo-bar.rst',
+            FILENAME_METADATA='(?P<date>\d{4}-\d{2}-\d{2})?')
+        expected = {
+            'date': SafeDatetime(2012, 11, 29),
+            'reader': 'rst',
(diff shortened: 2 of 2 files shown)


pelican/readers.py
```diff
@@ -672,7 +672,7 @@ def parse_path_metadata(source_path, settings=None, process=None):
                     # .items() for py3k compat.
                     for k, v in match.groupdict().items():
                         k = k.lower()  # metadata must be lowercase
-                        if k not in metadata:
+                        if k not in metadata and v is not None:
```
avaris: Not that big of a deal but I'd prefer switching the comparison order for performance reasons. Comparison to `None` would be faster and should short-circuit earlier:

```python
if v is not None and k not in metadata:
```

timwienk: I agree, that makes sense. I'll swap the order and replace this commit.

justinmayer:
Many thanks to @avaris and @ingwinlu for reviewing this pull request.

justinmayer:
You are my new hero, Tim. Why, you ask?

* clear description of your intended use case, perceived problem, and proposed solution
* rebased & squashed into a single commit — without having to be asked
* a single line change accompanied by three dozen lines of tests (!)

Thank you for fulfilling every project maintainer's dream. 😃 👍

PR #2118, opened 2017-03-13
Use awscli instead of s3cmd for S3 in Makefile

### LINKED ISSUE #1626: Use AWSCLI instead of s3cmd for publishing to AWS S3 bucket
Current `make` command (see [tools/templates/Makefile.in](https://github.com/getpelican/pelican/blob/3b22dd4f2df285f7511ae2439808cd3964ec6ca7/pelican/tools/templates/Makefile.in#L109)) file is for publishing to AWS S3 bucket using command `s3cmd`

Nowadays, there is much more widely used tool called AWSCLI which is able doing the same job very well.

My current `Makefile` is modified as follows:

```
s3_upload: publish
    #s3cmd sync $(OUTPUTDIR)/ s3://$(S3_BUCKET) --acl-public --delete-removed --guess-    mime-type
    aws s3 --profile ttr sync $(OUTPUTDIR)/ s3://$(S3_BUCKET) --acl public-read --delete
```

fxfitz:
Today I ran into an issue where MIME types weren't being set correctly when using `make s3_upload`, causing my website not to render because the stylesheets were being sent as html/plain.

After some research, it looks like this was already covered in #1626.

This implements the solution as described in #1626, switching over to use awscli. 

I've been using it now with no problems (everything uploads as expected, MIME types are set correctly), and can agree with the comments in #1626 stating that awscli is the recommended way of moving forward.

Fixes #1626

### DIFF
--- a/pelican/tools/templates/Makefile.in
+++ b/pelican/tools/templates/Makefile.in
@@ -111,7 +111,7 @@ ftp_upload: publish
 	lftp ftp://$$(FTP_USER)@$$(FTP_HOST) -e "mirror -R $$(OUTPUTDIR) $$(FTP_TARGET_DIR) ; quit"
 
 s3_upload: publish
-	s3cmd sync $(OUTPUTDIR)/ s3://$(S3_BUCKET) --acl-public --delete-removed --guess-mime-type --no-mime-magic --no-preserve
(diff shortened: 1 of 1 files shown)


justinmayer:
Many thanks for the enhancement, Fran!

Issue #2066, opened 2016-12-17
Weird make problem


richyvk:
Hi

I've been playing around with Pelican and come across what I think is an issue. I've created a repo on github and cloned it. Everything works fine, except it seems when I clone the repo into my Google Drive folder. Then make commands seem to blow up, e.g.

`make html` gives: 

`pelican: error: unrecognized arguments: Drive/projects/github-pages/content Drive/projects/github-pages/output Drive/projects/github-pages/pelicanconf.py
make: *** [html] Error 2`

Doing the exact same thing in my Documents folder gives no problems.

I'm guessing it's something to do with my Google Drive folder having a space in its name?

I'm running mac ox 10.11.6, pelican 3.7.0 installed via pip

Not sure if this is something you can help with or if it's even a bug??

jkpr:
You are exactly correct, source code is here: 
https://github.com/getpelican/pelican/blob/master/pelican/tools/templates/Makefile.in#L64

Escape the space, as in `\ `, or put quotes around the directory path. It might work to have `"$$(PELICAN)" "$$(INPUTDIR)" -o "$$(OUTPUTDIR)" -s "$$(CONFFILE)" "$$(PELICANOPTS)"` to replace the line linked to above.

papajohn:
The suggested fix above didn't work for me, FYI.

richyvk:
I didn't ry the fix, just renamed my Google Drive folder to Google_Drive :)

aldoridhoni:
I have the problem with space in directory name too. I just add a quote or double quote at the usage of directory in file Makefile and develop_server.sh. Can someone make a pull request, IMHO this has to be fix in master.

justinmayer:
Anyone care to submit a pull request to address this issue?

justinmayer:
The file in question no longer exists, having been replaced with a Jinja2 template.

PR #2124, opened 2017-03-25
Quote folder path. Refs #2066.

aldoridhoni:
Try to fix issue #2066

Quote folder path for fixing problems when there is space in the path.
Please check did I miss something. Thank you.

### DIFF
--- a/pelican/tools/templates/Makefile.in
+++ b/pelican/tools/templates/Makefile.in
@@ -61,63 +61,63 @@ help:
 	@echo '                                                                          '
 
 html:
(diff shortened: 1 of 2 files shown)


justinmayer:
@aldoridhoni: Thanks for your contribution, and apologies for taking so long to review it. Looks like tests are failing. Would you please rebase on current master, ensure tests pass locally, squash any fix-up commits, and push to your branch so we can review?

justinmayer:
The file in question no longer exists, having been replaced with a Jinja2 template.

Issue #2111, opened 2017-03-03
Can't import local module to pelicanconf.py


georgiastuart:
I have a file with my custom jinja2 filters. These work fine when pasted directly into `pelicanconf.py`. However, when I attempt to import the module using

```python
import jinja_custom_filters as jcf
```

I get the following error:

```
ModuleNotFoundError: No module named 'jinja_custom_filters'
```

I wrote a quick test file in the same directory as `pelicanconf.py` and `jinja_custom_filters.py`, and it has no problem importing `jinja_custom_filters`. 

I've combed the documentation, but I can't find anything that would assist with this.

ingwinlu:
is jinja_custom_filters.py in your pythonpath?
whats the general layout of your setup, i.e. where is pelicanconf.py and where is your filter module?

cropse:
pelican seen the package as plugin, so `PLUGINS = ['jinja_custom_filters']` would work as well.
[see the Docs](http://docs.getpelican.com/en/stable/plugins.html)

cropse:
It work for me
```python
import sys
sys.path.append('.')
import jinja_custom_filters
JINJA_FILTERS = {'filter_name_in_html': jinja_custom_filters.the_filter_function}
```
and It work fine for me, maybe need to update the Docs?
[Answer From](https://linkpeek.com/blog/how-to-add-a-custom-jinja-filter-to-pelican.html)

drummonds:
Works for me as well.  I have the problem on Windows with vritual env.  Is this something to do with pelican running in a different home directory.

eg a pelicanconf.py file like this:

    import sys
    sys.path.append('.')
    from base import *

justinmayer:
Closing due to inactivity. Please feel free to comment here if you feel the issue should be re-opened.

deutrino:
Well, it's still broken, so maybe it should be re-opened on that basis.

bedlamzd:
this workaround works, but feels more like unnecessary step and strange behavior
in my opinion import should work directly, without adding '.' to path
so the issue is that pelican don't import modules located in the same directory as pelicanconf.py
it would be nice if it could do so

Kristinita:
I think it would be nice to reopen this issue.

### 1. Problem

The part of my `pelicanconf.py` before a refactoring:

```python
from sys import path

path.extend((".", ".."))

# pylint: disable=wrong-import-position
# fmt: off

[...]

PR #2126, opened 2017-03-27
Enhance JINJA_FILTERS setting docs

cropse:
Might be clarified issue #2111, improve example.


justinmayer:
Many thanks for the contribution, @cropse.

Issue #2098, opened 2017-01-27
Reason that Draft articles are read twice?


brownnrl:
Hello!  Thank you very much for all the work on Pelican!  It's a great tool.

I'm probably doing something that I shouldn't, which is holding some stateful information in RST directives.  I like to know how many times each directive is hit when a page is generated for generally nefarious purposes.

Everything works wonderfully for regular articles, but drafts gave me a bit of trouble because it turns out they are processed twice.  I traced the behavior to the method ```generate_context``` in ```generators```, which has the lines

``` python
                # TODO needs overhaul, maybe nomad for read_file
                # solution, unified behaviour
                try:
                    article_or_draft = self.readers.read_file(
                        base_path=self.path, path=f, content_class=Article,
                        context=self.context,
                        preread_signal=signals.article_generator_preread,
                        preread_sender=self,
                        context_signal=signals.article_generator_context,
                        context_sender=self)
                except Exception as e:
                    logger.error(
                        'Could not process %s\n%s', f, e,
                        exc_info=self.settings.get('DEBUG', False))
                    self._add_failed_source_path(f)
                    continue

                if not is_valid_content(article_or_draft, f):
                    self._add_failed_source_path(f)
                    continue

                if article_or_draft.status.lower() == "published":
                    pass
                elif article_or_draft.status.lower() == "draft":
                    article_or_draft = self.readers.read_file(
                        base_path=self.path, path=f, content_class=Draft,
                        context=self.context,
                        preread_signal=signals.article_generator_preread,
                        preread_sender=self,
                        context_signal=signals.article_generator_context,
                        context_sender=self)
```

So the article is read, with an exception handling mechanism which passes on it anyway and then if it is in a published state, we continue.  However, if it's not in a published state, we end up reading the same article again.

Before I rework my plugin to account for this double read, I was just curious if this was the desired behavior?  It seems at first glance that the elif conditional block is redundant?

brownnrl:
Oh, I think I see now that I write the issue... the ```content_class``` is ```Article``` to start, but you can't get the status until after you read it.  So you have to read it twice, once to tell that it is a draft, and then a second time to pass the correct ```content_class``` parameter.

avaris:
Correct. And, no, it's not the *desired* behavior, but it's the behavior we have now. It wasn't worth to introduce somewhat ugly code to optimize it since people don't tend to have many drafts. Having said that, it's pretty likely that it'll be *fixed* in a later release.

[...]

Issue #1748, opened 2015-06-06
Rework Content


ingwinlu:
Current Content Structure and population has several problems
- ~~[`DEFAULT_STATUS` gets applied to both, pages and articles, making it effectively useless (every setting except published breaks either pages or articles if you don't want to set everything explicitly)~~
- ~~Drafts are 'Articles' that get discarded because of metadata, then get reread as drafts, resulting in unneeded procession and producing issues like #1356~~
resolved via #2128 

Proposal:
- ~~split up content specific `__init__` functions into the actual classes they belong to~~
- ~~remove DEFAULT_STATUS and replace with DEFAULT_PAGE_STATUS and DEFAULT_ARTICLE_STATUS or remove alltogether~~
started in #2128 , maybe able to refactor further

New structure should have the following steamlined flow:
1. generator throws path and class at readers.read_file
2. read_file calls appropriate reader on the path with reader.read
3. reader.read performs the content parse, then parses metadata according to the class (either page OR ArticleDraft) and returns None if one of the checks failed, or the appropiate content class
4. generator now only has to check if content is none and then file the content appropiatly without needing additional checks

Help with any of those points are welcome

justinmayer:
@ingwinlu: Considering this issue to be addressed. If you feel that's not the case, please comment here and we can re-open to discuss further.

PR #2128, opened 2017-03-28
Consolidate validation of content

ingwinlu:
Previously we validated content outside of the content class via
calls to `is_valid_content` and some additional checks in page /
article generators (valid status).
This commit moves those checks all into content.valid() resulting
in a cleaner code structure.
This allows us to restructure how generators interact with content,
removing several old bugs in pelican (#1748, #1356, #2098).

- move verification function into content class
- move generator verifying content to contents class
- remove unused quote class
- remove draft class (no more rereading drafts)
- move auto draft status setter into Article.__init__
- add now parsing draft to basic test output
- remove problematic DEFAULT_STATUS setting
- add setter/getter for content.status
  removes need for lower() calls when verifying status

### DIFF
--- a/pelican/contents.py
+++ b/pelican/contents.py
@@ -149,7 +149,7 @@ class Content(object):
     def __str__(self):
         return self.source_path or repr(self)
 
-    def _valid_mandatory_properties(self):
+    def _has_valid_mandatory_properties(self):
         """Test mandatory properties are set."""
         for prop in self.mandatory_properties:
             if not hasattr(self, prop):
@@ -159,7 +159,7 @@ class Content(object):
                 return False
         return True
 
-    def _valid_save_as(self):
+    def _has_valid_save_as(self):
         """Return true if save_as doesn't write outside output path, false
         otherwise."""
         try:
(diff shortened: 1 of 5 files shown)


ingwinlu:
@getpelican/reviewers rdy

pelican/contents.py
```diff
@@ -174,10 +171,35 @@ def valid_save_as(self):
...
+                    "Unknown status '%s' for file %s, skipping it.",
+                    self.status,
+                    self
+                )
+                return False
+
+        # if undefined we allow all
+        return True
+
+    def valid(self):
```
iKevinY: I would lean towards calling this method `is_valid` (so we get `article.is_valid()`).

pelican/contents.py
```diff
@@ -156,13 +149,17 @@ def __init__(self, content, metadata=None, settings=None,
     def __str__(self):
         return self.source_path or repr(self)
 
-    def check_properties(self):
+    def _valid_mandatory_properties(self):
```
iKevinY: Maybe `_has_mandatory_properties`, since we're checking for the existence of these properties.

pelican/contents.py
```diff
@@ -383,21 +416,36 @@ def relative_dir(self):
 
 class Page(Content):
     mandatory_properties = ('title',)
+    allowed_status = ('published', 'hidden')
```
iKevinY: Renaming this to `allowed_statuses` makes it clearer that each page can have multiple valid statuses (rather than just one).

ingwinlu:
implemented @iKevinY 's naming feedback and rebased on current master

justinmayer:
@ingwinlu: Do you want to squash these commits or leave them as separate ones?

ingwinlu:
squash before merge. kept them seperate for easier review of the additional
changes.

justinmayer:
Many thanks to @ingwinlu for the enhancements and to @iKevinY for the review!  (^_^)

PR #2129, opened 2017-03-29
Import readline in pelican_quickstart.py if available

jorgesumle:
I was confused when I tried to paste using <kbd>Control</kbd>+<kbd>Y</kbd> in my Bash shell because it printed C^Y instead of pasting the text I deleted previously with <kbd>Control</kbd>+<kbd>W</kbd>. With readline, users won't have this weird experience.

### DIFF
--- a/pelican/tools/pelican_quickstart.py
+++ b/pelican/tools/pelican_quickstart.py
@@ -12,7 +12,7 @@ import sys
 import pytz
 
 try:
-    import readline
+    import readline  #NOQA
 except ImportError:
     pass
 
--- a/pelican/tools/pelican_quickstart.py
+++ b/pelican/tools/pelican_quickstart.py
@@ -12,7 +12,7 @@ import sys
 import pytz
 
 try:
-    import readline  #NOQA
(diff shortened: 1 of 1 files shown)


avaris:
You need `# NOQA` to make flake8 happy.

```python
import readline   # NOQA
```

pelican/tools/pelican_quickstart.py
```diff
@@ -12,6 +12,11 @@
 import pytz
 
 try:
+    import readline  #NOQA
```
iKevinY: Adding a space between `#` and `NOQA` will fix `flake8`'s complaint about the unused import.

jorgesumle:
Whoops, too many silly mistakes, sorry about that. Now it passes the tests. Can you squash the commit and leave only the message of the first commit?

jorgesumle:
Could someone please review it?

justinmayer:
Many thanks to @jorgesumle for the contribution and to @iKevinY and @avaris for reviewing!

PR #2130, opened 2017-03-29
Fix more Python 3.6 regex warnings

ingwinlu:


justinmayer:
Many thanks to Winlu for the fix and to Deniz for the review.  (^_^)

PR #2134, opened 2017-04-16
Fix for the check: if the output directory is a part of the content path

vorakl:
Hi,

Erasing of the output directory doesn't work in a case when it is a part of the content directory but it is not a subdirectory which is possible. 

For example,

DELETE_OUTPUT_DIRECTORY = True
PATH = '/path/to/repo/docs.src/content'
OUTPUT_PATH = '/path/to/repo/docs'

This happens because currently is being checked only if PATH `startswith` OUTPUT_PATH. Yes, it starts but it is not a subdirectory and they are absolutely different.

This pull-request proposes a proper check for this situation.

P.S.
 I removed os.path.realpath() from the check because all paths have already been saved as absolute at that moment

### DIFF
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -163,7 +163,7 @@ class Pelican(object):
         # erase the directory if it is not the source and if that's
         # explicitly asked
         if (self.delete_outputdir and not
-                os.path.commonpath((self.path, 
+                os.path.commonpath((self.path,
                                     self.output_path)) != self.output_path):
             clean_output_dir(self.output_path, self.output_retention)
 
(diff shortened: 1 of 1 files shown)


justinmayer:
Perhaps you might consider running `flake8` locally first on the file(s) you changed? That way you'll know what needs changing before making and pushing your commit.

vorakl:
Yes, true. Sorry about that. 
I've never used/heard flake8 and didn't expect that one line patch could lead to so many formatting issues...

vorakl:
Too many mistakes for the simple thing.

justinmayer:
In the future, no need to create a new PR. Better to re-use the existing one. Thanks!

vorakl:
OK. Got it.
I'm afraid to do something wrong again... ;) What would you recommend to do now with all these PRs?

PR #2137, opened 2017-04-16
Fix for the erasing of an output directory

vorakl:
Hi,

It's another try...
This patch is about checking if a directory from 'output_path' is a part of a directory from 'path'. It basically, gets the most common part of both directories and compares if it equals to a directory specified for erasing. The erasing is possible only if it's allowed and an output directory is not a part of a content's path.

This is an example where the current check doesn't work:

PATH = '/path/to/repo/docs.src/content'
OUTPUT_PATH = '/path/to/repo/docs'

### DIFF
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -162,8 +162,16 @@ class Pelican(object):
 
         # erase the directory if it is not the source and if that's
         # explicitly asked
-        if (self.delete_outputdir and not
-                os.path.realpath(self.path).startswith(self.output_path)):
+        if (self.delete_outputdir and
+                # Compares two dirs similar to os.path.commonpath() from Py3.5+
+                # All paths at this point are already full and normalized.
+                # So, we can omit unnecessary checks
+                os.path.join(os.path.sep,
+                             *(_subdir1 for _subdir1, _subdir2 in
+                               zip(self.path.split(os.path.sep),
+                                   self.output_path.split(os.path.sep))
+                               if _subdir1 == _subdir2
+                               )) != self.output_path):
             clean_output_dir(self.output_path, self.output_retention)
 
         for p in generators:


vorakl:
Hi Kevin, 
You're right, it definitely has to be done in other way.
Thanks for the link, I'll try to come up with a better solution.

vorakl:
Hi Kevin,

As you suggested, I've replaced os.path.commonpath by an equivalent one-liner which works similar in that situation. I checked it in Py27. I also squashed all intermediate commits.

How do you think, what else should be done to get it merged? 

Thanks!

justinmayer:
Thanks to @vorakl for the submission and to @iKevinY for reviewing. A few years later, I have some comments on this PR...

* This would have benefited from a clearer description of the problem that this PR aims to fix. It is not at all clear, for example, that the perceived problem is _not_ that a directory was deleted when it should not have been, but instead that a directory that _should_ have been deleted was not, in fact, deleted as expected.

* The problem with the current code is that it compared directory paths as strings instead of comparing actual paths, but that information is buried in the commit message and not mentioned at all here in the pull request description.

* `os.path.commonpath` can now be used since we no longer support Python 2.7. I don't know what that original implementation looked like since it was rebased away, but I think the basic logic is to ensure that the to-be-deleted `OUTPUT_PATH` is not a parent directory of the `PATH` source content.

I submitted PR #2723 as a would-be substitute for this endeavor.

vorakl:
Thanks, @justinmayer . I'm glad that after 3 years of silence you've finally moved this bug to the logical end. You are right, my solution didn't use os.path.commonpath() but this is due to the support of Python 2.7, which was the case 3 years ago. I think, the final implementation is not so important as the need to fix an absolutely wrong original logic:

```python
if (self.delete_outputdir and not
                os.path.realpath(self.path).startswith(self.output_path)):
```

Thanks for fixing this bug!

PR #2147, opened 2017-04-29
Enhance GitHub User Pages docs in tips.rst

outlace:
I found that one of the easiest ways to publish to GitHub User Pages is to make Pelican a subdirectory within the ``<username>.github.io`` project, then generate the output files in the root level of ``<username>.github.io`` and push the entire project to GitHub.


justinmayer:
Thank you for the contribution, Brandon!

PR #2163, opened 2017-06-06
Ability to override docutils HTML translator

mosra:
The `RstReader` class can now use user-specified writer/translator classes instead of the hardcoded ones from `docutils`. This allows for far easier overriding of the default HTML output -- in the past one would need to copypaste and patch the internal `_parse_metadata()` and `_get_publisher()` functions and risk the code getting out of sync in the future (as done for example in the [bootstrap-rst](https://github.com/aqw/bootstrap-rst/blob/1366f7c8f72814e37ec565f3aca02215b8338b3b/bootstrap.py#L289) plugin). With hypothetical `Html5Writer` and `Html5FieldBodyTranslator` classes, based for example on `docutils.writers.html5_polyglot.Writer` and `docutils.writers.html5_polyglot.HTMLTranslator`, a plugin that overrides the default behavior would now look just like this:

```py
# (definition of Html5Writer and Html5FieldBodyTranslator classes omitted)

class Html5RstReader(RstReader):
    writer_class = Html5Writer
    field_body_translator_class = Html5FieldBodyTranslator

def add_reader(readers):
    readers.reader_classes['rst'] = Html5RstReader

def register():
    pelican.signals.readers_init.connect(add_reader)
```

I'm a Python noob and I'm also pretty new to Pelican, so please point out all the mistakes I did. Will be happy to update the code based on your reviews :)

And by the way, thanks a lot for this great and very flexible site generator. I really enjoy working with it.

### DIFF
--- a/pelican/readers.py
+++ b/pelican/readers.py
@@ -9,7 +9,7 @@ from collections import OrderedDict
 import docutils
 import docutils.core
 import docutils.io
-from docutils.writers.html4css1 import HTMLTranslator
+from docutils.writers.html4css1 import HTMLTranslator, Writer
 
 import six
 from six.moves.html_parser import HTMLParser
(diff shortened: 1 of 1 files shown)


mosra:
Ahh. Didn't know there is a difference.

So the proper way to override them would be just this, right?

```python
class Html5RstReader(RstReader):
    translator_class = Html5Translator
    field_body_translator_class = Html5FieldBodyTranslator
```

mosra:
Any ETA on when this and #2164 could be merged? If there's anything more I can (should) do, please let me know.

Thank you.

mosra:
I needed to override the writer class as well in order to inject a `Transform` class, so I updated the PR.

justinmayer:
Many thanks to @mosra for the enhancement and to @avaris for the review.  (^_^)

mosra:
Thank you for merging!

PR #2158, opened 2017-06-07
Wrong path for attached static content urls

diazona:
When I `{attach}` a file to an article, the URL Pelican uses to link to that file is based on the filesystem location rather than the base URL for the article, which can lead to broken links in the site. For example, let's say I set up the following project layout:

- `content/`
    - `attachments/`
        - `test_image.jpg` (contents irrelevant)
        - `test_link.jpg` (contents irrelevant)
    - `test.md` (contents shown below)
- `pelicanconf.py` (contents shown below)

with this `pelicanconf.py`:
```python
SITEURL = 'http://localhost'
TIMEZONE = 'UTC'

THEME = 'simple'
PATH = 'content'
OUTPUT_PATH = 'output'
STATIC_PATHS = ['.']
ARTICLE_URL = '{slug}.html'
ARTICLE_SAVE_AS = 'www/{slug}.html'

FEED_ALL_ATOM = None
CATEGORY_FEED_ATOM = None
TRANSLATION_FEED_ATOM = None
AUTHOR_FEED_ATOM = None
AUTHOR_FEED_RSS = None
```
and this the contents of `content/test.md`:
```markdown
Title: Test
Date: 2015-07-26 21:13:34
Slug: test
Author: Me

![test image]({attach}attachments/test_image.jpg)

[test link]({attach}attachments/test_link.jpg)
```
Then I run `pelican` from the project directory. The HTML file for the post, `output/www/test.html`, will contain
```html
<p><img alt="test image" src="http://localhost/www/attachments/test_image.jpg"></p>
<p><a href="http://localhost/www/attachments/test_link.jpg">test link</a></p>
```
whereas it should contain
```html
<p><img alt="test image" src="http://localhost/attachments/test_image.jpg"></p>
<p><a href="http://localhost/attachments/test_link.jpg">test link</a></p>
```
(without the `/www/` component in the URLs). This is with the latest git master, fa1d4d937c52044472d3243a3c8eb7773fb3e418, cloned locally and installed in a virtualenv using `pip install pelican/`.

I think I've traced this to [line 458 of `contents.py`](https://github.com/getpelican/pelican/blob/master/pelican/contents.py#L458), where the URL of a static file is computed from its filesystem path in kind of a "lazy" way that seems to cause this problem when `ARTICLE_URL` and `ARTICLE_SAVE_AS` don't start with the same prefix. (I suppose the same would be true for pages too, though I didn't test that yet.) But I'm not quite sure what the proper way to solve the issue is - as the comment there says, it's tricky to figure out whether the URL refers to a file or directory.


justinmayer:
Many thanks for the contribution, David. Sorry it took so long to get it merged! 😅

PR #2165, opened 2017-06-08
Fix rsync_upload in Makefile template

pananormalny:
In Makefile template `--cvs-exclude` option for `rsync_upload` is in bad
place - it should be placed before destination.

PoC:

```
/bin/sh: --cvs-exclude: command not found
make: *** [rsync_upload] Error 127
```


justinmayer:
Thanks for the fix, Karol!

PR #2171, opened 2017-06-19
In --autoreload mode, only reprocess modified files

gbrener:
This PR makes Pelican's `--autoreload` mode more efficient by only reprocessing modified files when the watched content changes.

### DIFF
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -488,8 +488,6 @@ def main():
             if next(watchers['theme']) is None:
                 logger.warning('Empty theme folder. Using `basic` theme.')
 
-            modified_files = [val for vals in list(modified.values())
-                              for val in vals]
             pelican.run()
 
     except Exception as e:
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -141,7 +141,7 @@ class Pelican(object):
                 )
                 self.settings[old] = self.settings[new]
 
-    def run(self):
+    def run(self, modified=None):
         """Run the generators and return"""
         start_time = time.time()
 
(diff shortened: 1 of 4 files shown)


justinmayer:
Seems like a very useful enhancement. Any comments on the implementation from @getpelican/reviewers?

avaris:
Unfortunately, this won't work properly. This will generate the site with *only* modified items. IOW, you have a site with, say, 10 articles. Modify 1 article, now you generate a site with 1 article (`index.html` and such will list only a single article). Modify settings or theme, you'll generate an empty site (nothing in it).

gbrener:
@avaris, are you sure about that? During the first iteration of --autoreload, all files are considered "modified" (`LAST_MTIME` == 0). The [pelican.utils.clean_output_directory](https://github.com/gbrener/pelican/blob/autogen_modified/pelican/utils.py#L420-L421) function subsequently avoids removing any files unless they were modified.

avaris:
@gbrener yes, pretty sure. I just tested it. But maybe you misunderstood me. I didn't say anything was removed. I'm not even giving the `-d/--delete-output-directory` option. So, `clean_output_dir` isn't relevant to what I'm saying.

It is true that for the **first iteration** that `autoreload` will generate everything. Subsequent iterations are the problem. 

After the first one, modify one article and check `index.html`: You should see only one article (modified one) will be listed. Other unmodified items will not be there.

Now modify the settings file (or something from the theme) and check `index.html` again: You should see an empty `index.html`, i.e. nothing will be listed there.

gbrener:
Ok, thanks for the feedback. Will try to reproduce.

gbrener:
You're right @avaris - this approach is incomplete. I won't have time to wrap this up, however I will create an issue in case someone else has sufficient interest in this feature to revive it.

PR #2174, opened 2017-06-30
Suppress error messages when PID files cannot be found

PeterMosmans:


iKevinY:
Maybe we should also `return 1` if the PID files are missing to indicate that something went wrong.

PeterMosmans:
Hi @iKevinY wouldn't that be counter-intuitive ? If the PID files aren't there, then it might be that the server is already shut down. One would expect an 'OK' then: the server has been shut down.

iKevinY:
Ah, yeah that makes sense. It's probably fine to leave it as-in then.

justinmayer:
The develop_server.sh script has been removed in a recent pull request, so this is likely moot at this point. Feel free to let me know if that's not the case.

PR #2176, opened 2017-07-10
Remove develop_server.sh in favour of pelican serving static files itself

josch:
Competing static site generators integrate the functionality of regenerating content and serving it into their main executable. In pelican this functionality used to be in an external script `develop_server.sh` which resides in the blog base directory. This has the disadvantage that changes in pelican can break the `develop_server.sh` scripts which will not automatically be upgraded together with pelican by package managers. Thus, pelican should integrate this functionality into its main executable.

To this end, this commit removes `develop_server.sh` and adds three command line options to the pelican executable:

 * `-l/--listen` starts the HTTP server (`-s/--serve` was already taken)
 * `-p/--port` specifies the port to listen at
 * `-b/--bind` specifies the IP to bind to

`--listen` and `--autoreload` can be used together to achieve the same effect that other static site generators offer: Serve files via HTTP while at the same time auto-generating the content.

Since the `develop_server.sh` script was removed, pelican-quickstart looses the `develop` option.

Since the `develop_server.sh` script was removed, the Makefile looses the `stopserver` target and the `devserver` target is replaced by running `pelican -l` in the foreground.

Since pelican now offers the `--listen` option, the fabfile uses that instead of starting the socketserver itself.


josch:
I think on remarkable thing worth mentioning here is, that despite this PR *adding* functionality, it reduces the overall count of lines of code by 78 and thus *reduces* complexity of the overall code base which is a good thing. :)

docs/quickstart.rst
```diff
@@ -61,11 +61,10 @@ ignored for now.)
...
 -----------------
 
-Open a new terminal session and run the following commands to switch to your
-``output`` directory and launch Pelican's web server::
+Open a new terminal session and run the following commands to launch Pelican's
+web server::
 
-    cd ~/projects/yoursite/output
-    python -m pelican.server
+    pelican -l content
```
avaris: `content` is redundant here.

Also there are a couple of assumptions (and possibly issues) here that are not mentioned:
 - if you define `OUTPUT_PATH` (to something non-default) in your settings and your settings is not called `pelicanconf.py` (default), this won't pickup your output path: you should supply the `-s configfile.py` as well
 - if you don't define `OUTPUT_PATH` but instead use `-o/--output`, similar situation: you should define same here as well
 - in most default(ish) cases, this should be run from the project root

pelican/__init__.py
```diff
@@ -327,6 +329,16 @@ def parse_arguments():
...
+    parser.add_argument('-l', '--listen', dest='listen', action='store_true',
+                        help='Serve content files via HTTP and port 8000.')
+
+    parser.add_argument('-p', '--port', dest='port', type=int, default=8000,
+                        help='Port to serve HTTP files at. (default: 8000)')
+
+    parser.add_argument('-b', '--bind', dest='bind', default='127.0.0.1',
+                        help='IP to bind to when serving files via HTTP '
+                        '(default: 127.0.0.1)')
+
```
avaris: `-p/--port` and `-b/--bind` are useless without `-l/--listen`. It would be nice to warn the user if `-p` and/or `-b` used without `-l`.

pelican/__init__.py
```diff
@@ -410,73 +500,20 @@ def main():
...
 
-        if args.autoreload:
+        if args.autoreload and args.listen:
+            autoreload_thread = threading.Thread(
+                target=autoreload, args=(watchers, args, old_static))
+            listen_thread = threading.Thread(
+                target=listen,
+                args=(args.bind, args.port, settings.get("OUTPUT_PATH")))
+            autoreload_thread.start()
+            listen_thread.start()
```
avaris: Well... Did you test this? :)

Couple issues here. 
- Cannot quit this (used to be `^C` aka `KeyboardInterrupt`),  you need to kill the process.
- And exceptions inside threads won't propagate back to main caller. i.e. an exception in `--autoreload` mode with `--debug` is supposed to stop pelican (see line 454) but won't. Again, it hangs.

pelican/server.py
```diff
@@ -3,20 +3,45 @@
...
     SUFFIXES = ['', '.html', '/index.html']
 
+    def translate_path(self, path):
+        # abandon query parameters
+        path = path.split('?', 1)[0]
+        path = path.split('#', 1)[0]
+        # Don't forget explicit trailing slash when normalizing. Issue17324
+        trailing_slash = path.rstrip().endswith('/')
+        try:
+            path = urllib.parse.unquote(path, errors='surrogatepass')
```
avaris: This generates an error in Python 2.7:
```
TypeError: unquote() got an unexpected keyword argument 'errors'
```

docs/quickstart.rst
josch: I just tried to replicate the commend 10 lines above that line where it also tells the user to run `pelican content`. So I think your first two remarks then equally apply to that existing line, no?

Your third remark is justified. I'll fix this.

pelican/__init__.py
josch: Good point. I'll fix that.

pelican/__init__.py
josch: Interesting. Yes, I tested it and pressing `^C` worked fine here and generated a `KeyboardInterrupt` as expected.

I'll take care of the exeception handling but I don't know how to test the problem you face with quitting the server because I cannot reproduce the issue here.

pelican/server.py
josch: This code snippet comes directly from the Python 3.5 source code which explains that it doesn't work with Python 2.7. I guess an easy fix would be to just take the same snippet from the Python 2.7 sources instead because that code should work with Python 3.5.

pelican/__init__.py
avaris: Hmm, apparently it's another Python 2 vs 3 differences. Python 3 seems fine. The issue is with Python 2.7.

docs/quickstart.rst
avaris: Yes and no. `pelican content` will do something even if you have no setting file (default config will be enough). On the other hand, this makes a bunch of assumptions about the output path which may not be true. In that case it won't do anything. But... Given the context of this page, I guess it is fine. Just adding "run from project root" and removing the unnecessary `content` should suffice:

```
pelican --listen
```

josch:
@avaris I updated this PR, addressing your comments. The biggest change was to switch from `threading.Thread` to `multiprocessing.Process` and to use a Queue to communicate exceptions to the main process. Threads seem to not be a good choice here, because we cannot kill threads from the parent, so we use multiprocessing instead. Any exception raised in the subprocesses without being caught now means that both processes will get terminated.

I'd appreciate another review. :)

josch:
Since there were some conflicts, I rebased this branch on top of your master branch.

pelican/settings.py
```diff
@@ -146,6 +146,8 @@ def load_source(name, path):
     'LOAD_CONTENT_CACHE': False,
     'WRITE_SELECTED': [],
     'FORMATTED_FIELDS': ['summary'],
+    'PORT': 8000,
+    'BIND': '',
```
avaris: Can you document these?

pelican/__init__.py
```diff
@@ -382,6 +406,96 @@ def get_instance(args):
...
+            if excqueue is not None:
+                excqueue.put(traceback.format_exception_only(type(e), e)[-1])
+            return
+
+        except Exception as e:
+            if (args.verbosity == logging.DEBUG):
+                if excqueue is not None:
+                    excqueue.put(
+                        traceback.format_exception_only(type(e), e)[-1])
+                return
```
avaris: In `--autoreload` *without* `--listen` but *with* `--debug` exceptions silently ignored. `--autoreload` terminates but you don't see any error. A quick fix suggestion:
```python
if excqueue is not None:
    #...
else:
    raise
```

josch:
@avaris I addressed your new comments in my latest update of this pull request. Thanks!

pelican/__init__.py
```diff
@@ -382,6 +406,97 @@ def get_instance(args):
...
+        logging.error("Could not listen on port %s, server %s.", port, server)
+        if excqueue is not None:
+            excqueue.put(traceback.format_exception_only(type(e), e)[-1])
+        return
+
+    logging.info("Serving at port %s, server %s.", port, server)
+    try:
+        httpd.serve_forever()
+    except Exception as e:
+        if excqueue is not None:
```
Lucas-C: Maybe also add a logline here ? In case `excqueue` is `None`

Lucas-C:
There are also mentions to `develop_server.sh` in https://github.com/getpelican/pelican/blob/master/docs/install.rst and https://github.com/getpelican/pelican/blob/master/docs/publish.rst

justinmayer:
Hey @josch. Thank you for all of your patience. Would you please rebase on current master so we can do a (hopefully final) review?

josch:
Hi @justinmayer, sorry because Pelican didn't do what I wanted, I moved on and am now putting my time elsewhere. But I'm sure, you'll manage to rebase my work as well.

justinmayer:
@Lucas-C: Perhaps you would like to pick up this endeavor and submit a new pull request?

mosra:
I would very much like to have this in, the less files needed on the user side the better :) Wouldn't this also make #2308 / #2312 obsolete?

Lucas-C:
I'd love to continue the work on this, but cannot guarantee when I will have the time for it
(certainly not before next week)

rminderhoud:
I would be interested in this functionality being included in a release so I've gone ahead and rebased @josch's work. I am not familiar with internals of pelican so I will rely on maintainers to review. Please see #2315

Lucas-C:
I said above I'll help on this, so I just rebased @rminderhoud PR #2315 in a new PR #2351.

I hope it'll help to close this issue.
If yout think it only add noise, tell me and I'll gladly close it.

justinmayer:
Superseded by the replacement PR from @Lucas-C.

Issue #1620, opened 2015-02-09
Excluding Pages from DEFAULT_METADATA


nfarrar:
When I set the following in pelicanconf.py:

    DEFAULT_METADATA = {
        'status': 'draft',
    }

My pages break:

    ERROR: Unknown status 'Published' for file pages/about.md, skipping it.

In `PATH` I have the directories 'pages' and 'posts'. I attempted to fix this by explicitly setting:

    ARTICLE_PATHS = ['posts']
    PAGE_PATHS = ['pages']

However this still generates the same errors for pages. Any suggestions?

iKevinY:
I tested this locally, and it seems like you need to use `Status: published` (it's case-sensitive). That being said, perhaps the status check *should* be case-insensitive. Thoughts, @justinmayer?

avaris:
It *should* be case-insensitive, `ArticlesGenerator` [does that](https://github.com/getpelican/pelican/blob/master/pelican/generators.py#L517). For some odd reason, it was [omitted](https://github.com/getpelican/pelican/blob/master/pelican/generators.py#L644) in `PagesGenerator`. So, this is a bug.

iKevinY:
@avaris Ah, thanks for confirming. Do you mind if I submit a PR fixing this?

avaris:
@iKevinY not at all :), I can't submit a PR tonight anyway, so go ahead.

nfarrar:
So, I'm trying to understand what's going on here - I went back and checked the [docs](http://docs.getpelican.com/en/3.5.0/settings.html) and the first thing I noticed was that my syntax was not written the way the docs say it should be, so I adjusted this to:

    DEFAULT_METADATA = (
        ('Status', 'draft'),
    )

... and also using 'draft' (undercase) as you mentioned above. This still generates an error for markdown pages:

    ERROR: Unknown status 'draft' for file pages/about.md, skipping it.

- If I set `Status: draft` on the `about.md` page itself, the same error is generated.
- If I set `Status: published` on the page, no errors are generated.

So it looks like right now you can't set the status of a page to anything except `published`.

This is a separate issue, but also related to `DEFAULT_METADATA`: I set (I also tried using a list, and unicode strings, but nothing worked):

    DEFAULT_METADATA = (
        ('Authors', 'Nathan Farrar'),
    )

Then I get a different unicode error:

    CRITICAL: 'unicode' object has no attribute 'name'

[...]

PR #2181, opened 2017-07-16
Allow directory entries in EXTRA_PATH_METADATA

andrew-vant:
This lets the user apply arbitrary metadata to everything in a directory. My specific use case is changing the stylesheet and template, but it also provides workarounds for cases when DEFAULT_METADATA isn't enough; e.g. applying the [default draft recipe](http://docs.getpelican.com/en/3.6.3/content.html#publishing-drafts) from the docs may not work as expected because [pages and articles accept different status types](https://github.com/getpelican/pelican/issues/1620#issuecomment-73736406).

### DIFF
--- a/pelican/readers.py
+++ b/pelican/readers.py
@@ -698,8 +698,20 @@ def path_metadata(full_path, source_path, settings=None):
         if settings.get('DEFAULT_DATE', None) == 'fs':
             metadata['date'] = SafeDatetime.fromtimestamp(
                 os.stat(full_path).st_mtime)
-        metadata.update(settings.get('EXTRA_PATH_METADATA', {}).get(
-            source_path, {}))
+
+        # Apply EXTRA_PATH_METADATA for the source path and the paths of any
+        # parent directories. Sorting EPM first ensures that the most specific
+        # path wins conflicts.
+
+        epm = settings.get('EXTRA_PATH_METADATA', {})
+        for path, meta in sorted(epm.items()):
+            # Enforce a trailing slash when checking for parent directories.
+            # This prevents false positives when one file or directory's name
+            # is a prefix of another's.
+            dirpath = os.path.join(path, '')
+            if source_path == path or source_path.startswith(dirpath):
+                metadata.update(meta)
+
     return metadata
 
 
--- a/pelican/tests/test_readers.py
+++ b/pelican/tests/test_readers.py
@@ -273,6 +273,38 @@ class RstReaderTest(ReaderTest):
         }
         self.assertDictHasSubset(page.metadata, expected)
 
+    def test_article_extra_path_metadata_recurse(self):
+        parent = "TestCategory"
+        notparent = "TestCategory/article"
+        path = "TestCategory/article_without_category.rst"
+
+        epm = {
+            parent: {'epmr_inherit': parent,
+                     'epmr_override': parent, },
+            notparent: {'epmr_bogus': notparent},
+            path:   {'epmr_override': path, },
+            }
+        expected_metadata = {
+            'epmr_inherit': parent,
+            'epmr_override': path,
+            }
+
+        page = self.read_file(path=path, EXTRA_PATH_METADATA=epm)
+        self.assertDictHasSubset(page.metadata, expected_metadata)
+
+        # Make sure vars aren't getting "inherited" by mistake...
+        path = "article.rst"
+        page = self.read_file(path=path, EXTRA_PATH_METADATA=epm)
+        for k in expected_metadata.keys():
+            self.assertNotIn(k, page.metadata)
+
+        # Same, but for edge cases where one file's name is a prefix of
+        # another.
+        path = "TestCategory/article_without_category.rst"
+        page = self.read_file(path=path, EXTRA_PATH_METADATA=epm)
+        for k in epm[notparent].keys():
+            self.assertNotIn(k, page.metadata)
+
     def test_typogrify(self):
         # if nothing is specified in the settings, the content should be
         # unmodified
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -783,7 +783,8 @@ Metadata
 
    Extra metadata dictionaries keyed by relative path. Relative paths require
    correct OS-specific directory separators (i.e. / in UNIX and \\ in Windows)
-   unlike some other Pelican file settings.
+   unlike some other Pelican file settings. Paths to a directory apply to all
+   files under it. The most-specific path wins conflicts.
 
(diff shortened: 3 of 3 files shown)


justinmayer:
Any comments on this PR from @getpelican/reviewers would be much appreciated. What say you?

ingwinlu:
like the idea, but the usual suspects i.e. test cases and documentation is missing.

andrew-vant:
I did update the documentation -- or at least I thought I did. Is there another piece I should have included?

Acknowledged on test cases. I'll fix that.

ingwinlu:
@andrew-vant nah it's okay, thought we had more on that...

Since you worked on that part of pelican, would you say it is possible to consolidate all the Metadata settings(`FILENAME_METADATA`, `PATH_METADATA` and `EXTRA_PATH_METADATA`) into one?

andrew-vant:
I don't think so, or at least not easily. Some context: When I first started messing with EXTRA_PATH_METADATA, I wondered why it doesn't take regexes. On #pelican, Avaris told me part of the reason is that there's no good deterministic way to handle multiple conflicting matches. I think they were right.

PATH_METADATA and FILENAME_METADATA both rely on regex to do their magic, and EXTRA_PATH_METADATA relies on regexlessness to be deterministic. You could maybe eliminate FILENAME_METADATA on the grounds that everything it can do can already be done by PATH_METADATA, but I don't know if it would gain much.

justinmayer:
Hey Andrew. Many thanks for your contribution. Would you mind adding some tests so we can get it merged?

andrew-vant:
Oops. Back when I opened this, I started using my own branch and forgot I still had the PR open. I'll see what I can do about tests. Thanks for the reminder.

andrew-vant:
Okay, tests added. One test fails on my local machine, but it also failed on the commit I branched from, so I'm guessing it's not my problem.

It's been a long time since I opened this PR. I can rebase the branch on master if you want me to.

andrew-vant:
Ping? Not sure if I need to do anything with this, but I noticed it's still open a couple months later.

justinmayer:
Any last comments from @getpelican/reviewers?

pelican/readers.py
```diff
@@ -630,8 +630,20 @@ def path_metadata(full_path, source_path, settings=None):
...
+        # most specific entry wins.
+
+        paths = []
+        path = source_path
+        while path:
+            paths.append(path)
+            path = os.path.split(path)[0]
+        epm = settings.get('EXTRA_PATH_METADATA', {})
+        for path in reversed(paths):
+            metadata.update(epm.get(path, {}))
```
oulenz: I'm wondering whether it is really necessary to break down the `source_path` into its ancestors like that, and whether it wouldn't be simpler to try to match the `epm` paths:

```
epm = [(p, m) for p, m in epm.items() if p == source_path or source_path.startswith(os.path.join(p, '')))
for _, m in sorted(epm):
    metadata.update(m)
```
In particular, this would automatically skip all comparison if `epm` is empty. (This version also tolerates trailing slashes, so we don't need to warn people against those.)

If you prefer your own version, could you at least add a check whether `epm` isn't empty?

oulenz:
I've added a stylistic comment, and maybe the commits should be squashed. Otherwise, this looks good to me.

pelican/readers.py
avaris: I think it was done this way to guarantee "if multiple EPM entries apply conflicting metadata to the same file, the most specific entry wins."

andrew-vant: I think avaris is right, but it's been a while so I don't remember for sure.

oulenz:
Of course, but that's not the point I was trying to make, and my version
should also let more specific paths overrule others (that's what `sorted`
ensures).

Den man. 25. feb. 2019, 23.21 skrev Andrew Vant <notifications@github.com>:

pelican/readers.py
oulenz: Of course, but that's not the point I was trying to make, and my version should also let more specific paths overrule others (that's what `sorted` ensures). The following is probably clearer:

```
for p, m in sorted(epm.items()):
    if p == source_path or source_path.startswith(os.path.join(p, ''))):
        metadata.update(m)
```

andrew-vant: Hrm. That is clearer, and easier to follow than my version. What does os.path.join do with a blank string, exactly? I'm guessing it adds a slash if one isn't present. I worry that .startswith will produce false positives, but I'm having trouble coming up with a  path that would cause it.

I'll fiddle with it later and maybe update the PR.

oulenz: - yes, it adds a slash if one isn't present. It's a bit hacky, admittedly.
- I don't see how `startswith` could produce false positives with the final slash.

The main thing that got me thinking was that your version generates all the ancestors of `source_path` even if `epm` is empty, which is probably the case for 99% of installations. As long as you address that, I'm happy. :-)

avaris: Is slash necessary? Can't think of a reason why `source_path.startswith(p)` isn't enough.

oulenz: Potentially you could have articles `foo/bar.md` and `foobar/baz.md`, then if you key extra metadata with `foo`, you only want it to affect the former, not the latter.

avaris: Oh, right :).

andrew-vant:
Updated to use Oulenz's suggestion, and I rebased/squashed the commits too. (I still have the old ones if needed)

pelican/tests/test_readers.py
```diff
@@ -273,6 +273,37 @@ def test_article_extra_path_metadata_dont_overwrite(self):
...
+
+        # Make sure vars aren't getting "inherited" by mistake...
+        path = "article.rst"
+        page = self.read_file(path=path, EXTRA_PATH_METADATA=epm)
+        for k in expected_metadata.keys():
+            self.assertNotIn(k, page.metadata)
+
+        # Same, but for edge cases where one file's name is a prefix of
+        # another.
+        page = self.read_file(path=path, EXTRA_PATH_METADATA=epm)
```
oulenz: I think you want to test this with `path = "TestCategory/article_without_category.rst"` as defined at the beginning of this test, but you've redefined  `path` to `"article.rst"` in the previous block.

andrew-vant: Oops, yes, thanks. Fixed.

justinmayer:
@avaris: Everything look okay to you here?

justinmayer:
Many thanks to Andrew for the enhancement and to @ingwinlu, @avaris, and @oulenz for reviewing.

PR #2196, opened 2017-08-07
Make URL part joining aware of absolute URLs

### LINKED ISSUE #2031: {filename} issue when {{ SITEURL }} and {{ STATIC_URL }} are different absolute URLS

mosra:
Previously, with `RELATIVE_URLS` disabled, when both `SITEURL` and `STATIC_URL` were absolute, the final generated data URLs looked wrong like this (two absolute URLs joined by `/`):

    http://your.site/http://static.your.site/image.png

With this patch, the data URLs are correctly:

    http://static.your.site/image.png

This also applies to all `*_URL` configuration options (for example, ability to have pages and articles on different domains) and behaves like one expects even with URLs starting with just `//`, thanks to making use of `urllib.parse.urljoin()`.

However, when `RELATIVE_URLS` are enabled, urllib.parse.urljoin() doesn't handle the relative base correctly. In that case, simple `os.path.join()` is used. That, however, breaks the above case, but as `RELATIVE_URLS` are meant for local development (thus no data scattered across multiple domains), I don't see any problem.

Just to clarify, this is a fully backwards-compatible change, it only enables new use cases that were impossible before.

Properly implements, obsoletes and thus closes #2031 and #2034. Also, besides doing the above, there's a commit that moves the local link replacer function to a `_link_replacer()` function that's callable from outside -- a minimal subset of #2164 that should not do any harm.

Would be nice if this patch could still make it into 3.8, I'm depending on it quite heavily. Thank you a lot!

### DIFF
--- a/pelican/contents.py
+++ b/pelican/contents.py
@@ -228,6 +228,68 @@ class Content(object):
         key = key if self.in_default_lang else 'lang_%s' % key
         return self._expand_settings(key)
 
+    def _link_replacer(self, siteurl, m):
+        what = m.group('what')
+        value = urlparse(m.group('value'))
+        path = value.path
+        origin = m.group('path')
+
+        # XXX Put this in a different location.
+        if what in {'filename', 'attach'}:
+            if path.startswith('/'):
+                path = path[1:]
+            else:
+                # relative to the source path of this content
+                path = self.get_relative_source_path(
+                    os.path.join(self.relative_dir, path)
+                )
+
+            if path not in self._context['filenames']:
+                unquoted_path = path.replace('%20', ' ')
+
+                if unquoted_path in self._context['filenames']:
+                    path = unquoted_path
+
+            linked_content = self._context['filenames'].get(path)
+            if linked_content:
+                if what == 'attach':
+                    if isinstance(linked_content, Static):
+                        linked_content.attach_to(self)
+                    else:
+                        logger.warning(
+                            "%s used {attach} link syntax on a "
+                            "non-static file. Use {filename} instead.",
+                            self.get_relative_source_path())
+                origin = '/'.join((siteurl, linked_content.url))
+                origin = origin.replace('\\', '/')  # for Windows paths.
+            else:
+                logger.warning(
+                    "Unable to find '%s', skipping url replacement.",
+                    value.geturl(), extra={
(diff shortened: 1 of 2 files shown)


mosra:
Eh. Seems that I broke relative links with this change and I can't quite figure out what to change where to make it working again. It's either just `filename.ext` (now, the last path component is dropped) or `..//filename.ext` when I add additional `/` suffix to the string coming out of `get_relative_path()`.

Any ideas? Thank you.

mosra:
Update: looked again into this, rebased onto current master and fixed broken behavior with `RELATIVE_URLS` enabled. The PR description is updated to reflect that. The Travis build is failing only because some flake8 tests that are broken in master as well. Should be safe to merge, in my opinion.

Thanks in advance!

justinmayer:
Thanks, @mosra. I fixed the PyCodeStyle issues via 56a48347 and have re-triggered your PR's build, so everything is green here now. I will take a look at this PR as soon as I can.

mosra:
Oh, I did the same (rebased and triggered the build) after seeing the commit, because I saw it earlier than your comment :)

Thank you a lot!

mosra:
I have a bunch of new features (such as cache busting for static files etc) that depend on this PR (and mainly 1f30306e2329e5a1f0c5dd39844d9bb0a0c04573 in particular -- it just moves a function around, nothing more), would it be possible to get this merged anytime soon so I can rely on something that's already considered stable?

Thank you a lot.

justinmayer:
Apologies for the delay, @mosra. Any comments from @getpelican/reviewers?

justinmayer:
Many thanks for the contribution and for your patience, Vladimír. Much appreciated!

mosra:
Thanks for the merge! :)

PR #2201, opened 2017-08-09
Update tips.rst

zachgoll:
Added troubleshooting support link for User Pages instructions.

### DIFF
--- a/docs/tips.rst
+++ b/docs/tips.rst
@@ -88,6 +88,9 @@ Again, you can take advantage of ``ghp-import``::
 
     $ pelican content -o output -s pelicanconf.py
     $ ghp-import output
+    
+`Make sure you have a valid SSH key <https://help.github.com/articles/adding-a-new-ssh-key-to-your-github-account/>`_ and then push changes to your Github Pages::
+
     $ git push git@github.com:elemoine/elemoine.github.io.git gh-pages:master
 
 The ``git push`` command pushes the local ``gh-pages`` branch (freshly updated


justinmayer:
As [mentioned in IRC](https://botbot.me/freenode/pelican/msg/89617878/), I wonder whether it might be better to link to the HTTPS URL, perhaps along with a note explaining that SSH keys can also be used.

@getpelican/reviewers: Any thoughts?

iKevinY:
Agreed — I think GitHub encourages cloning using HTTPS rather than SSH.

zachgoll:
I do agree with the https URL because as a "beginner," when I started setting up my Github Page site, the first thing I did was search for the relevant section in the Pelican documentation (so definitely agree with keeping it there), and then attempt to run those three commands in bash.

I then spent three hours trying to figure out what the error I was getting was, what an ssh key was, and how to authenticate so that I could use those three commands.  From a usability standpoint, this part of the documentation is necessary and might be easier via https rather than ssh.

justinmayer:
@zachgoll: In that case, would you like to take another pass at your PR?

jorgesumle:
In my opinion there should be no tips to make Pelican work with **only** some third party services. At least not in the official documentation of Pelican. That should be explained in posts of the website of Pelican or tech blogs. The documentation talks about GitHub, but not everyone uses GitHub.

The section on how to add videos is ridiculous, you just have to copy the embed code in the post content, there is no need to install a plugin. It shouldn't be a tip to add videos from services which don't respect the privacy of users and only work with JavaScript enabled. And again the same problem, what if I don't use Vimeo or YouTube? For me it would be more important to give a tip on how to use a HTML 5 video (standard, works everywhere, even with JS disabled) than third party services.

I would just remove this section, or put more relevant content.

zachgoll:
@jorgesumle I do agree that these kind of tutorials should be hosted on Pelican's home website or some external blog.  They are definitely useful tips, but trying to put them on the official documentation pages limits the detail you can go in, which in effect confuses people (like me) who see these sections and try to follow the instructions.  Either increase the detail that these tips go into (not a fan of) or make them more concise by linking out to external tutorials.  @justinmayer I see that the pelican blog itself does not have a place for tutorials though??  Is this something that you see as a useful addition?

justinmayer:
@zachgoll: Yes, I think tutorials posted to the [Pelican Blog](https://github.com/getpelican/pelican-blog) would be quite useful. Please feel free to submit a tutorial article to that repository! :sparkles:

PR #2203, opened 2017-08-12
Fix typo from documentation

jorgesumle:
I think that the documentation should say simply tags (no code formatting). @justinmayer can you confirm it, as you were the one who wrote it four years ago? I may be missing something.


justinmayer:
I agree with your changes here. Thanks, Jorge!

Issue #2139, opened 2017-04-18
Drop Python 3.3 Support


ssteinerx:
I'm working on Pelican itself and the Python 3.3 tests are failing because I don't have, never have had, and am not going to have Python 3.3.  It's in "source only" mode at this point so it's a huge PITA to install.

Kind of a minor bookkeeping issue, but it's used so little (and even less i.e. not at all by the developers who are the target audience for this project) that I would think dropping it now makes sense.

- It's in "source only" mode
- Hasn't been patched in over a year
- Support will completely end in a few months (Sept 2017)
- Wouldn't impact anyone
- Would speed up every `tox` run
- Would let me say that all tests pass without a 3.3 caveat 😺 .

justinmayer:
I agree that it's time to drop official support for Python 3.3. Any opinions from @getpelican/reviewers on this topic?

ingwinlu:
if it is not just cutting 3.3 tags from the test environment and setup.py, i.e. also remove any 3.3 traces in the code that we might have in there...

hugovk:
Python 3.3 support ends in 10 days.
 
https://en.wikipedia.org/wiki/CPython#Version_history

avaris:
Source-wise, I think [`test_testsuite.py`](https://github.com/getpelican/pelican/blob/master/pelican/tests/test_testsuite.py#L12) has the only 3.3 specific piece of code apart from documentation.

justinmayer:
Thanks, Avaris. I removed that reference in the CONTRIBUTING docs.

Anyone want to submit a pull request that removes Python 3.3 references from the test environment, setup.py, and any other needed places?

oulenz:
The documentation still says under `Installing Pelican` that `Pelican currently runs best on Python 2.7.x and 3.3+; earlier versions of Python are not supported.` Just wanted to point this out because I was trying to get the tests to work under 3.3 before I saw this. (I can't install Pelican under 3.3 because `feedgenerator` requires 3.4.)

jorgesumle:
Actually the documentation from the file `docs/install.rst` says this:

> Pelican currently runs best on Python 2.7.x and 3.4+; earlier versions of Python are not supported.

The text you quoted is not in the development version of Pelican, so I suppose you must have confused looking at some outdated version of the documentation.

avaris:
You need to look at the `latest` documentation: http://docs.getpelican.com/en/latest/install.html

`stable` is the last release (3.7.1) which was before these changes.

[...]

PR #2216, opened 2017-10-02
Drop unsupported Python 3.3

hugovk:
Resolves https://github.com/getpelican/pelican/issues/2139.

(The unrelated 3.5 error should hopefully be resolved by #2217.)


justinmayer:
Thanks for tackling this, Hugo. A month ago I undertook a [similar endeavor for another project](https://github.com/bennylope/django-organizations/pull/138/commits/66d91eff12074b786f6806eedde14b87db62145b), which I think may offer a more comprehensive approach to this issue. Would you, and perhaps some @getpelican/reviewers,  take a look and let me know your thoughts?

justinmayer:
Upon further consideration, I think this approach is simpler and probably sufficient for our current needs. Thanks for the assistance, Hugo!

PR #2222, opened 2017-10-10
Wordpress import - Count post_titles with a single space as blank

stuaxo:
When importing my wordpress blog, I ended up with loads of filenames "-.md".

These were drafts with no title, because the post_title xml attribute returned a string with a single space.
Possibly I should check against any amount of spaces.

Any objection if I make draft posts named like  draft_{postid}  ?


stuaxo:
With this patch I can import without using --dir-page and --dir-cat, without  it I need --dir-page and --dir-cat as all the "-.md" files collide.

PR #2224, opened 2017-10-11
RFC - pelican-import, fix urls of attached files

stuaxo:
Don't merge this yet - this probably needs some tweaks..

With this patch wp-attach modifies ```<img``` and  ```<a href``` urls to point at the attachment path - so  ```http://example.com/wp-content/uploads/blah.png```   becomes    ```wp-content/uploads/blah.png```

This needs wp-content in pelicanconf STATIC_FILES

I thought I could use {filename} but since the importer is generates source html this doesn't work.
It would be nice if there was a way to customise the output path, but not sure how that would work.


hitrust:
very nice!

stuaxo:
Apologies for the many commits, I've been tweaking this in the web editor!

stuaxo:
I should probably fix the URLs after the content has been converted by pandoc whatever format to ```{filename}path/to/file```.

This has the advantage that the code will get a little simpler too.

stuaxo:
This now changes the url to have a leading slash: ```{filename}/path/to/file``` and works with files generated by pandoc.

stuaxo:
I'm going to see if I can extract the functionality as a set of smaller PRs

oulenz:
I wonder whether this functionality hasn't already been added by https://github.com/getpelican/pelican/pull/2380? (BTW, sorry for not referencing this PR, I hadn't seen it at the time.)

stuaxo:
Mostly, the thumbnails aren't getting fixed up, but I'm looking into it.

I'm going to close this PR, as I think between the but you linked, the smaller PRs this should all be covered.

PR #2226, opened 2017-10-13
Fix metadata intrasite links

charlesfleche:
Fix intrasite links for non-'summary' metadata

Metadata like `MyArticleBanner: alt text` would be property parsed (as defined in `FORMATTED_FIELDS`), but the intrasite links would not be processed.

Only the summary gets its intrasite links processed, has its value is either generated from the content (calling self._update_content at some point) or self._update_content is explicitly called if summary is passed as metadata.

This PR expands the paths as soon as possible in (`Content.__init__`) for metadata defined in `FORMATTED_FIELDS`.

### DIFF
--- a/pelican/contents.py
+++ b/pelican/contents.py
@@ -151,8 +151,8 @@ class Content(object):
 
 
         # store the summary metadata if it is set
-        if 'summary' in metadata:
-            self._summary = metadata['summary']
+        if 'summary' in self.metadata:
+            self._summary = self.metadata['summary']
 
         signals.content_object_init.send(self)
 
@@ -348,7 +348,7 @@ class Content(object):
         content.
         """
         if hasattr(self, '_summary'):
-            return self._update_content(self._summary, siteurl)
+            return self._summary
 
         if self.settings['SUMMARY_MAX_LENGTH'] is None:
             return self.content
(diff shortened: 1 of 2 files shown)


charlesfleche:
The failing test is from `master`. No other tests fails here.

charlesfleche:
This PR is certainly related to #2164, but fixes the issue without introducing a new Jinja filter.

charlesfleche:
Something looks wrong with this patch : `context['filenames']` is empty at that point, so if the code works in the tests, in doesn't in production. I'm investigating.

charlesfleche:
That should work betterr: interlinks in metadata are parsed during `Content.__init__`, keeping compatibility with existing code. But by exposing the `Content.refresh_metadata_intersite_links` both at the `Content` and `Generator` level, those interlinks can be refreshed at a later time when the contexts are generated in `Pelican.run`.

justinmayer:
Hi @charlesfleche. Many thanks for the contribution. Would you care to tidy up a bit by squashing any fix-up commits? Looking forward to merging. Thanks!

charlesfleche:
Not sure to understand: isn't squashing something you do at merge time in GitHub ?
https://help.github.com/articles/about-pull-request-merges/

justinmayer:
I can squash all of them into a single commit, yes. I just wanted to check and see if you'd prefer to organize them into fewer (but still multiple) logical commits.

PR #2227, opened 2017-10-16
FIX: freenode.org/freenode.net

JulienPalard:


justinmayer:
Thanks for the fix, @JulienPalard. It seems that HTTPS is supported, so would you be willing to update the web link in this PR with the following address?  https://freenode.net/

JulienPalard:
Sure! Done!

justinmayer:
Great. Thanks, Julien!

PR #2234, opened 2017-10-24
Add setting so template extensions are configurable

OmegaDroid:
I had needed to generate svg's based on data so I've added a setting to extend the template extensions to search for.


justinmayer:
Many thanks for the contribution, Daniel. And thanks to @mosra for reviewing!

Issue #1615, opened 2015-02-07
pagination_patterns settings example in documentation requires additional configuration


blanche:
If someone wants to use the PAGINATION_PATTERNS example as provided in the documentation (see below), it will only for the main page. When generating tag/categories/authors the fixed "index.html" causes name conflicts.

original documentation:

```
PAGINATION_PATTERNS = (
    (1, '{base_name}/', '{base_name}/index.html'),
    (2, '{base_name}/page/{number}/', '{base_name}/page/{number}/index.html'),
)
```

this causes `tag/my-tag.html` to be renamed to `tag/index.html`
the next tag `tag/my-other-tag.html` also renamed to `index.html` will cause an error 

```
CRITICAL: File /path/to/blog/output/tag/index.html is to be overwritten
```
### Additional settings required to fix the issue:

```
TAG_URL = 'tag/{slug}/'
TAG_SAVE_AS = 'tag/{slug}/index.html'
TAGS_URL = 'tags/'
TAGS_SAVE_AS = 'tags/index.html'

AUTHOR_URL = 'author/{slug}/'
AUTHOR_SAVE_AS = 'author/{slug}/index.html'
AUTHORS_URL = 'authors/'
AUTHORS_SAVE_AS = 'authors/index.html'

CATEGORY_URL = 'category/{slug}/'
CATEGORY_SAVE_AS = 'category/{slug}/index.html'
CATEGORYS_URL = 'categories/'
CATEGORYS_SAVE_AS = 'categories/index.html'
```

with these setting, the generates pages will look like the following and achieve the desired result:

```
# main page
index.html
page/2/index.html

# categories/tags/authors
category/misc/index.html
category/misc/page/2/index.html
```

blanche:
if someone could confirm this issue, i will try to provide an update for the documentation

smargh:
Confirmed. Just ran into this bug tonight. Exactly as you described.

number5:
Have the same issue here.  Additional config hadn't fixed it.

#1137 is also about this issue

number5:
The workaround works with latest master branch 249a91

mseri:
Same issue for me. The workaround is not working with pelican 3.5.0

ingenioustechie:
This solution is not working with pelican 3.5.0

ingwinlu:
no idea why the sample uses `base_name` instead of `name`.

thoughts on changing it?

justinmayer:
[...]

Issue #1137, opened 2013-10-26
Custom PAGINATION_PATTERNS not working correctly


calfzhou:
Pelican v3.3.0 document provides an example setting of PAGINATION_PATTERNS (http://docs.getpelican.com/en/3.3.0/settings.html#using-pagination-patterns), which is:

``` python
PAGINATION_PATTERNS = (
    (1, '{base_name}/', '{base_name}/index.html'),
    (2, '{base_name}/page/{number}/', '{base_name}/page/{number}/index.html'),
)
```

But this configuration will cause a **CIRTICAL** error if there are more than one tags (or categories). The error is:

```
CRITICAL: File <blahblah>/output/tag/index.html is to be overwritten
make: *** [html] Error 1
```

The reason is that for example there are two tags, tagA and tagB. When processing tagA, it will generate `output/tag/index.html`. And for tagB, it will try to generate another `output/tag/index.html`, but this file already exists.

A solution is to add a `{name}` placeholder to the patterns, such as:

``` python
PAGINATION_PATTERNS = (
    (1, '{base_name}/', '{base_name}/{name}/index.html'),
    (2, '{base_name}/page/{number}/', '{base_name}/{name}/page/{number}/index.html'),
)
```

This will not cause any error, but the generate tags (or categories) files are: `output/tag/tag/tagA/index.html` and `output/tag/tag/tagB/index.html`. As you could see, there are two level of `tag` folder. And for posts, their index pages will be saved in `output/index/index.html` and `output/index/page/2/index.html`.

veddermatic:
I agree, having to choose between not having pagination or having "bad" URLs doesn't seem like a good choice.

PR #2235, opened 2017-10-26
Allow to use page URL in pagination patterns

mosra:
Currently it was only possible to use page "save as" name or part of it for generating pagination links. That's not sufficient when page URLs differ a lot from actual filenames. With this patch it's possible to use the `{url}` placeholder in the `PAGINATION_PATTERNS` setting. For example, the paginated archives would be saved to:

    blog/index.html
    blog/2/index.html
    blog/3/index.html

while the actual URLs would be like this (with the help of Apache's `mod_rewrite`, for example):

    http://blog.my.site/
    http://blog.my.site/2/
    http://blog.my.site/3/

The configuration that corresponds to this is roughly the following:

```py
ARCHIVES_SAVE_AS = 'blog/index.html'
ARCHIVES_URL = 'http://blog.my.site/'
PAGINATION_PATTERNS = [
    (1, '/{url}', '{base_name}/index.html'),
    (2, '/{url}{number}/', '{base_name}/{number}/index.html')
]
```

**Note:** I added a test case for this, but didn't update any docs as the various placeholders are not mentioned anywhere in the documentation so I didn't know where to put that. Possibly related (regarding documentation): #1137 #1615

Thank you!

### DIFF
--- a/pelican/writers.py
+++ b/pelican/writers.py
@@ -141,7 +141,7 @@ class Writer(object):
         return feed
 
     def write_file(self, name, template, context, relative_urls=False,
-                   paginated=None, override_output=False, **kwargs):
+                   paginated=None, override_output=False, url=None, **kwargs):
         """Render the template and write the file.
 
         :param name: name of the file to output
@@ -153,6 +153,7 @@ class Writer(object):
         :param override_output: boolean telling if we can override previous
             output with the same name (and if next files written with the same
             name should be skipped to keep that one)
+        :param url: url of the file (needed by the paginator)
         :param **kwargs: additional variables to pass to the templates
         """
 
@@ -202,7 +203,7 @@ class Writer(object):
         if paginated:
 
             # pagination needed, init paginators
-            paginators = {key: Paginator(name, val, self.settings)
+            paginators = {key: Paginator(name, url, val, self.settings)
                           for key, val in paginated.items()}
 
             # generated pages, and write
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -464,13 +464,28 @@ respectively.
 
    The location to save per-year archives of your posts.
 
+.. data:: YEAR_ARCHIVE_URL = ''
+
+   The URL to use for per-year archives of your posts. Used only if you have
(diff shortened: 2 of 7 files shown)


mosra:
Okay, the Travis builds fails because of flake8 checks, but these errors are the same as with current master and I assume something got different on Travis and caused these:

```
pelican/tests/support.py:58:15: E741 ambiguous variable name 'l'
pelican/tools/pelican_quickstart.py:17:1: E722 do not use bare except'
pelican/tools/pelican_quickstart.py:117:52: E741 ambiguous variable name 'l'
pelican/tools/pelican_quickstart.py:183:13: E722 do not use bare except'
pelican/tools/pelican_themes.py:20:1: E722 do not use bare except'
```

justinmayer:
Thanks, @mosra. I fixed the PyCodeStyle issues via 56a48347 and have re-triggered your PR's build, so everything is green here now. I will take a look at this PR as soon as I can.

mosra:
Great, thank you!

justinmayer:

pelican/writers.py
```diff
@@ -140,11 +140,12 @@ def write_feed(self, elements, context, path=None, feed_type='atom',
                 complete_path, context=context, feed=feed)
         return feed
 
-    def write_file(self, name, template, context, relative_urls=False,
+    def write_file(self, name, url, template, context, relative_urls=False,
```
avaris: This is quite a change and could possibly break some plugins. Did you do any checks? Somewhat *nicer* way would be making this parameter kwarg.

mosra: Made `url` into an optional parameter, defaulting to `None`. Is that okay? :)

avaris: Yeah, should be fine now. Thanks.

justinmayer:
Excellent description of the problem to be solved, references to related issues, documentation, and test coverage. If only all pull requests looked like this one! 😁 

Bravo, @mosra. Many thanks for the contribution. (And thanks @avaris for reviewing!)

PR #2238, opened 2017-10-29
Feed URL settings, making feeds aware of absolute URLs

mosra:
With this PR, there are new `FEED_*_URL` configuration options that allow to specify custom URLs for feeds, which is helpful in case the feed filename and the actual URL differ a lot -- for example if a feed is saved to
    
    blog/feeds/all.atom.xml
    
but the actual URL from the user PoV is
    
    http://blog.your.site/feeds/all.atom.xml
    
This setting affects the generated feed XML (both RSS and Atom) and also the default `simple` and `notmyidea` themes were updated to be aware of the `*_URL` settings. This change is also fully backwards compatible, so if the `FEED_*_URL` setting is not present, the value of `FEED_*` is used for both file location and URL, as it was before.

Besides that, the feed generator is now aware of absolute URLs. So for example in case both `SITEURL` and `ARTICLE_URL` is absolute, the generated URLs are not improperly
    
    http://your.site/http://blog.your.site/the-article/
    
but rather
    
    http://blog.your.site/the-article/

This particular change is closely related to #2196, is referring to the same design decisions and should behave exactly the same (but it's not depending on it, so these PRs can be merged independently).

**Note:** Since feed generation and `RELATIVE_URL`s don't really make sense together, the URL joining code could be made a bit simpler, assuming the URLs in feeds should always be absolute. I just made it exactly the same way as in #2196 for consistency.

**Also:** I'm aware that the new feed URL settings are not very well tested, but I couldn't find a place where the feed generation could be tested separately without adding extensive amount of new sample data. Help in this direction very appreciated.

Thanks in advance!

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -811,46 +811,95 @@ the ``TAG_FEED_ATOM`` and ``TAG_FEED_RSS`` settings:
 
 .. data:: FEED_ATOM = None, i.e. no Atom feed
 
-   Relative URL to output the Atom feed.
+   The location to save the Atom feed.
+
+.. data:: FEED_ATOM_URL = None
+
+   Relative URL of the Atom feed. If not set, ``FEED_ATOM`` is used both for
+   save location and URL.
 
 .. data:: FEED_RSS = None, i.e. no RSS
 
-   Relative URL to output the RSS feed.
+   The location to save the RSS feed.
+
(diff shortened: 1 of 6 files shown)


mosra:
@justinmayer Any chance to get this merged? :)

justinmayer:
Many thanks for the contribution, @mosra!

PR #2240, opened 2017-10-29
Make use of SITESUBTITLE setting also in Atom feeds

mosra:
For some reason, `feedgenerator.py` uses the `description` argument only for RSS and the `subtitle` argument only for Atom. So setting both to the same value.
    
In order to avoid unnecessary changes, if `SITESUBTITLE` is not present, the `subtitle` is set to `None` instead of `''`, so the generated Atom feed doesn't contain an empty `<subtitle></subtitle>` tag (in contrast, the RSS feed *does contain* an empty `<description></description>` tag now, though).

I updated the tests to make use of the `SITESUBTITLE` setting in both RSS and Atom feeds but it kinda blow up across all sample output files, so if you think that's excessive, I can discard the second commit.

Thanks again!


mosra:
@justinmayer Any chance to get this merged? :)

justinmayer:
Many thanks for the contribution, and your patience, @mosra.

Issue #2239, opened 2017-10-29
pelican-import: handle the multiple categories from Wordpress


fgallaire:
As #349 and #350 are closed with no fix, I re-open here.

I'm trying to migrate two Wordpress blogs that have multiple categories by post.

The most important problem is that the great ``pelican-import`` is broken for a widespread Wordpress feature (whether it is a good or a bad one is not the point here). And I think it will have more and more Wordpress migrations in the coming years.

The most objectionable problem is that Pelican is not even coherent in its way of handling multiple categories. ``pelican-import --dir-cat`` take the first categorie as directorie name, but the Category metadata is the list of the categories separated by commas, and treated by Pelican as a weird category name.

I'm sure nobody wants this result:

    Categories

    Documentation (2)
    Documentation, Media (1)
    Documentation, Site (2)
    Documentation, Tools (1)
    Documentation, Translations (13)

justinmayer:
Addressed by #2434.

PR #2241, opened 2017-10-30
WP import: Category is categories[0]

fgallaire:
Same behaviour as --dir-cat
Fix part of #2239

### DIFF
--- a/pelican/tools/pelican_import.py
+++ b/pelican/tools/pelican_import.py
@@ -523,7 +523,7 @@ def build_header(title, date, author, categories, tags, slug,
     if author:
         header += ':author: %s\n' % author
     if categories:
-        header += ':category: %s\n' % ', '.join(categories)
+        header += ':category: %s\n' % categories[0]
     if tags:
         header += ':tags: %s\n' % ', '.join(tags)
     if slug:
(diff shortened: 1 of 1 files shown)


ssteinerx:
I haven't looked at the code but, assuming this once worked, is `categories` always a list or can it be a string if there's only one?

Also, is there a test that fails with current code, but passes with new?

fgallaire:
I modify nothing, ``categories`` is still the list of the wordpress categories of the post and I only take the first category to become the pelican "who only supports one category by design" category.
This is the same behaviour as the --dir-cat option, making the whole coherent.

fgallaire:
Needed to avoid the creation of list categories nobody wants as import result:

```
Documentation (2)
Documentation, Media (1)
Documentation, Site (2)
Documentation, Tools (1)
Documentation, Translations (13)
```

justinmayer:
@fgallaire: Many thanks for the contribution, Florent. My apologies for the long delay. I think perhaps the plugin that @oulenz documented in #2434 obviates the need to make changes to how multiple categories are handled by the WP importer in Pelican core. That said, if you still feel further changes in core are warranted, please comment here and we can re-open. Thanks again! 😅

PR #2246, opened 2017-11-03
Make THEME_STATIC_PATHS work for files as well

mosra:
(Sorry for opening so many PRs lately. I hope that's okay...)

The documentation implies that `THEME_STATIC_PATHS` should work for files as well, but that was not the case. Fixed that and added a test.

My use case is the following: a [theme I'm using](http://mcss.mosra.cz/) provides multiple CSS color styles and for publishing I need just one of them. So instead of 

```py
THEME_STATIC_PATHS = ['static']
```

I say, for example, just 

```py
THEME_STATIC_PATHS = ['static/m-dark.compiled.css']
```

and thus all the other, unneeded, files from the `static/` dir are ignored.

This PR should be totally harmless without any backward compatibility breakages, so I think it's safe to merge. Thank you.

### DIFF
--- a/pelican/tests/test_generators.py
+++ b/pelican/tests/test_generators.py
@@ -699,6 +699,53 @@ class TestStaticGenerator(unittest.TestCase):
     def set_ancient_mtime(self, path, timestamp=1):
         os.utime(path, (timestamp, timestamp))
 
+    def test_theme_static_paths_dirs(self):
+        """Test that StaticGenerator properly copies also files mentioned in
+           TEMPLATE_STATIC_PATHS, not just directories."""
+        settings = get_settings(
+            PATH=self.content_path,
+            filenames={})
+        context = settings.copy()
+        context['staticfiles'] = []
+
+        StaticGenerator(
+            context=context, settings=settings,
+            path=settings['PATH'], output_path=self.temp_output,
+            theme=settings['THEME']).generate_output(None)
+
+        # The content of dirs listed in THEME_STATIC_PATHS (defaulting to
+        # "static") is put into the output
+        self.assertTrue(os.path.isdir(os.path.join(self.temp_output,
+                                                   "theme/css/")))
+        self.assertTrue(os.path.isdir(os.path.join(self.temp_output,
+                                                   "theme/fonts/")))
+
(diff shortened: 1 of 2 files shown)


mosra:
(Travis `flake8` job fails because of import order in `pelican/__init__.py` (which should probably be ignored?), unrelated to this. Why did it break again all of a sudden?)

justinmayer:
@mosra: Thank you for the pull request and for bringing the Flake8 test failure to attention. It broke again because of this: https://github.com/PyCQA/flake8-import-order/commit/95fe1b1f2ed1db6051acd8d1163dfb12a6b4784a

The `flake8-import-order` 0.14.3 release contains that commit, which seems to have unintended side effects. I filed an issue here: https://github.com/PyCQA/flake8-import-order/issues/120

justinmayer:
I addressed the flake8-import-order issue via 8ebc120, so all lights are green here now.  🎉

justinmayer:

pelican/tests/test_generators.py
```diff
@@ -699,6 +699,53 @@ def tearDown(self):
...
+        # Only the content of dirs and files listed in THEME_STATIC_PATHS are
+        # put into the output, not everything from static/
+        self.assertFalse(os.path.isdir(os.path.join(self.temp_output,
+                                                    "theme/css/")))
+        self.assertFalse(os.path.isdir(os.path.join(self.temp_output,
+                                                    "theme/fonts/")))
+        self.assertTrue(os.path.isfile(os.path.join(self.temp_output,
+                                                    "theme/font.css")))
+        self.assertTrue(os.path.isfile(os.path.join(self.temp_output,
+                                                    "theme/fonts.css")))
```
avaris: This looks like there is only two files but there will be other files as well ([from `static/fonts`](https://github.com/getpelican/pelican/tree/master/pelican/themes/notmyidea/static/fonts)). I'd prefer to list and test them all here.

mosra: Done. :)

justinmayer:
Such effort put into the tests... Be still my heart! ❤️ 

Sincere thanks to @mosra for the contribution, and to @avaris for reviewing.

PR #2249, opened 2017-11-09
preserve_case option for category names slugification (fix #2248)

### LINKED ISSUE #2248: Category names slugification problem
In ``pelican-import`` with the ``--dir-cat`` option, categorie names are slugified to generate the directory names.

But it's not coherent with the handling of the Category header where there's no slugification.

This leads to problems in very simple cases. After an import if you had the category `Documentation`, all the files are in the directory ``documentation`` and have ``Categorie: Documentation`` in the header. The post category is shown as `Documentation` . Now create a new content file whitout Categorie header, the post category is shown as `documentation`.

A same category should only have one and only one visible name.

So it could be possible to have the slugified category names added in the header. But I'm sure I don't want that since I prefere to have an uppercase first letter category names.

Slugification is for URL, but it this really needed for filenames (i.e. file system question) ?
Or maybe make category names slugification an option ?

fgallaire:


fgallaire:
I have amended the commit just to restart a completly failing Travis run. Loose time because of this shit (false error messages) 2 times on 3 PR, it's a shame.

ssteinerx:
I haven't looked further, but what happens when a set of case sensitive categories hits a case insensitive file system?

fgallaire:
Although I shaven't done tests, I think it change nothing from an insensitive file system point of view.

justinmayer:
Thoughts on this PR from @getpelican/reviewers? (especially @avaris)

avaris:
It'd be nice to add a test or two.

justinmayer:
@fgallaire: Would you mind adding some corresponding tests? 😊

fgallaire:
@justinmayer The tests are run with preserve_case enabled as it will be the default option. Do you want more ?

avaris:
@fgallaire I meant just a couple basic tests for `slugify` to directly test that option. Something like:
```python
slugify('Foo') == 'foo'
slugify('Foo', preserve_case=False) == 'foo'
slugify('Foo', preserve_case=True) == 'Foo'
```
in [here](https://github.com/getpelican/pelican/blob/master/pelican/tests/test_utils.py#L123)

justinmayer:
@fgallaire: Would you mind adding a few tests as @avaris suggested so we can merge your contribution?

justinmayer:
Hi @fgallaire. Any chance you could add the tests @avaris mentioned so we can merge your pull request?

justinmayer:
These changes are now in master.

PR #2251, opened 2017-11-09
Change imported content directory name to "content". Fixes #2250

### LINKED ISSUE #2250: pelican-import output directory confusing name
The pelican output directory default name is ``output`` for html generated files, but the pelican-import output directory default name is ``output`` for content source files.

This could be very confusing for beginers. I think that a sane pelican-import output directory default name should be ``content``, as it's the pelican default name for content source files directory.

fgallaire:
… #2250)


justinmayer:
Many thanks to Florent for the fix and to Deniz for the review.  (^_^)

PR #2252, opened 2017-11-13
Fixed pandoc invocation while importing feeds

kotnik:
Newer version of `pandoc` normalize by default. If this legacy option is present, `pandoc` fails to run.

### DIFF
--- a/pelican/tools/pelican_import.py
+++ b/pelican/tools/pelican_import.py
@@ -728,8 +728,7 @@ def fields2pelican(
                 fp.write(new_content)
 
             parse_raw = '--parse-raw' if not strip_raw else ''
-            cmd = ('pandoc --normalize {0} --from=html'
-                   ' --to={1} -o "{2}" "{3}"')
+            cmd = ('pandoc --from=html --to={1} -o "{2}" "{3}"')
             cmd = cmd.format(parse_raw, out_markup,
                              out_filename, html_filename)
 


justinmayer:
Hi Nikola. Thanks for the contribution. It seems tests are failing for this PR. Did you run the tests locally? More more details, see: http://docs.getpelican.com/en/stable/contribute.html

justinmayer:
Hi @kotnik. Any chance you could resolve the issue with failing tests in your PR?

jelly:
This requires another change since parse_raw is also deprecated in favor of +raw_html. I believe the fx is:

```
            cmd = ('pandoc {0} --from=html+raw_html'
                   ' --to={1} -o "{2}" "{3}"')
```

dtzWill:
Anyone have a chance to revisit this? Would adding the change suggested above (perhaps using the commit by @yan12125) be worth trying w/CI?  Increasingly pandoc commonly available is newer than this breakage point, so this'd be good to fix...? :)  Thoughts/comments welcome, thanks all!

davidag:
@dtzWill If you're using pandoc 2, this was fixed in PR #2366. What version of Pelican and pandoc are you using?

justinmayer:
Thanks to @davidag, I believe this has already been fixed in master. If that is not the case, please open a new issue with details regarding your environment and steps to reproduce.

PR #2259, opened 2017-12-10
Fix documentation for *_SAVE_AS in themes.rst

ix5:
Bring up to par with the defaults in the [settings]( http://docs.getpelican.com/en/stable/settings.html#url-settings) documentation


justinmayer:
Many thanks, @ix5!

PR #2269, opened 2017-12-27
Support SSL with pelican.server by adding --ssl , --cert and --key options

tonymet:
## Updated 4/1/2018
`pelican.server` now supports `--ssl` , `--cert` and `--key`.  see `tips.rst` for documentation

## Intro
This WIP PR adds SSL support to the local devserver (`make devserver` or `python -m pelican.server`).  SSL is turned on for any port ending in `443` e.g. `8443` .   Set `PORT=8443` in your sites `Makefile`. Otherwise HTTP is used. 

## Motivation
Testing with SSL helps catch SSL-only bugs.  Dev & prod should be as similar as possible.

#### Steps to Enable HTTPS
1. Create self-signed cert & key (see below)
1. add `PORT=8443` to your site's Makefile
1. update `pelicanconf.py` , `SITEURL=https://localhost:8443/` 
1. `make devserver`

#### Generating the self-signed certificates:

```
# openssl will prompt for the "FQHN" -- use the hostname part of your `SITEURL` from `pelicanconf.py`
$ openssl req -x509 -newkey rsa:4096 -keyout key.pem -out cert.pem -days 365 -nodes
$ make html
$ cp *pem output/
$ make devserver
```

## Discussion
Do you like the idea?  If so, I can add a `pelican createcert` command to wrap `openssl`.  Let me know the UX you'd like.

### DIFF
--- a/pelican/server.py
+++ b/pelican/server.py
@@ -1,8 +1,10 @@
 # -*- coding: utf-8 -*-
 from __future__ import print_function, unicode_literals
 
+import argparse
 import logging
 import os
+import ssl
 import sys
 
 try:
@@ -12,6 +14,26 @@ except ImportError:
 
 from six.moves import SimpleHTTPServer as srvmod
 from six.moves import socketserver
+
+
+def parse_arguments():
+    parser = argparse.ArgumentParser(
+        description='Pelican Development Server',
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter
+    )
+    parser.add_argument("port", default=8000, type=int, nargs="?",
+                        help="Port to Listen On")
+    parser.add_argument("server", default="", nargs="?",
+                        help="Interface to Listen On")
+    parser.add_argument('--ssl', action="store_true",
+                        help='Activate SSL listener')
+    parser.add_argument('--cert', default="./cert.pem", nargs="?",
+                        help='Path to certificate file. ' +
+                        'Relative to current directory')
+    parser.add_argument('--key', default="./key.pem", nargs="?",
+                        help='Path to certificate key file. ' +
+                        'Relative to current directory')
+    return parser.parse_args()
 
 
 class ComplexHTTPRequestHandler(srvmod.SimpleHTTPRequestHandler):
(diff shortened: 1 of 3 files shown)


justinmayer:
I personally like the idea of using HTTPS everywhere, even locally. What do other @getpelican/contributors think?

avaris:
Agreed, though I don't like the implementation. I'd rather have an explicit switch (e.g. `--ssl` cli option) instead of a magical port number. Also, It'd be better to be able to provide the files instead of hardcoded `key.pem`/`cert.pem` (again with a cli option). And this needs to be documented ([here](http://docs.getpelican.com/en/latest/quickstart.html#preview-your-site)?).

tonymet:
i like the feedback. if you guys want to come up with the cli interface (e.g. `--ssl` and `--certfile` ) , i'm happy to help code it up> 

Also, i need the interface for generating the certs e.g. `pelican --generate-self-signed-cert`. 

I think if you guys come up with the interface it'll be more consistent. i can do the legwork. 

Lastly, is it ok if I shell out to "openssl" rather than trying to futz around with some python module for creating certs?

tonymet:
also, I'll need the naming for the `peliconconf.py` variables.  That way folks don't need the cli flags with every invocation of `pelican`

tonymet:
@avaris  @justinmayer i'd love to help with this. could you give me feedback on the cli & pelicanconf.py interface?

avaris:
I'm not sure it's reasonable to get the configs in `pelicanconf.py`. Server doesn't get that right now. This is for `pelican.server` not `pelican` itself. They won't be needed for each `pelican` invocation. Only if you want to renew server.

Also, I don't think you need an `pelican` interface to generate certificates. That's a one-off thing to do. Just a note in the docs about how to do it, or links to places would suffice.

And I was suggesting the cli options for `pelican.server`, not `pelican` itself. So, you'd do something like:

```shell
python -m pelican.server --port 8443 --ssl --key foo/key.pem --cert foo/cert.pem
```

tonymet:
@avaris I mostly kept your interface.  I retained the PORT & SERVER positional arguments for compatibility

```
~/sotion/pelican/venv/bin/python -m pelican.server --help
usage: server.py [-h] [--ssl] [--cert [CERT]] [--key [KEY]] [port] [server]

Pelican Development Server

positional arguments:
  port           Port to Listen On (default: 8000)
  server         Interface to Listen On (default: )

optional arguments:
  -h, --help     show this help message and exit
  --ssl          Activate SSL listener (default: False)
  --cert [CERT]  Path to certificate file. Relative to current directory
                 (default: ./cert.pem)
  --key [KEY]    Path to certificate key file. Relative to current directory
                 (default: ./key.pem)
```

justinmayer:
Seems reasonable to me. What do you think, @avaris?

pelican/server.py
```diff
@@ -14,6 +16,27 @@
...
+        parser.add_argument('--ssl', action="store_true",
+                            help='Activate SSL listener')
+        parser.add_argument('--cert', default="./cert.pem", nargs="?",
+                            help='Path to certificate file. ' +
+                            'Relative to current directory')
+        parser.add_argument('--key', default="./key.pem", nargs="?",
+                            help='Path to certificate key file. ' +
+                            'Relative to current directory')
+        self.options = parser.parse_args()
+
```
avaris: Why is this a class? If you wanted to leave it out of `__main__` block for clarity sake, you could just make it a function that returns `parser.parse_args()`.

avaris: Like [this](https://github.com/getpelican/pelican/blob/master/pelican/__init__.py#L262) and [usage](https://github.com/getpelican/pelican/blob/master/pelican/__init__.py#L395)

tonymet: @avaris not a problem i will update

tonymet:
@avaris class changed to `parse_arguments`

justinmayer:
Many thanks to @tonymet for the enhancement and to @avaris for reviewing! 🚀

tonymet:
@justinmayer thanks guys for all the clear feedback and support. i'm really proud to finally be a contributor, even if it's just something tiny.

PR #2278, opened 2018-01-22
No need to mention SimpleHTTPServer in quickstart script

lqez:
In `pelican_quickstart.py`, there is still `simpleHTTP` in the question for `develop` option. I'm pretty sure that it means `SimpleHTTPServer` of Python 2. From `3.6`, Pelican provides unified http serving via `pelican.server`, so I think we don't have to show it.

### DIFF
--- a/pelican/tools/pelican_quickstart.py
+++ b/pelican/tools/pelican_quickstart.py
@@ -267,7 +267,7 @@ needed by Pelican.
 
     automation = ask('Do you want to generate a Fabfile/Makefile '
                      'to automate generation and publishing?', bool, True)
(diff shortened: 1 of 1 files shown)


avaris:
I think, it's better to say `HTTP server` instead:

```python
develop = ask('Do you want an auto-reload & HTTP server script '
```

lqez:
@avaris I made it simpler.

justinmayer:
Many thanks to @lqez for the contribution and to @avaris for the review.  🎊

lqez:
Thanks!

Issue #2263, opened 2017-12-14
handle_ref() cannot handle &gid=


ix5:
When you have links formatted like this: `http://www.linkedin.com/groupAnswers?viewQuestionAndAnswers=&discussionID=12741155&gid=87954&trk=EML_anet_qa_ttle-0Pt79xs2RVr6JBpnsJt7dBpSBA)`, the `?gid=` part makes the pelican HTMLParser hiccup when truncating for feeds:
(This popped up while helping ramonsuarez on his bug #2258 )

```
CRITICAL: ValueError: substring not found
Traceback (most recent call last):
  File "/home/me/data/pip/lib/python3.6/site-packages/pelican/utils.py", line 556, in handle_entityref
    codepoint = html_entities.name2codepoint[name]
KeyError: 'gid'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/me/data/pip/bin/pelican", line 11, in <module>
    sys.exit(main())
  File "/home/me/data/pip/lib/python3.6/site-packages/pelican/__init__.py", line 487, in main
    pelican.run()
  File "/home/me/data/pip/lib/python3.6/site-packages/pelican/__init__.py", line 179, in run
    p.generate_output(writer)
  File "/home/me/data/pip/lib/python3.6/site-packages/pelican/generators.py", line 599, in generate_output
    self.generate_feeds(writer)
  File "/home/me/data/pip/lib/python3.6/site-packages/pelican/generators.py", line 300, in generate_feeds
    self.settings['FEED_ALL_ATOM'])
  File "/home/me/data/pip/lib/python3.6/site-packages/pelican/writers.py", line 123, in write_feed
    self._add_item_to_the_feed(feed, elements[i])
  File "/home/me/data/pip/lib/python3.6/site-packages/pelican/writers.py", line 52, in _add_item_to_the_feed
    description = item.summary
  File "/home/me/data/pip/lib/python3.6/site-packages/pelican/contents.py", line 310, in summary
    return self.get_summary(self.get_siteurl())
  File "/home/me/data/pip/lib/python3.6/site-packages/pelican/utils.py", line 173, in __call__
    value = self.func(*args)
  File "/home/me/data/pip/lib/python3.6/site-packages/pelican/contents.py", line 306, in get_summary
    self.settings['SUMMARY_MAX_LENGTH'])
  File "/home/me/data/pip/lib/python3.6/site-packages/pelican/utils.py", line 583, in truncate_html_words
[...]

PR #2285, opened 2018-02-08
Fix utils.truncate_html_words() to work with invalid HTML references

andreacorbellini:
Invalid references like those missing semicolons (e.g. `&mdash`) or
those causing overflows (e.g. `&#9999999999;`) are now gracefully
handled and no exception is thrown.

This commit also adds tests and comments where needed.

Fixes issue #2263.


justinmayer:
Any comments from @getpelican/reviewers?

justinmayer:
> It would be good to get a second pair of eyes to review this as well, in case someone has more familiarity with the nuances of HTML & Unicode than I do!

Our resident Unicode shaman @avaris is just the person for the job!  😁

justinmayer:
Any thoughts on this one, @avaris?

Issue #2162, opened 2017-06-06
Release Pelican 4.0


justinmayer:
This issue tracks the release of Pelican 4.0.

W7PEA:
I was just doing some site updates today and thought I'd ask if there was anything we can do to help get an update out.

Thanks so  much for Pelican!

jorgesumle:
Some things may not work after the upgrade (for example, the `make serve` instruction from Makefiles generated with old Pelican versions). Ideally, we should talk about that in a blog post and give advice on how to upgrade.

justinmayer:
@jorgesumle: Agreed. Would you be interested in starting that process by adding a stub/draft release post in the [Pelican Blog repository](https://github.com/getpelican/pelican-blog)?

jorgesumle:
> Would you be interested in starting that process by adding a stub/draft release post in the Pelican Blog repository?

Sure

jorgesumle:
> > Would you be interested in starting that process by adding a stub/draft release post in the Pelican Blog repository?
> Sure

Sorry, but I didn't know about [the lack of license](https://github.com/getpelican/pelican-blog/issues/9) of the Pelican blog. I'm against this kind of restrictive copyright, so unless I can submit my contribution under a [libre culture](https://freedomdefined.org/Libre) license, I won't be working on that.

justinmayer:
Unless there is sufficient reason to do otherwise, I am inclined to release Pelican 3.8, based on current master, within the next few days. So if anyone feels there is something outstanding they want to see included in this release, now is the time to take appropriate action to that effect. Just wanted to communicate that and gives folks a heads-up. 😊

avaris:
> I found an issue with #2226 / #2288 where replacement of metadata links is somehow depending on order in which pages are processed (so links referencing pages that are not yet processed are reported as Unable to find). Not good. Will investigate and try to post a patch later today.

Commented there.

> One of my plugins (m.metadata) is not requiring page title in reST sources but since #2311, Pelican warns me that this is illegal. Not sure what to do here -- what about requiring at most one title instead of exactly one title? Technically, I could see title being optional also in case of pages (for example when using a custom template for a particular page).

This is less about titles but more about metadata. reST requires a particular layout to be able to parse metadata. All content in core pelican requires a `title`. So I'm fine with that warning. I'd suggest [suppressing warnings](https://docs.python.org/2/library/warnings.html#temporarily-suppressing-warnings) in your plugin if you want.

[...]

PR #2288, opened 2018-02-09
Fix metadata intrasite links

charlesfleche:
Squashed version of previous PR https://github.com/getpelican/pelican/pull/2226

### DIFF
--- a/pelican/contents.py
+++ b/pelican/contents.py
@@ -140,9 +140,8 @@ class Content(object):
         if not hasattr(self, 'status'):
             self.status = getattr(self, 'default_status', None)
 
-        # store the summary metadata if it is set
-        if 'summary' in metadata:
-            self._summary = metadata['summary']
+        if len(self._context.get('filenames', [])) > 0:
+            self.refresh_metadata_intersite_links()
 
         signals.content_object_init.send(self)
 
--- a/pelican/contents.py
+++ b/pelican/contents.py
@@ -413,6 +413,16 @@ class Content(object):
                     os.path.abspath(self.source_path),
                     os.path.abspath(self.settings['PATH']))))
 
+    def refresh_metadata_intersite_links(self):
+        for key in self.settings['FORMATTED_FIELDS']:
+            if key in self.metadata:
+                value = self._update_content(
+                    self.metadata[key],
+                    self.get_siteurl()
+                )
+                self.metadata[key] = value
+                setattr(self, key.lower(), value)
+
 
 class Page(Content):
     mandatory_properties = ('title',)
--- a/pelican/contents.py
+++ b/pelican/contents.py
@@ -337,8 +337,8 @@ class Content(object):
         This is based on the summary metadata if set, otherwise truncate the
         content.
         """
-        if hasattr(self, '_summary'):
-            return self._update_content(self._summary, siteurl)
+        if 'summary' in self.metadata:
+            return self.metadata['summary']
 
         if self.settings['SUMMARY_MAX_LENGTH'] is None:
             return self.content
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -170,6 +170,10 @@ class Pelican(object):
             if hasattr(p, 'generate_context'):
                 p.generate_context()
 
+        for p in generators:
+            if hasattr(p, 'refresh_metadata_intersite_links'):
+                p.refresh_metadata_intersite_links()
+
         signals.all_generators_finalized.send(generators)
 
         writer = self.get_writer()
--- a/pelican/generators.py
+++ b/pelican/generators.py
@@ -581,6 +581,14 @@ class ArticlesGenerator(CachingGenerator):
         self.generate_pages(writer)
         signals.article_writer_finalized.send(self, writer=writer)
 
+    def refresh_metadata_intersite_links(self):
+        for e in chain(self.articles,
+                       self.translations,
+                       self.drafts,
+                       self.drafts_translations):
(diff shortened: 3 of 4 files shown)


charlesfleche:
Ah, just noticed I probably could have pushed to the previous branch with `git push origin +name-of-branch`...

justinmayer:
Hehe. No worries. I was just about to let you know that you could have pushed to the previous branch/PR, but it seems you noticed that on your own.

Thanks again for the contribution!

pelican/contents.py
```diff
@@ -140,9 +140,8 @@ def __init__(self, content, metadata=None, settings=None,
         if not hasattr(self, 'status'):
             self.status = getattr(self, 'default_status', None)
 
-        # store the summary metadata if it is set
-        if 'summary' in metadata:
-            self._summary = metadata['summary']
+        if len(self._context.get('filenames', [])) > 0:
+            self.refresh_metadata_intersite_links()
```
mosra: @charlesfleche @justinmayer As I commented in https://github.com/getpelican/pelican/issues/2162#issuecomment-406952691, this line causes link replacement to be dependent on the order in which files are processed -- because it's done during parsing of a single article/page and not later when everything is ready (so e.g. links to pages in metadata fields that were not yet processed are reported as `Unable to find`).

I investigated further and realized `refresh_metadata_intersite_links()` is called again later below, after all articles/pages are processed -- and in that second step the links that were previously `Unable to find` are then replaced correctly. I don't know the code very well, but in my case when I commented the above two lines out, the warnings were gone (and links in metadata fields still correctly replaced). I imagine removing the above might cause some plugin to cease working (since e.g. the `summary` metadata field won't be present yet), so not really sure how to fix that. On the other hand, I don't think calling `refresh_metadata_intersite_links()` *twice* on everything is a good idea (at least from a perf PoV, but I could imagine some potential issues with `{attach}` as well).

Any ideas how to fix this?

avaris: Agreed, this is not ideal.

Removing these won't make `summary` vanish, but it'll be unprocessed if one accesses before everything is processed. And that is actually fine, since accessing `.summary` in a plugin earlier than `all_generators_finalized` is a "bug" in itself for this same reason (it'll spew `Unable to find` warnings).

Another issue is that, this removed `_summary`. Even though it's "internal", some plugins used that.

From what I can see, this modification is all for preventing `summary` being processed twice. I'd say, reverting this bit and `get_summary` should be enough.

And maybe excluding `summary` from `refresh_metadata_intersite_links`, if one is worried enough about the double processing.

mosra: Finally got back to this (sorry for the delay), the proposed change is in #2406.

> And maybe excluding summary from refresh_metadata_intersite_links, if one is worried enough about the double processing.

Not sure what you mean, reverting the above bit is (as far as I can see) enough to stop the double processing.

avaris: I meant, if you revert the above part *and `get_summary`* then `summary` will be processed twice.
* Once when `refresh_metadata_intersite_links` is called
* Once when `content.summary` is accessed (or any other way `.get_summary` is called).

Furthermore, reverting only this part is pointless, since you end up with a `._summary` attribute that's not used anywhere.

avaris:
Also... Did anyone test this metadata intrasite links with `RELATIVE_URLS = True`? I have a strong suspicion that it won't work...

pelican/contents.py
mosra: Um, I'm afraid I lost all the context over the time we're discussing this :see_no_evil:

> Furthermore, reverting only this part is pointless, since you end up with a ._summary attribute that's not used anywhere.

I did that only to fix this:

> Another issue is that, this removed `_summary`. Even though it's "internal", some plugins used that.

So, what I ended up doing in #2406 is:

- keeping `_summary`, because some plugins used that, as you said
- not calling `refresh_metadata_intersite_links` too soon

I could also just delete this part altogether (and thus `_summary` would be no longer). Can we move the discussion to #2406?

mosra:
You mean because now it's processed only once (and exactly once) while before it was processed on each access and thus being able to generate different values based on where the content/summary is accessed from?

(Disclaimer: I personally only tried enabling `RELATIVE_URLS` once (long before this PR was merged) and realized it didn't work well in my case where I have URLs like `/blog/category/article/`: it didn't generate proper amount of `../` parts in the relative URLs and I didn't bother further.)

pelican/contents.py
avaris: Sure, but `_summary` is not used in `get_summary` anymore. This PR also changes that. I mean those should be reverted together.

Plugins modifying `_summary` do so in order to change `.summary`, since you can't directly access it too early. Now they will access and change something irrelevant.

avaris:
Exactly.

PS: If you're getting invalid links with `RELATIVE_URLS = True`, either there is a config error or a bug. You should report if it is the latter :).

pelican/contents.py
mosra: Oh! Now I get you. I thought `_summary` was accessed only for read, never for write. Now it all makes sense, sorry for my denseness.

#2406 is updated (the CI is failing for an unrelated reason, same on master), but instead of just reverting the second part I realized the changes to `_summary` won't be picked up by `refresh_metadata_intersite_links()` and the `_summary` handling would need to creep there as well. So now I introduced a `_summary` property getter/setter that's basically just a "view" onto `metadata["summary"]`. Also, "because some plugins used that" I assume it's not a workflow worth preserving so the getter/setter now prints a deprecation warning when the property is used.

I hope this is finally getting closer to how it should be :) And sorry for the delays, I'm extremely busy.

PR #2289, opened 2018-02-11
Make pelican-import support pandoc2. Fixes #2255

### LINKED ISSUE #2255: pelican-import corrupts posts imported from WordPress
Hello,

I tried to move my blog ( https://eax.me/ ) from WordPress to Pelican using pelican-import. It works for the most part. There were a few difficulties but I managed to find workarounds for most of them. Particularly pelican-import doesn't handle images well and doesn't support CodeColorer plugin so I had to use some regular expressions (see below).

Also I discovered that pelican-import doesn't work with pandoc 2.0.2 properly so I had to patch the script around line 729:

```
            parse_raw = ''
            cmd = ('pandoc {0} --from=html'
                   ' --to={1}+raw_html -o "{2}" "{3}"')
            cmd = cmd.format(parse_raw, out_markup,
                             out_filename, html_filename)
```

These are minor issues though. The most serious issues are the following.

1. After migration characters like `'`, `"` and `$` were replaced to `\'`, `\"` and `\$`. I can't just replace these sequences back to `'`, `"` and `$` since these sequences sometimes are used in code snippets. There is a similar issue with the dash symbol that after migrations turn from `---` to `\-\--`.

2. Also in some code snippets code like `#include <something.h>` just turns into `#include`.

3. Last but not least one of the posts created by pelican-import (see below) hangs `make devserver` with 100% CPU usage.

You can reproduce all these issues using the XML file that I exported from WordPress. Please leave your email and I will send it to you.

Exact steps to reproduce:

```
perl -pi -e 's/\[cci.*?\]/<code>/g' ~/temp/del-me/wordpress.2017-11-29.xml
perl -pi -e 's/\[\/cci.*?\]/<\/code>/g' ~/temp/del-me/wordpress.2017-11-29.xml
perl -pi -e 's/\[cc_(\w+).*?\]/<pre><code>:::$1/g' ~/temp/del-me/wordpress.2017-11-29.xml
perl -pi -e 's/\[cc.*?lang="(\w+)".*?\]/<pre><code>:::$1/g' ~/temp/del-me/wordpress.2017-11-29.xml
perl -pi -e 's/\[cc.*?\]/<pre><code>/g' ~/temp/del-me/wordpress.2017-11-29.xml
perl -pi -e 's/\[\/cc.*?\]/<\/code><\/pre>/g' ~/temp/del-me/wordpress.2017-11-29.xml
perl -pi -e 's/<img .*?((src=".*?" alt=".*?")|(alt=".*?" src=".*?)).*?>/<img $1 \/>/g' ~/temp/del-me/wordpress.2017-11-29.xml
perl -pi -e 's/<span style="white\-space: nowrap;">(.*?)<\/span>/$1/g' ~/temp/del-me/wordpress.2017-11-29.xml
perl -pi -e 's!"https?:\/\/eax\.me\/([a-z0-9\-_]+)\/"!"https:\/\/eax.me\/$1.html"!g' ~/temp/del-me/wordpress.2017-11-29.xml
pelican-import -o content/ -m markdown --wpfile ~/temp/del-me/wordpress.2017-11-29.xml
```
Run `make devserver` to make sure that content/mojolicious.md hangs the server (issue 3). Temporary replace it somewhere and run `make devserver` again.

Then see:

```
http://localhost:8000/diy-presentation-remote.html
http://localhost:8000/elliptic-curves-crypto.html
```
^ symbols `$`, `'` and `"` turned into `\$`, `\'` and `\"` (issue 1).

```
http://localhost:8000/cpp-gtest.html
```
^ `#include <something.h>` was replaced to `#include` (issue 2).

Pelican version is 3.7.1.

davidwilemski:
This patch allows `pelican-import` to select the correct flags to use based on whether the version of `pandoc` installed is greater than 2.0.

I've only tested with >2.0 on my machine but have tried to ensure that the regex actually works as advertised to support both older Pandoc installs and those that have been upgraded.

This could be simplified if support for Pandoc <2.0 was dropped but I don't have enough info about how frequently various package managers update in order to suggest only supporting a newer major version.

### DIFF
--- a/pelican/tools/pelican_import.py
+++ b/pelican/tools/pelican_import.py
@@ -774,7 +774,7 @@ def fields2pelican(
 def is_pandoc2_installed():
     cmd = ['pandoc', '--version']
     try:
-        output = subprocess.check_output(cmd)
+        output = subprocess.check_output(cmd).decode('utf-8')
     except subprocess.CalledProcessError:
         return None
 
--- a/pelican/tools/pelican_import.py
+++ b/pelican/tools/pelican_import.py
@@ -682,6 +682,9 @@ def fields2pelican(
         dircat=False, strip_raw=False, disable_slugs=False,
         dirpage=False, filename_template=None, filter_author=None,
         wp_custpost=False, wp_attach=False, attachments=None):
+
+    pandoc2 = is_pandoc2_installed()
+
     for (title, content, filename, date, author, categories, tags, status,
             kind, in_markup) in fields:
         if filter_author and filter_author != author:
@@ -727,9 +730,13 @@ def fields2pelican(
 
                 fp.write(new_content)
 
-            parse_raw = '--parse-raw' if not strip_raw else ''
-            cmd = ('pandoc --normalize {0} --from=html'
-                   ' --to={1} -o "{2}" "{3}"')
+            parse_raw = '--parse-raw' if not strip_raw and not pandoc2 else ''
+            if pandoc2:
+                cmd = ('pandoc {0} --from=html '
+                       '--to=gfm+raw_html -o "{2}" "{3}"')
+            else:
+                cmd = ('pandoc --normalize {0} --from=html'
+                       ' --to={1} -o "{2}" "{3}"')
             cmd = cmd.format(parse_raw, out_markup,
                              out_filename, html_filename)
 
@@ -762,6 +769,18 @@ def fields2pelican(
         print("downloading attachments that don't have a parent post")
         urls = attachments[None]
         download_attachments(output_path, urls)
+
+
+def is_pandoc2_installed():
+    cmd = ['pandoc', '--version']
+    try:
+        output = subprocess.check_output(cmd)
+    except subprocess.CalledProcessError:
+        return None
+
+    if re.search(r'^pandoc 2\..+$', output, re.M):
+        return True
+    return False
 
 
 def main():


davidwilemski:
It appears even with the Py3 fix for my regex, this still needs some work:

- .rst may not be handled as an output extension any more
- There appears to be some issue with test_code_in_list but it isn't clear what that problem is yet. I'm not yet sure whether this is due to some change in Pandoc or a bug in my patch.

```
(pelican-dev-35) ~/pelican [fix-2255 ↑·1|✔]
21:17 $ python -Wd -m unittest discover
............................................................................................F..............WARNING: /Users/dtw/.virtualenvs/pelican-dev-35/lib/python3.5/site-packages/bs4/builder/_lxml.py:132: DeprecationWarning: inspect.getargspec() is deprecated, use inspect.signature() instead
  self.parser.feed(data)

...FE............................................................................................................
ERROR: test_code_in_list (pelican.tests.test_importer.TestWordpressXmlImporter)
  File "/Users/dtw/pelican/pelican/tests/test_importer.py", line 267, in test_code_in_list
    sample_line = re.search(r'-   This is a code sample', md).group(0)
AttributeError: 'NoneType' object has no attribute 'group'

FAIL: test_invalid_symlink_is_overwritten (pelican.tests.test_generators.TestStaticGenerator)
  File "/Users/dtw/pelican/pelican/tests/test_generators.py", line 955, in test_invalid_symlink_is_overwritten
    self.assertEqual(os.path.realpath(self.endfile), self.startfile)
AssertionError: '/private/var/folders/wd/q8nyqm5x2lx_sx6r8y[49 chars]file' != '/var/folders/wd/q8nyqm5x2lx_sx6r8ypg4p3800[41 chars]file'
- /private/var/folders/wd/q8nyqm5x2lx_sx6r8ypg4p380000gn/T/testcontent.4on9kz_l/static/staticfile
? --------
+ /var/folders/wd/q8nyqm5x2lx_sx6r8ypg4p380000gn/T/testcontent.4on9kz_l/static/staticfile


FAIL: test_can_toggle_raw_html_code_parsing (pelican.tests.test_importer.TestWordpressXmlImporter)
  File "/Users/dtw/pelican/pelican/tests/test_importer.py", line 202, in test_can_toggle_raw_html_code_parsing
    self.assertTrue(any('<iframe' in rst for rst in rst_files))
AssertionError: False is not true

Ran 220 tests in 15.074s
```

davidwilemski:
Hrrm. Well, the fact that CI passed but the tests fail for me locally means that either Pandoc 2 changed behavior even farther or I'm not using the correct combination of flags to get identical behavior to previous versions of Pandoc.

CI is not using pandoc2. From the build logs:
> Setting up pandoc (1.12.2.1-1build2) ...

davidwilemski:
I think that the result for `test_invalid_symlink_is_overwritten` noted above is not related to this issue. It appears that it's some issue with symlink comparison, potentially a MacOS specific issue since that's what I've been using for development thus far.

I'm working on getting an environment with Pandoc <2 up and running so that I can test this change on both versions.

pelican/tools/pelican_import.py
```diff
@@ -727,9 +730,13 @@ def fields2pelican(
 
                 fp.write(new_content)
 
-            parse_raw = '--parse-raw' if not strip_raw else ''
-            cmd = ('pandoc --normalize {0} --from=html'
-                   ' --to={1} -o "{2}" "{3}"')
+            parse_raw = '--parse-raw' if not strip_raw and not pandoc2 else ''
```
ashwinvis: If I were you, I would have placed this in the `if-else` construct below.

pelican/tools/pelican_import.py
```diff
@@ -727,9 +730,13 @@ def fields2pelican(
...
+            if pandoc2:
+                cmd = ('pandoc {0} --from=html '
+                       '--to=gfm+raw_html -o "{2}" "{3}"')
```
ashwinvis: Aren't you assuming `gfm` here? How about this:

```python
if pandoc2:
    parse_raw = '+raw_html' if not strip_raw else ''
    cmd = ('pandoc --from=html --to={1}+{0} -o "{2}" "{3}"')
else:
....
```

justinmayer:
@davidwilemski: Any comments on the feedback here? What are the next steps for this?

sandzwerg:
FYI: I manual used the version from this PR and it worked for me with pelican 3.7.1, pandoc 2.0.6 and markdown as output format. Would love to see this merged. Options used: ` pelican-import --wpfile  -m markdown  --dir-page `. Thanks everyone involved for the work :)

davidag:
I'm also having this issue and I'd prefer to help fix it than use a quick-hack. Is there anything I can do to help move this fix forward? Thank you!

davidag:
Sure, I've been recently playing with pandoc 2 to do [something similar](https://gitlab.com/davidalfonso/tiddlywiki-migrator) to what Pelican does, although probably much less generic.

justinmayer:
Closing in favor of #2366.

PR #2290, opened 2018-02-12
Register function missing in docs

benjaminabel:


justinmayer:
Many thanks to @benjaminabel for the contribution and to @ingwinlu for the review.  🎊

PR #2292, opened 2018-02-16
Remove hardcoded sorting of articles within categories, tags, authors, feeds

oulenz:
Currently, `ARTICLE_ORDER_BY` only affects the index page, it gets overwritten for category, tag and author pages and feeds. This should fix that. Passes tests under 2.7, 3.4, 3.5 and 3.6.

### DIFF
--- a/pelican/generators.py
+++ b/pelican/generators.py
@@ -21,8 +21,9 @@ from pelican import signals
 from pelican.cache import FileStampDataCacher
 from pelican.contents import Article, Page, Static
 from pelican.readers import Readers
-from pelican.utils import (DateFormatter, copy, mkdir_p, posixize_path,
-                           process_translations, python_2_unicode_compatible)
+from pelican.utils import (DateFormatter, copy, mkdir_p, order_content,
+                           posixize_path, process_translations,
+                           python_2_unicode_compatible)
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


oulenz:
@avaris Fixed, sorry! I didn't realise that flake8 specifically tests for PEP 8 adherence, and that the errors it produces are intended and not indicative of some infrastructure problem. Any chance it could be added to the offline test?

avaris:
@oulenz add to which offline test? We have it in [`tox` config](https://github.com/getpelican/pelican/blob/master/tox.ini#L36).

oulenz:
I am not sure how tox works. Please correct me if I am wrong, but I now get the impression that there exists a workflow for running all these tests locally that is not mentioned under http://docs.getpelican.com/en/latest/contribute.html#running-the-test-suite . I just ran `python -Wd -m unittest discover` in four different conda environments.

avaris:
@oulenz [`tox`](https://tox.readthedocs.io/en/latest/#basic-example) is mentioned a bit [further](http://docs.getpelican.com/en/latest/contribute.html#testing-on-python-2-and-3). It's basically installing and then running `tox` at pelican root folder.

oulenz:
Thanks, I may come up with a pr for the docs at some point in the future, to give that a bit more prominence.

justinmayer:
@avaris: Everything look good to you here?

justinmayer:
Many thanks to @oulenz for the contribution and to @avaris for reviewing. 🤘

Issue #2303, opened 2018-03-12
articles with duplicate slugs


oulenz:
I have a project where a large number of articles have the same slug, and so I get swamped in warnings `There are 2 original (not translated) items with slug` so-and-so. I could either just ignore these warning or hack together some way to differentiate the slugs, like in #1427.

But I'd rather contribute a cleaner solution. I am thinking of either a setting that suppresses this warning, or a  setting that skips looking for translations altogether. Would this be desirable, and if so, which would you prefer?

voidplayer:
Not sure if this is bug of feature

I was using the slug as a part of the url as in ```subcategory/slug``` so even tho there were many articles with the same slug, there wasnt really a url conflict

I just removed the slug and added a metadata called ```path``` and used that instead of the slug to craft the urls

No more warnings :)

avaris:
A setting to turn off translation processing would be a fine solution.

*Warning*: This might end up being a moderate sized refactoring. Currently article ordering is embedded into translation processor. A decent solution would require decoupling them, which is the preferred behavior (at least for me :) ).

oulenz:
@avaris I already had to decouple them in another pull request. :-) Would be great if you or someone else could review that https://github.com/getpelican/pelican/pull/2292

oulenz:
I thought of another potential solution: give people the option (by way of a new value for ` SLUGIFY_SOURCE`) to incorporate the category of an article in its slug (e.g. `category/title`). That would even allow for otherwise duplicate slugs in combination with translations.

(Disabling translations could still be a setting independent of this.)

PR #2309, opened 2018-03-22
Control translation identification scope with new translation_id settings

oulenz:
Fixes https://github.com/getpelican/pelican/issues/2303, dependent on https://github.com/getpelican/pelican/pull/2292.

Prevents duplicate slug warnings if the articles belong to different categories. A cursory look at the [i18n_subsites plugin](https://github.com/getpelican/pelican-plugins/tree/master/i18n_subsites) suggests it shouldn't be affected by this, but I'd be happy to receive confirmation from someone more knowledgeable.

@avaris

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -1038,6 +1038,18 @@ more information.
 
    The default language to use.
 
+.. data:: ARTICLE_TRANSLATION_ID = 'slug'
+
+   The metadata attribute(s) used to identify which articles are translations
+   of one another. May be a string or a collection of strings. Set to ``None``
+   or ``False`` to disable the identification of translations.
+
+.. data:: PAGE_TRANSLATION_ID = 'slug'
+
+   The metadata attribute(s) used to identify which pages are translations
+   of one another. May be a string or a collection of strings. Set to ``None``
+   or ``False`` to disable the identification of translations.
+
 .. data:: TRANSLATION_FEED_ATOM = 'feeds/all-%s.atom.xml'
 
    The location to save the Atom feed for translations. [3]_
--- a/pelican/generators.py
+++ b/pelican/generators.py
@@ -597,12 +597,14 @@ class ArticlesGenerator(CachingGenerator):
                 all_drafts.append(article)
             self.add_source_path(article)
 
-        self.articles, self.translations = process_translations(all_articles)
-        self.articles = order_content(
-            self.articles,
-            order_by=self.settings['ARTICLE_ORDER_BY'])
-        self.drafts, self.drafts_translations = \
-            process_translations(all_drafts)
+        def _process(arts):
+            origs, translations = process_translations(
+                arts, translation_id=self.settings['ARTICLE_TRANSLATION_ID'])
+            origs = order_content(origs, self.settings['ARTICLE_ORDER_BY'])
+            return origs, translations
+
+        self.articles, self.translations = _process(all_articles)
+        self.drafts, self.drafts_translations = _process(all_drafts)
 
         signals.article_generator_pretaxonomy.send(self)
 
@@ -701,12 +703,15 @@ class PagesGenerator(CachingGenerator):
                 draft_pages.append(page)
             self.add_source_path(page)
 
-        self.pages, self.translations = process_translations(all_pages)
-        self.pages = order_content(self.pages, self.settings['PAGE_ORDER_BY'])
-        self.hidden_pages, self.hidden_translations = \
-            process_translations(hidden_pages)
-        self.draft_pages, self.draft_translations = \
-            process_translations(draft_pages)
+        def _process(pages):
+            origs, translations = process_translations(
+                pages, translation_id=self.settings['PAGE_TRANSLATION_ID'])
+            origs = order_content(origs, self.settings['PAGE_ORDER_BY'])
+            return origs, translations
+
+        self.pages, self.translations = _process(all_pages)
+        self.hidden_pages, self.hidden_translations = _process(hidden_pages)
+        self.draft_pages, self.draft_translations = _process(draft_pages)
 
         self._update_context(('pages', 'hidden_pages', 'draft_pages'))
 
--- a/pelican/utils.py
+++ b/pelican/utils.py
@@ -639,77 +639,94 @@ def escape_html(text, quote=True):
     return escape(text, quote=quote)
 
 
-def process_translations(content_list):
+def process_translations(content_list, translation_id=None):
     """ Finds translations and returns them.
 
-    Returns a tuple with two lists (index, translations).  Index list includes
+    For each content_list item, populates the 'translations' attribute, and
+    returns a tuple with two lists (index, translations). Index list includes
     items in default language or items which have no variant in default
     language. Items with the `translation` metadata set to something else than
-    `False` or `false` will be used as translations, unless all the items with
-    the same slug have that metadata.
+    `False` or `false` will be used as translations, unless all the items in
+    the same group have that metadata.
 
-    For each content_list item, sets the 'translations' attribute.
+    Translations and original items are determined relative to one another
+    amongst items in the same group. Items are in the same group if they
+    have the same value(s) for the metadata attribute(s) specified by the
+    'translation_id', which must be a string or a collection of strings.
+    If 'translation_id' is falsy, the identification of translations is skipped
+    and all items are returned as originals.
     """
-    content_list.sort(key=attrgetter('slug'))
-    grouped_by_slugs = groupby(content_list, attrgetter('slug'))
-    index = []
-    translations = []
 
+    if not translation_id:
+        return content_list, []
+
+    if isinstance(translation_id, six.string_types):
+        translation_id = {translation_id}
+
+    index = []
+
+    try:
+        content_list.sort(key=attrgetter(*translation_id))
+    except TypeError:
+        raise TypeError('Cannot unpack {}, \'translation_id\' must be falsy, a'
+                        'string or a collection of strings'
+                        .format(translation_id))
+    except AttributeError:
+        raise AttributeError('Cannot use {} as \'translation_id\', there'
+                             'appear to be items without these metadata'
+                             'attributes'.format(translation_id))
+
+    for id_vals, items in groupby(content_list, attrgetter(*translation_id)):
+        items = list(items)
+        with_str = 'with' + ', '.join([' {} "{{}}"'] * len(translation_id))\
+            .format(*translation_id).format(*id_vals)
+        original_items = get_original_items(items, with_str)
+        index.extend(original_items)
+        for a in items:
+            a.translations = [x for x in items if x != a]
+
+    translations = [x for x in content_list if x not in index]
+
+    return index, translations
+
+
+def get_original_items(items, with_str):
     def _warn_source_paths(msg, items, *extra):
         args = [len(items)]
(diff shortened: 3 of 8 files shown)


oulenz:
I've expanded the existing test to cover the added functionality.

justinmayer:
Would @getpelican/reviewers be so kind as to review this pull request and either approve it or provide feedback/comments?

justinmayer:
@smartass101 / @jorgesumle: Any comments on this, since it relates to the i18n_subsites plugin?

smartass101:
I currently don't have a test env available, but whoever has  can run the `i18n_subsites` test suite.

pelican/generators.py
```diff
@@ -556,7 +553,9 @@ def generate_context(self):
             self.add_source_path(article)
 
         self.articles, self.translations = process_translations(
-            all_articles,
+            all_articles, 'category')
+        self.articles = order_content(
+            self.articles,
             order_by=self.settings['ARTICLE_ORDER_BY'])
         self.drafts, self.drafts_translations = \
             process_translations(all_drafts)
```
avaris: You forgot to order `drafts` (or `hidden_pages` below).

pelican/utils.py
```diff
@@ -622,79 +629,117 @@ def process_translations(content_list, order_by=None):
...
     grouped_by_slugs = groupby(content_list, attrgetter('slug'))
     index = []
-    translations = []
-
-    def _warn_source_paths(msg, items, *extra):
-        args = [len(items)]
-        args.extend(extra)
-        args.extend((x.source_path for x in items))
-        logger.warning('{}: {}'.format(msg, '\n%s' * len(items)), *args)
+    warnings = []
```
avaris: Why are you storing (a.k.a. delaying) warnings?

pelican/utils.py
```diff
@@ -613,7 +613,14 @@ def escape_html(text, quote=True):
     return escape(text, quote=quote)
 
 
-def process_translations(content_list, order_by=None):
+def _get_warning_with_source_paths(msg, items, *extra):
+    args = [len(items)]
+    args.extend(extra)
+    args.extend((x.source_path for x in items))
+    return '{}: {}'.format(msg, '\n{}' * len(items)).format(*args)
```
avaris: Nope! We are using `'%s'` warnings for a reason: Giving logger the ability to do special stuff if needed.

pelican/utils.py
```diff
@@ -622,79 +629,117 @@ def process_translations(content_list, order_by=None):
...
+                lang_items,
+                with_str,
+                lang))
+
+    # items with `translation` metadata will be used as translations...
+    candidate_items = list(filter(
+        lambda i:
+        i.metadata.get('translation', 'false').lower() == 'false',
+        items))
+
```
avaris: Since you're rewriting stuff, can we replace `list(filter(...))` with proper list comprehension?

pelican/utils.py
```diff
@@ -622,79 +629,117 @@ def process_translations(content_list, order_by=None):
...
+            'All items ({}) {} are translations',
+            items,
+            with_str))
+        candidate_items = items
+
+    # find items with default language
+    original_items = list(filter(
+        attrgetter('in_default_lang'),
+        candidate_items))
+
```
avaris: Again, `list(filter(...))` :disappointed:.

justinmayer:
Hi @oulenz. Would you be so kind as to review the feedback here and make any changes you feel are warranted?

oulenz:
Thanks for the review @avaris! I have been travelling, I hope to have a chance to look at it over the weekend.

pelican/utils.py
oulenz: I had to separate generating and raising the warnings because the warnings generated in the function call in line 653 only have to be raised if the subsequent if-clause tests false. At that point, storing them all until the end seemed most elegant. However, I can see why not issuing these warnings as soon as possible is undesirable, so I'll change that.

pelican/utils.py
oulenz: Sure

pelican/utils.py
oulenz: Ok, will change this back

pelican/generators.py
avaris: Yes they were. `process_translations` did that.

pelican/utils.py
avaris: Oh, I see. I missed that. Looking at the code again, it should be possible to do it in one pass, no?

```python
if secondary_attr:
    # groupby lang and secondary_attr
    groupfn = lambda item: (getattr(item, 'lang'), getattr(item, secondary_attr))
else:
    # groupby lang
    groupfn = attrgetter('lang')

original_items = get_original_items(items, groupfn, ...)
```

pelican/generators.py
oulenz: `process_translations` only did that if the `order_by` parametre was true. For drafts and hidden pages, `order_by` defaulted to None.

Actually, I have to apologise for the presentation here. This is part of the first commit, which has already been merged as a separate PR (which at the time you ok-ed). I am sorry for letting you look at this twice.

avaris: Well, scratch that. I'm not good at reading tonight. For some reason, I read as if this commit removed the first sort (by `slug`), which is not the case. So all good here :).

avaris: Well, thinking about it again, even though it wasn't like that previously, wouldn't it make sense to *order* `drafts` and `hidden_pages` same as `articles` or `pages`?

pelican/utils.py
oulenz: If I understand your suggestion correctly, that would imply a strict approach: translations are only identified if they have the same category as the original article. I was wondering whether there might not be use cases where this isn't true. The rest of `process_translations` seems to be rather tolerant in what it allows, so I tried to stick with that, and only use the secondary attribute if there is ambiguity.

pelican/generators.py
oulenz: I was wondering about that as well (it took me ages to recall why I wrote it like this). I guess they aren't sorted now because they don't feature on the `index` page and on the various category, tag and author pages? But I guess nothing stops people from creating a page with a list of drafts, and then they would want those sorted.

pelican/utils.py
avaris: Yes, but that's what this fix intends to do, right?

If translations can have different category compared to the original, there are some issues :)

Consider this with `default_lang='en'`
```
article1:
 slug: slug
 lang: en
 category: cat

article2 (translation of article1):
 slug: slug
 lang: fr
 category: translation_of_cat

article3 (completely different one):
 slug: slug
 lang: en
 category: some_other_cat
```

How do you match `article2` as translation of `article1`?

oulenz: My idea was the following: if you only have `article1` and `article2`, then `article1` is identified as the original, and so we accept `article2` as a translation, even though the categories mismatch. If you add `article3`, then both `article1` and `article3` are identified as originals, in which case if you have any translations they should have a matching category, otherwise there's no way to select the right original. In this case, `article2` has a different category and so ends up identified as an original article in its own right, even though it isn't in the default language.

(One could write a plugin to use a different secondary attribute to disambiguate with, or we could even add that as a setting.)

avaris: I think we need a more generalized solution. 
The problem: there should be an `id`-like attribute that is common among translations. 

So, how about a new setting `TRANSLATION_ID` that's a function which would generate that `id` from a content to aggregate items by. By default, it would be `attrgetter('slug')` (current behavior).

If you want that to be `(slug+category)`? Fine, define the function that does that. Or if you want to add a completely separate metadata for this purpose, no problem, define that in articles and set the config appropriately.

PS: for convenience, we can automatically wrap the strings for that setting with `attrgetter`. So that you don't need to do `TRANSLATION_ID = attrgetter('id')` but only `TRANSLATION_ID = 'id'`.

oulenz: I agree, that seems most sensible. And let users set it to `None` or `[]` if they simply want to skip looking for translations.  I think a collection of attributes suffices for the setting. I'm not sure there is a use case for custom functions, and if there is, users could create an appropriate metadata attribute dynamically, right?

avaris: Dynamically? With a plugin, yes.

A list of attributes sounds fine. I suggested function for flexibility but you're probably right. It's unlikely you need anything besides attributes.

oulenz: Hm, I was thinking of `PATH_METADATA` and `EXTRA_PATH_METADATA`, but I realise now that these aren't so flexible. May have to add a setting to create attributes from other attributes, but that's for another PR.

pelican/utils.py
```diff
@@ -659,79 +659,86 @@ def escape_html(text, quote=True):
...
-
-        # display warning if there are several original items
-        if len(original_items) > 1:
-            _warn_source_paths(
-                'There are %s original (not translated) items with slug "%s"',
-                original_items,
-                slug)
+    index = []
 
+    content_list.sort(key=attrgetter(*translate_per))
```
avaris: Maybe a `try/except` to provide better error messages in case you pass bad attribute names.

docs/settings.rst
```diff
@@ -1044,6 +1044,18 @@ more information.
...
+   The metadata attribute(s) used to identify which articles are translations
+   of one another. May be a string or a collection of strings. Set to ``None``
+   or ``False`` to disable the identification of translations.
+
+.. data:: PAGE_TRANSLATE_PER = {'slug'}
+
+   The metadata attribute(s) used to identify which pages are translations
+   of one another. May be a string or a collection of strings. Set to ``None``
+   or ``False`` to disable the identification of translations.
+
```
avaris: Even though I get the intention, `set`s aren't the most known collections in python. I'm a little hesitant to expose them to the user like that. A `list` or a `tuple` would be more intuitive I think. Or defaults can be simply
```
..._PER = 'slug'
```

oulenz:
@smartass101 the i18n_subsites plugin is incompatible with the current master of Pelican as long as https://github.com/getpelican/pelican-plugins/pull/953 isn't resolved. Therefore I cannot run its tests. But as far as I can tell, the i18n_subsites plugin doesn't override the identification of translations, and so should be compatible with this PR.

pelican/utils.py
avaris: Messages are good, but you need to raise exceptions (with those messages) instead of logging them. This isn't really a recoverable error, pelican should stop. My intent was to provide more relevant error messages instead of the generic python errors.

```python
    except TypeError:
         raise TypeError('Cannot unpack %s, \'translation_id\' must be falsy, a'
                      'string or a collection of strings' % translation_id)
     except AttributeError:
         raise AttributeError('Cannot use %s as \'translation_id\', there appear to be'
                      'items without these metadata attributes' % translation_id)
```

oulenz: Makes sense, I've changed it to reraise

avaris: Almost. Unlike logging, this time you're supposed to do the formatting :)

```python
raise TypeError('... %s ...' %s foo)
# or
raise TypeError('... {} ...'.format(foo))
```

Otherwise you're passing a tuple.

oulenz: Doh!

justinmayer:
Awesome! 👏  Many thanks, @oulenz and @avaris.

oulenz:
@avaris Yeah thanks for the thorough review and for suggesting the best way to implement this! :)

PR #2312, opened 2018-03-26
Replaces develop_server.sh with .bat script. Fixes #2308

### LINKED ISSUE #2308: pelican-quickstart - develop_server script for windows
When using pelican-quickstart on Windows and answering Y for the question "Do you want an auto-reload HTTP script to assist with theme and site development?" the file develop_server.sh is created although you can't really use it on a Windows system.

Perhaps a good approach would be to not even show the "develop" question if os.name == 'nt' (or platform.system() == 'Windows')

Another suggestion is that a develop_server.bat script is added with appropriate content if the system is Windows. For starting the server it could for example be this simple:

develop_server.bat:
pushd output
python -m pelican.server
popd

daverstam:
As stated in issue #2308 Windows users have no usage of the develop_server.sh script being created when running pelican-quickstart, this fix should address that.

develop_server.bat will run the pelican server and spawn a new window running the pelican command that deploys the content folder.

This has been tested and verified on Windows 10 running Python 3.6.4 and 2.7.14
Tests has also been done on Debian 8 (jessie) running Python 3.4.2 and 2.7.9 to make sure other functionality was not broken.

### DIFF
--- a/pelican/tools/pelican_quickstart.py
+++ b/pelican/tools/pelican_quickstart.py
@@ -386,22 +386,34 @@ needed by Pelican.
             if isinstance(value, six.string_types) and ' ' in value:
                 value = '"' + value.replace('"', '\\"') + '"'
             conf_shell[key] = value
-        try:
-            with codecs.open(os.path.join(CONF['basedir'],
+        if sys.platform == 'win32':
+            try:
+                with codecs.open(os.path.join(CONF['basedir'],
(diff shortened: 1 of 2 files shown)


justinmayer:
@daverstam: It seems #2351, which will probably be merged soon, may obviate the need for a separate .bat script for Windows. Would you have a look and let us know what you think?

PR #2313, opened 2018-03-27
Use gitignore.io to extend .gitignore with common IDE / OS ignores

gwax:
The existing .gitignore misses many common ignore patterns for IDE/OS combinations used in Python development.

I have pulled a fairly extensive Python ignore from https://gitignore.io to avoid committing anything from my dev environment in a subsequent PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,15 +1,354 @@
-*.egg-info
-.*.swp
-.*.swo
-*.pyc
-.DS_Store
-docs/_build
-docs/fr/_build
-build
-dist
-tags
-.tox
-.coverage
-htmlcov
-six-*.egg/
+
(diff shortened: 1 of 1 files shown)


jorgesumle:
This is overkill. Just add the things you need for your development environment.

mosra:
I would suggest setting up a [global gitignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore) (e.g., in `~/.gitignore_global` on Linux) and putting all the stuff you need for your particular IDE/OS combination there.

In my opinion, project-local `.gitignore` files should be used for ignoring files specific to a particular project, not for IDE/OS-specifics. It's not sustainable to add Emacs-specific ignores to every project that could be ever edited with Emacs, for example.

gwax:
Ok.

PR #2311, opened 2018-03-27
Multiple reST headers yields "could not find information about 'date'"

gwax:
When defining posts using ReStructuredText, if one specified more than one chapter header, pelican issues the error:

```
ERROR: Skipping meta/pelican-fail.rst: could not find information about 'date'
```

This error message is particularly opaque when trying to diagnose and repair the underlying issue.

For further testing / validation, here is a failing and succeeding test case that issue the above error (or not):

# Fail

```restructuredtext
Title Line
==========

:slug: pelican-fail
:date: 2018-01-01 12:13:14+00:00
:tags: meta
:category: meta

Test file

Section Header
==============

Content
```

# Succeed

```restructuredtext
Title Line
==========

:slug: pelican-succeed
:date: 2018-01-01 12:13:14+00:00
:tags: meta
:category: meta

Test file

Section Header
--------------

Content
```

The notable difference is the character used to underline "Section Header"

### DIFF
--- a/pelican/readers.py
+++ b/pelican/readers.py
@@ -203,11 +203,18 @@ class RstReader(BaseReader):
     def __init__(self, *args, **kwargs):
         super(RstReader, self).__init__(*args, **kwargs)
 
-    def _parse_metadata(self, document):
+    def _parse_metadata(self, document, source_path):
         """Return the dict containing document metadata"""
         formatted_fields = self.settings['FORMATTED_FIELDS']
 
         output = {}
+
+        if document.first_child_matching_class(docutils.nodes.title) is None:
+            logger.warning(
+                'Document title missing in file %s: '
+                'Ensure exactly one top level section',
+                source_path)
+
         for docinfo in document.traverse(docutils.nodes.docinfo):
             for element in docinfo.children:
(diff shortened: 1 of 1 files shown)


gwax:
I have added a clarifying warning that will point people in the right direction when they encounter this error.

For purposes of #1397 consider this my official permission to change the license to MIT or BSD for any of my contributions.

pelican/readers.py
```diff
@@ -202,12 +202,20 @@ def __init__(self, *args, **kwargs):
 
     def __init__(self, *args, **kwargs):
         super(RstReader, self).__init__(*args, **kwargs)
+        self._source_path = None
 
     def _parse_metadata(self, document):
```
avaris: Pass `source_path` to `_parse_metadata` instead. It doesn't make sense to store it in the instance. Reader instances aren't per file.

gwax: Done.

gwax:
Updated to address review feedback.

justinmayer:
Many thanks to @gwax for the contribution and to @avaris for reviewing. 🤘

PR #2315, opened 2018-03-28
REBASE: Remove develop_server.sh in favour of pelican serving static files it…

rminderhoud:
…self

Competing static site generators integrate the functionality of regenerating
content and serving it into their main executable. In pelican this
functionality used to be in an external script `develop_server.sh` which
resides in the blog base directory. This has the disadvantage that changes in
pelican can break the `develop_server.sh` scripts which will not automatically
be upgraded together with pelican by package managers. Thus, pelican should
integrate this functionality into its main executable.

To this end, this commit removes `develop_server.sh` and adds three command
line options to the pelican executable:

 * `-l/--listen` starts the HTTP server (`-s/--serve` was already taken)
 * `-p/--port` specifies the port to listen at
 * `-b/--bind` specifies the IP to bind to

`--listen` and `--autoreload` can be used together to achieve the same
effect that other static site generators offer: Serve files via HTTP
while at the same time auto-generating the content.

Since the `develop_server.sh` script was removed, pelican-quickstart looses the
`develop` option.

Since the `develop_server.sh` script was removed, the Makefile looses the
`stopserver` target and the `devserver` target is replaced by running `pelican
-l` in the foreground.

Since pelican now offers the `--listen` option, the fabfile uses that instead
of starting the socketserver itself.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -103,7 +103,6 @@ can optionally add yourself if you plan to create non-chronological content)::
     ├── content
     │   └── (pages)
     ├── output
-    ├── develop_server.sh
     ├── fabfile.py
     ├── Makefile
     ├── pelicanconf.py       # Main settings file
--- a/docs/publish.rst
+++ b/docs/publish.rst
@@ -201,10 +201,7 @@ separate terminal sessions, but you can run both at once via::
     make devserver
 
 The above command will simultaneously run Pelican in regeneration mode as well
-as serve the output at http://localhost:8000. Once you are done testing your
-changes, you should stop the development server via::
-
-    ./develop_server.sh stop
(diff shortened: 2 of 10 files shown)


justinmayer:
Thank you for reworking this PR, Ralph. There have been a couple changes... Would you mind rebasing again on to latest master?

rminderhoud:
Updated

justinmayer:
I'm so sorry, Ralph. I merged #2269 the other day, which makes changes to `pelican.server` that conflict with the changes in this pull request. Would you be so kind as to take another look, rebase on current master (again), and make any changes necessary to support the SSL/TLS feature added in #2269?

Once again, please accept my apologies for the oversight. Thank you for your patience! 😅

Issue #2279, opened 2018-01-23
List of available metadata keywords


gDelgado14:
Is there a list somewhere containing the set of allowed metadata keys? (Title, Date, Tags, etc ..)??

voidplayer:
Probably is documented somewhere. But what this is what I did
* Use a non-existent metadata key
* In debug mode, pelican will crash and raise the exception
* Capture with debug program
* Examine existing variables (I dont remember the name, but theres a dict used to make the substitutions with all the valid metadata values)
* ???
* profit

gDelgado14:
Ouu, so it looks like the snippet that @avaris shared is the set of reserved names.

I'm super down to add a section to the docs if it's deemed helpful!

justinmayer:
@gDelgado14: Please feel free to submit a PR that would add this to the docs. Thanks!

justinmayer:
Addressed by above-referenced PR.

PR #2321, opened 2018-04-02
List reserved metadata keywords

gDelgado14:
Addresses https://github.com/getpelican/pelican/issues/2279


avaris:
`template`, `save_as` and `url` are also used by pelican for providing overrides.

justinmayer:
Many thanks to Giorgio for the contribution and to @avaris for the additions.

PR #2324, opened 2018-04-10
correct suffix order in ComplexHTTPRequestHandler

oulenz:
https://github.com/getpelican/pelican/pull/1040 introduced functionality to the dev server to resolve a `'foo/bar'` get request to `'foo/bar.html'` or `'foo/bar/index.html'`, but it is not working as intended because when  `os.path.exists(path + '/index.html')`, then also `os.path.exists(path + '')`. The reason why it nonetheless seems to resolve to `'foo/bar/index.html'` is because `SimpleHTTPRequestHandler.send_head` goes looking for index files if the path in the get request is a directory.

https://github.com/getpelican/pelican/pull/1441 added functionality to strip a trailing `'/'` before applying the suffixes, but it did this by adding a second for loop, for a total of six combination, whereas one more pass in the original for loop would have been sufficient.

This fixes both these issues.

In fact, it is not strictly necessary to include `'/'` in `SUFFIXES`, but this avoids a permanent redirect from `'foo/bar'` to `'foo/bar/'` by `SimpleHTTPRequestHandler.send_head` which would be permanently stored in the browser cache and is non-trivial to undo.

### DIFF
--- a/pelican/server.py
+++ b/pelican/server.py
@@ -42,8 +42,7 @@ def parse_arguments():
 
 
 class ComplexHTTPRequestHandler(srvmod.SimpleHTTPRequestHandler):
-    SUFFIXES = ['', '.html', '/index.html']
-    RSTRIP_PATTERNS = ['', '/']
+    SUFFIXES = ['.html', '/index.html', '/', '']
 
     def translate_path(self, path):
         # abandon query parameters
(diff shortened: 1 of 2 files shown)


oulenz:
@avaris @eliben @ssbarnea

oulenz:
@eliben The motivation for this PR is the first issue. I agree that there is limited gain in removing the second for loop, I only changed it since I was writing this PR anyway. (The motivation here isn't performance, but imho simpler logic --- although I see that we disagree on this.)

oulenz:
I really I didn't spell out the practical problem that this solves: without this, `foo/bar` isn't resolved to `foo/bar.html` if `foo/bar` is a directory.

oulenz:
I've added a test.

I wasn't sure how far to go down the server rabbit hole, so for the moment this only tests the path resolving, but there's also some declarations that aren't strictly necessary but which might be built upon if we want to test the server more thoroughly.

I didn't use the mock library because I haven't used it before and couldn't find a use for it here.

oulenz:
Squashed & rebased

justinmayer:
Humble request for assistance from @getpelican/reviewers... Would you be so kind as to take a moment and review this pull request?

pelican/server.py
```diff
@@ -67,34 +66,28 @@ def translate_path(self, path):
...
+        self.path = self.get_path_that_exists(original_path)
+
+        if not self.path:
             logging.warning("Unable to find `%s` or variations.",
-                            self.original_path)
+                            original_path)
+            return
+
+        logging.info("Found `%s`.", self.path)
+        srvmod.SimpleHTTPRequestHandler.do_GET(self)
```
avaris: I've never noticed this before... What's the point of this compared to `super(...).do_GET()`?

avaris: PS: I know, you didn't write that. That code just looks weird to me :).

justinmayer:
Many thanks to @oulenz for the enhancement and to @eliben and @avaris for reviewing! 🚀

pelican/server.py
oulenz: When was `super` introduced? Maybe it's from before that?

avaris: `super` was added way, waaay before this. It came with new style classes, so around python 2.2 or 2.3 (i think?).

Issue #2316, opened 2018-03-28
More flexible implementation of slug substitutions


oulenz:
My use case is that I want to preserve `'+'` in slugs. With respect to the setting `SLUG_SUBSTITUTIONS`, the docs say:

> `skip` is a boolean indicating whether or not to skip replacement of non-alphanumeric characters. Useful for backward compatibility with existing URLs.

I therefore expected that setting `SLUG_SUBSTITUTIONS = ((' ', '-', True, ), )` would leave slugs unchanged (except for replacing spaces by dashes), and in particular, preserve instances of `'+'`.

But I discovered to my surprise that  `skip` only takes effect if the slug was changed by any of the substitutions provided by the user.:

https://github.com/getpelican/pelican/blob/81b7e7481f78c0fe0a2ff1af169525e225df6286/pelican/utils.py#L304

What is the rationale behind this?

I would like to propose that a new setting be introduced that can be set with a list of regex replacement pairs, with the default being the current removal of non-alphanumerical characters. If you agree, I could write a PR.

jonathan-s:
What is the reason you would like to preserve `+` in slugs? I'm curious about the underlying purpose.

oulenz:
I have two subcategories named simply `+` and `++`, and I want that to be part of their urls. As far as I can tell `'+'` is url-safe.

rlafuente:
Seconding this problem; I'm trying to preserve period characters (`.`) in my slugs, and the only way is to add a filler character to the category names so that they can be replaced and fill the non-matching condition. #2330 seems to be a useful fix to avoid clunky hacks.

PR #2326, opened 2018-04-11
control slug substitutions from settings with regex

oulenz:
Fixes https://github.com/getpelican/pelican/issues/2316

At present, this is not backwards compatible. I welcome advice on how pelican handles deprecation.

### DIFF
--- a/pelican/contents.py
+++ b/pelican/contents.py
@@ -98,14 +98,16 @@ class Content(object):
         if not hasattr(self, 'slug'):
             if (settings['SLUGIFY_SOURCE'] == 'title' and
                     hasattr(self, 'title')):
-                self.slug = slugify(self.title,
-                                    settings.get('SLUG_SUBSTITUTIONS', ()))
+                self.slug = slugify(
+                    self.title,
+                    regex_subs=settings.get('SLUG_REGEX_SUBSTITUTIONS', []))
             elif (settings['SLUGIFY_SOURCE'] == 'basename' and
                     source_path is not None):
                 basename = os.path.basename(
                     os.path.splitext(source_path)[0])
                 self.slug = slugify(
-                    basename, settings.get('SLUG_SUBSTITUTIONS', ()))
+                    basename,
+                    regex_subs=settings.get('SLUG_REGEX_SUBSTITUTIONS', []))
 
         self.source_path = source_path
 
--- a/pelican/settings.py
+++ b/pelican/settings.py
@@ -6,6 +6,7 @@ import inspect
 import locale
 import logging
 import os
+import re
 from os.path import isabs
 from posixpath import join as posix_join
 
@@ -145,7 +146,12 @@ DEFAULT_CONFIG = {
     'TEMPLATE_PAGES': {},
     'TEMPLATE_EXTENSIONS': ['.html'],
     'IGNORE_FILES': ['.#*'],
-    'SLUG_SUBSTITUTIONS': (),
+    'SLUG_REGEX_SUBSTITUTIONS': [
+        (r'[^\w\s-]', ''),  # remove non-alphabetical/whitespace/'-' chars
+        (r'(?u)\A\s*', ''),  # strip leading whitespace
+        (r'(?u)\s*\Z', ''),  # strip trailing whitespace
+        (r'[-\s]+', '-'),  # reduce multiple whitespace or '-' to single '-'
+    ],
     'INTRASITE_LINK_REGEX': '[{|](?P<what>.*?)[|}]',
     'SLUGIFY_SOURCE': 'title',
     'CACHE_CONTENT': False,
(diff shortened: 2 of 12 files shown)


oulenz:
Ok, I have added backwards compatibility plus a deprication warning :)

justinmayer:
@oulenz: Looks like there are some test failures related to `TestWordpressXmlImporter` in this PR.

oulenz:
@justinmayer Sorry! Fixed now. (I was lulled into a false sense of security by these tests getting skipped locally because I didn't have pandoc installed.)

pelican/contents.py
```diff
@@ -98,14 +98,18 @@ def __init__(self, content, metadata=None, settings=None,
         if not hasattr(self, 'slug'):
             if (settings['SLUGIFY_SOURCE'] == 'title' and
                     hasattr(self, 'title')):
-                self.slug = slugify(self.title,
-                                    settings.get('SLUG_SUBSTITUTIONS', ()))
+                self.slug = slugify(
+                    self.title,
+                    regex_subs=settings.get('SLUG_REGEX_SUBSTITUTIONS', []),
+                    substitutions=settings.get('SLUG_SUBSTITUTIONS'))
```
avaris: This isn't the way we handle deprecations. If a setting is replaced with another and old one can be converted to the new one, old one is transferred to the new one in `settings.py` (see [example](https://github.com/oulenz/pelican/blob/master/pelican/settings.py#L383)). Then, old setting is no longer used in the code.

Here you can easily transfer `SLUG_SUBSTITUTIONS` to the new regex counterpart (using `re.escape` possibly).

oulenz:
@avaris I have followed your suggestion, please have another look.

I agree this is a more principled solution. The reason I didn't choose this originally is because 1) it cannot give 100% backwards compatibility, since in the old set-up, a `skip` flag only affects a string if the string triggers that substitution and 2) there are at least three plugins that directly call `slugify`. I would be happy to write another pull request to fix those.

avaris:
@oulenz I went over briefly and generally looks fine. I'll look in detail this weekend. Just one quick note, the word is "depr**e**cated" :).

I think it's fine to break the `skip` flag. As I see it, that was an obscure (and somewhat awkward) setting. Not sure how much use it saw.

oulenz:
Doh! Fixed. Thanks!

pelican/settings.py
```diff
@@ -156,79 +162,67 @@ def load_source(name, path):
...
-        local_settings = get_settings_from_file(path)
-        # Make the paths relative to the settings file
+        settings = dict(get_settings_from_file(path), **settings)
+
+    if settings:
+        settings = handle_deprecated_settings(settings)
+
+    if path:
+        # Make relative paths absolute
+        def getabs(path, pluginpath):
```
avaris: `pluginpath` not the best variable name? 

Probably,

```python
def getabs(base_path, path):
```

or since `base_path` is always `path`, maybe:
```python
def getabs(path, base_path=path):
```

pelican/settings.py
```diff
@@ -245,6 +239,139 @@ def get_jinja_environment(settings):
     return settings
 
 
+def handle_deprecated_settings(settings):
+    """Converts deprecated settings and issues warnings. Issues an exception
+    if both old and new setting is specified.
+    """
+    settings = copy.deepcopy(settings)
```
avaris: Why copy?

pelican/settings.py
```diff
@@ -156,79 +162,67 @@ def load_source(name, path):
...
 def read_settings(path=None, override=None):
+    settings = override or {}
+
     if path:
-        local_settings = get_settings_from_file(path)
-        # Make the paths relative to the settings file
+        settings = dict(get_settings_from_file(path), **settings)
+
+    if settings:
+        settings = handle_deprecated_settings(settings)
```
avaris: This probably belongs to `configure_settings`.

pelican/settings.py
```diff
@@ -139,7 +140,12 @@ def load_source(name, path):
     'TEMPLATE_PAGES': {},
     'TEMPLATE_EXTENSIONS': ['.html'],
     'IGNORE_FILES': ['.#*'],
-    'SLUG_SUBSTITUTIONS': (),
+    'SLUG_REGEX_SUBSTITUTIONS': [
+        (r'[^\w\s-]', ''),
+        (r'(?u)\A\s*', ''),
+        (r'(?u)\s*\Z', ''),
+        (r'[-\s]+', '-'),
+    ],
```
avaris: Since this is now exposed, it would be nice to add a comment for each line to describe what it's doing. Same with the docs.

pelican/tests/test_importer.py
```diff
@@ -84,10 +85,12 @@ def test_dircat(self):
         with temporary_folder() as temp:
             fnames = list(silent_f2p(test_posts, 'markdown',
                                      temp, dircat=True))
+        settings = read_settings()
+        subs = settings['SLUG_REGEX_SUBSTITUTIONS']
```
avaris: You don't need to do `read_settings()` if you don't use the full settings. This could easily be:
```python
subs = DEFAULT_CONFIG['SLUG_REGEX_SUBSTITUTIONS']
```

[Same goes for all the other similar cases below]

pelican/tests/test_settings.py
```diff
@@ -182,4 +183,73 @@ def test_theme_and_extra_templates_exception(self):
...
+        self.assertRaises(Exception, handle_deprecated_settings, settings)
+
+    def test_deprecated_slug_substitutions(self):
+        default_slug_regex_subs = self.settings['SLUG_REGEX_SUBSTITUTIONS']
+        settings = {}
+        settings = handle_deprecated_settings(settings)
+        self.assertNotIn('SLUG_REGEX_SUBSTITUTIONS', settings)
+        self.assertNotIn('TAG_REGEX_SUBSTITUTIONS', settings)
+        self.assertNotIn('CATEGORY_REGEX_SUBSTITUTIONS', settings)
+        self.assertNotIn('AUTHOR_REGEX_SUBSTITUTIONS', settings)
```
avaris: Not sure what's tested here. "`handle_deprecated_settings` doesn't pollute"?

pelican/urlwrappers.py
```diff
@@ -115,10 +116,11 @@ class Category(URLWrapper):
...
         if self._slug is None:
-            substitutions = self.settings.get('SLUG_SUBSTITUTIONS', ())
-            substitutions += tuple(self.settings.get('CATEGORY_SUBSTITUTIONS',
-                                                     ()))
-            self._slug = slugify(self.name, substitutions)
+            if 'CATEGORY_REGEX_SUBSTITUTIONS' in self.settings:
+                subs = self.settings['CATEGORY_REGEX_SUBSTITUTIONS']
+            else:
+                subs = self.settings.get('SLUG_REGEX_SUBSTITUTIONS', [])
+            self._slug = slugify(self.name, regex_subs=subs)
```
avaris: There is a behavior change here. Before, `CATEGORY` was added on top of `SLUG`. Now, it isn't. Is this intentional?

[Same for `TAG` and `AUTHOR`]

pelican/settings.py
oulenz: Yes, good idea

pelican/settings.py
oulenz: There was already some deprecation handling here, directly in `read_settings`. This unites all deprecation handling in a dedicated method, which I feel improves maintainability. Handling of deprecated settings that involve paths needs to occur before making relative paths absolute, so by extension before `configure_settings`.

pelican/settings.py
oulenz: Agree on the name, I am thinking `maybe_relative` for `pluginpath`. I don't understand the rest of your comment: `path` is the path of the settings file, `pluginpath` is a relative path in the settings. (This function was already in the code, I just moved it and broadened its use.)

pelican/settings.py
```diff
@@ -156,79 +162,67 @@ def load_source(name, path):
...
+                path_normed = os.path.normpath(path_joined)
+                path_absolute = os.path.abspath(path_normed)
+                return path_absolute
+
         for p in ['PATH', 'OUTPUT_PATH', 'THEME', 'CACHE_PATH']:
-            if p in local_settings and local_settings[p] is not None \
-                    and not isabs(local_settings[p]):
-                absp = os.path.abspath(os.path.normpath(os.path.join(
-                    os.path.dirname(path), local_settings[p])))
+            if settings.get(p) is not None:
```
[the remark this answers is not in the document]
oulenz: `p` might be in the settings but with value `None`. In the master branch, it's:
```
if p in local_settings and local_settings[p] is not None [...]
```
So I just shortened that.

pelican/settings.py
```diff
@@ -156,79 +162,67 @@ def load_source(name, path):
...
 
+        if settings.get('PLUGIN_PATHS') is not None:
             pluginpath_list = [getabs(path, pluginpath)
                                for pluginpath
-                               in local_settings['PLUGIN_PATHS']]
-            local_settings['PLUGIN_PATHS'] = pluginpath_list
-    else:
-        local_settings = copy.deepcopy(DEFAULT_CONFIG)
+                               in settings['PLUGIN_PATHS']]
+            settings['PLUGIN_PATHS'] = pluginpath_list
```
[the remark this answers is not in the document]
oulenz: That is unchanged from the master branch. I thought about changing it but ended up with the same number of lines, so left it. But I agree not needlessly initialising a variable is a good reason to change it, will fix.

pelican/settings.py
oulenz: So the original dictionary isn't affected. Mostly out of principle.

pelican/tests/test_importer.py
oulenz: True, will change.

pelican/tests/test_settings.py
oulenz: Yes, in particular because `DEFAULT_CONFIG` can only take effect if these are not already present in the settings.

pelican/urlwrappers.py
oulenz: Yes, this seems simpler to me, especially if for some reason you don't want some of the substitutions in `SLUG_SUBSTITUTIONS` to take effect for categories. Note that `AUTHOR` behaved differently previously, I also thought it simpler to bring it in line.

pelican/settings.py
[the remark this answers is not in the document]
avaris: Yeah, but I don't think `None` is a valid option for any of these settings. I'll double check but it'll error anyways.

pelican/settings.py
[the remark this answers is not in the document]
avaris: Yeah, I know it was like that before. Since you're doing an overhaul, might as well fix the old oddities :).

pelican/settings.py
avaris: `getabs` is used only in two places and always with `path` as the initial argument. So, I'm suggesting:
```python
        def getabs(path, settings_path=path):
            if isabs(path):
                return path
            else:
                path_dirname = os.path.dirname(settings_path)
                path_joined = os.path.join(path_dirname, path)
                path_normed = os.path.normpath(path_joined)
                path_absolute = os.path.abspath(path_normed)
                return path_absolute
```
and use it like:
```python
getabs(settings[p])
```

FWIW, second part could be:
```python
                return os.path.abspath(
                           os.path.normpath(
                           os.path.join(
                               os.path.dirname(settings_path), 
                               path)))
```
but I don't mind the more verbose one.

pelican/settings.py
avaris: Oh, I definitely agree with moving deprecation to its own function. No argument there. I guess it's fine. I was just suggesting moving the function call in `configure_settings`, but yeah, to do that `make absolute` bit also need to be moved I guess. I guess this is fine :).

pelican/settings.py
avaris: Understandable, but meh... `configure_settings` does similar stuff without copy :). It's just unnecessary. We don't really care what happens to the original.

pelican/urlwrappers.py
avaris: Agreed. Just making sure it was intended :).

PS: Special handling of `AUTHOR` was annoying me as well.

pelican/tests/test_settings.py
avaris: Fair enough, then in that regard, can I request some comment explaining each "test group" in this function?

avaris: Something like:
```
# If no deprecated setting was there, don't produce new ones
....
# Transform SLUG_SUBSTITUTIONS, shouldn't create {TAG, CATEGORY}_REGEX_SUBSTITUTIONS
...
# etc
```

oulenz:
@avaris Not sure whether you were finished, but I've gone ahead and tried to fix the things you pointed out.

oulenz:
@avaris does this look ok now?

justinmayer:

justinmayer:
@avaris: Any last comments on this before it is merged?

oulenz:
Squashed & rebased

justinmayer:
Many thanks to @oulenz for the hard work and patience, and to @avaris for the detailed review and suggestions! ✨

justinmayer:
Looks like I spoke too soon. In an effort to save @oulenz from the trouble of having to resolve a merge conflict, I attempted to resolve it via the GitHub interface but perhaps botched it, potentially causing the test failure referenced below. (Perhaps I should avoid doing this in the future. No good deed goes unpunished! 😝 )

@oulenz: Would you be so kind as to have a look at the test failure and see what's going on?

oulenz:
Already on it :-)

justinmayer:
Yay! Thanks again, @oulenz and @avaris.

PR #2328, opened 2018-04-13
Add ARTICLE_RESTRICT_TO_TAGS setting

inducer:
This allows restricting the set of articles considered for publishing with the help of a set of tags.This is useful, for example, for generating multiple versions of a site with overlapping content--e.g. a professional and a personal one.

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -179,6 +179,12 @@ Basic settings
    A list of directories to exclude when looking for articles in addition to
    ``PAGE_PATHS``.
 
+.. data:: ARTICLE_RESTRICT_TO_TAGS = []
+
+    All articles that do not have all tags in this list will be excluded from
+    the output. This is useful, for example, for generating multiple versions
+    of a site with overlapping content--e.g. a professional and a personal one.
+
 .. data:: OUTPUT_SOURCES = False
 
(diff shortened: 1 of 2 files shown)


jorgesumle:
I didn't understand your idea. Can you explain a use case as an example? Do you think this can be done with a plugin?

inducer:
Sure. I have two sites:

* https://andreask.cs.illinois.edu/
* https://mathema.tician.de/

They're generated off the same code (by two different config files). The 'professional' version only contains articles with a certain tag to keep overly-personal things out of my professional web presence.

I don't think this can be done as a plugin (without monkeypatching) because it modifies core article enumeration.

avaris:
One way to achieve this: place your articles into folders `personal`, `professional` and `common`. Then in your config files

```python
# personal one
ARTICLE_PATHS = ['personal', 'common']

# professional one
ARTICLE_PATHS = ['professional', 'common']
```

And, this can be done with a plugin quite easily. Connect to [`article_generator_pretaxonomy`](https://github.com/getpelican/pelican/blob/master/pelican/generators.py#L608) signal and remove articles from `generator.articles`, `generator.translations`, etc as you wish.

inducer:
Ok, fair. Thanks for taking a look.

PR #2330, opened 2018-04-17
Retain non-alphanumeric characters in slug

jnbrex:
Currently, users are unable to retain non-alphanumeric characters in their URLs unless a substitution specified in their SLUG_SUBSTITUTIONS setting effects a change in their URL. With this change users will be able to use non-alphanumeric characters without having to jump through an extra hoop. This may require a rework (placement into a different setting?) because it will break backwards compatibility for some users who rely on this behavior.

### DIFF
--- a/pelican/utils.py
+++ b/pelican/utils.py
@@ -299,10 +299,8 @@ def slugify(value, substitutions=()):
     for src, dst, skip in substitutions:
         orig_value = value
         value = value.replace(src.lower(), dst.lower())
-        # if replacement was made then skip non-alphanum
-        # replacement if instructed to do so
-        if value != orig_value:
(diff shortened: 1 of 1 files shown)


justinmayer:
Hi Jason. It seems Travis is reporting some test failures in this pull request. Would you care to investigate the cause so we can consider merging your contribution?

oulenz:
I think this is rendered moot by https://github.com/getpelican/pelican/pull/2326

justinmayer:
Thanks, @oulenz.

@jnbrex: Thank you for the contribution. It seems the original problem may now be resolved in master. If you feel there is further work to be done related to this issue, please feel free to comment here so we can consider re-opening.

PR #2344, opened 2018-05-16
add more information to contribution guidelines

oulenz:
I've added some changes to the contribution guidelines that address the 'friction' I experienced when I filed my first pull requests to Pelican, i.e. things that seem logical now that I know, but which could have been clearer.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -114,15 +114,21 @@ Using Git and GitHub
   `install hub <https://github.com/github/hub/#installation>`_ and then run
   `hub pull-request <https://github.com/github/hub/#git-pull-request>`_ to
   turn your GitHub issue into a pull request containing your code.
+* After you have issued a pull request, Jenkins will run the test suite for all
+  supported Python versions and check for PEP8 compliance. If any of these
+  checks fail, you should fix them. (If the tests seem to pass locally, some
+  tests may be getting skipped.)
 
 Contribution quality standards
 ------------------------------
 
-* Adhere to `PEP8 coding standards`_ whenever possible. This can be eased via
-  the `pep8 <http://pypi.python.org/pypi/pep8>`_ or `flake8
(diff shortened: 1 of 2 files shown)


jorgesumle:
Probably, we want to update the reference to pep8. See the new package [pycodestyle (formerly called pep8) - Python style guide checker](https://pypi.org/project/pycodestyle/). It was renamed, as you can see in the latest release note (1.7.1 (2017-10-22)):

> Prominently note via warning message that the tool is no longer released as pep8 and will only be fixed in the pycodestyle package

> If you are using PyCharm and are relying on its warnings for PEP8 compliance, note that you have to decrease the line length to 79, since its default is 120.

Notes about particular text editors shouldn't be added, in my opinion. Otherwise, we would have tons of recommendations which will make the text longer to read and less relevant to people who don't use that particular editor.

avaris:
I agree with @jorgesumle about editor-specific notes.

And ` Testing on Python 2 and 3` section seems... odd. Not your change. That whole section. I think it's somewhat outdated :). It should be basically "After making changes run tests on all supported python versions or use `tox`."

PS: Personally, I don't recommend `conda` but, meh, sure.

oulenz:
pep8 -> pycodestyle: changed

PyCharm: made the comment more general, removing specific mention of PyCharm.

Conda: @avaris I don't want to put words in your mouth. :-) Is there something specific about Conda that you would say makes it less suitable for Pelican development?

Python 2/3: I've rewritten that section, please have a look.

oulenz:
@avaris Are you satisfied with the changes so this can be merged?

oulenz:
Thanks, I've squashed the commits so this can be merged.

justinmayer:
Merged via 8ef9d11b. Many thanks to Oliver for the improved documentation and to @ingwinlu, @jorgesumle, and @avaris for reviewing. Much appreciated!

PR #2352, opened 2018-06-13
Clean simple theme

Shir0kamii:
Hi!

I just made my own theme based on the `simple` one. I found that a lot of things on some pages are not present on others (like a better title block) while they would be better imo.

So I tried to make the `simple` theme more coherent as a whole, and made minor tweaks to improve some things. I made very small commits so you can filter what you want to keep if not everything.

Please tell me if you think of anything else!

### DIFF
--- a/pelican/tests/output/basic/categories.html
+++ b/pelican/tests/output/basic/categories.html
@@ -2,7 +2,7 @@
 <html lang="en">
 <head>
         <meta charset="utf-8" />
-        <title>A Pelican Blog</title>
+        <title>A Pelican Blog - Categories</title>
         <link rel="stylesheet" href="/theme/css/main.css" />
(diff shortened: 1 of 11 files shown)


jorgesumle:
The theme uses space characters, don't use tabs. Why did you replace Tags with Categories in `tags.html`?

Shir0kamii:
Sorry, I switched `tags.html` and `categories.html` titles. It's fixed. Same thing for spaces instead of tabs, I replaced them all over the theme (to complete the cleaning).

avaris:
Looks good once commits are squashed.

Aside: I'm actually surprised that these changes don't affect more of the test outputs...

Shir0kamii:
@avaris This is done!

Indeed, I think not all parts of the theme output are tested.

justinmayer:
Many thanks to @Shir0kamii for the contribution and to @avaris and @jorgesumle for reviewing. 🚀

PR #2353, opened 2018-06-13
Adding support for ctags generation through GENERATE_CTAGS setting

Lucas-C:
The goal is to provide tags autocompletion in all editors that support ctags (e.g. `vim` natively).

Here is an example of simple integration of such tags completion in Notepad++ using the "Python Script" plugin:
```
import os
ctags_filepath = os.path.join(os.path.dirname(notepad.getCurrentFilename()), 'tags')
if os.path.exists(ctags_filepath):
    with open(ctags_filepath) as ctags_file:
        ctags = set(line.split('\t')[0] for line in ctags_file.readlines())
    editor.autoCShow(0, ' '.join(sorted(ctags)))
```

When bound to a shortcut, this will provide autocompletion of existing articles tags.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -5,6 +5,7 @@ Next release
 ============
 
 * New signal: ``feed_generated``
+* New setting: ``GENERATE_CTAGS`` to generate a "tags" file following the CTags format and containing all articles tags.
 
 3.7.1 (2017-01-10)
 ==================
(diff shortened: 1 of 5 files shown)


Lucas-C:
I don't understand why the Travis CI build is failing :(

Lucas-C:
Closing this, I made it a plugin instead : https://github.com/getpelican/pelican-plugins/pull/1038

PR #2355, opened 2018-06-16
print info where the server is located.

nerdoc:
Django does this nicely.


nerdoc:
I made the default port namely 8000, as the pelican.server uses that anyway. If you don't specify a port at the CLI, the $port var is empty else, and all the variable uses are wrong. Saw this when I used it in my first commit her in this patch.

PR #2359, opened 2018-06-22
Add missing Wordpress options to importer doc

davidag:
* Add missing options in the import documentation.
* Fix a typo in a help string of pelican-import.
* Change WordPress sample data URL to a more recent one (2016 vs 2008).


justinmayer:
Many thanks to @davidag for the contribution and to @avaris for reviewing.

PR #2365, opened 2018-06-25
Resolve symlinks before comparison test. Fixes #2358

### LINKED ISSUE #2358: invalid_symlink_is_overwritten test fails on macOS
Running `python -Wd -m unittest discover` on current master (80f212cc) on macOS 10.13.5 yields a test failure:
```
FAIL: test_invalid_symlink_is_overwritten (pelican.tests.test_generators.TestStaticGenerator)
  File "/Virtualenvs/pelican-dev/src/pelican/tests/test_generators.py", line 1027, in test_invalid_symlink_is_overwritten
    self.assertEqual(os.path.realpath(self.endfile), self.startfile)
AssertionError: '/private/var/folders/hh/k8ll_mqs7xvcm_6d0k[49 chars]file' != '/var/folders/hh/k8ll_mqs7xvcm_6d0k01xq6h00[41 chars]file'
- /private/var/folders/hh/k8ll_mqs7xvcm_6d0k01xq6h0000gn/T/testcontent.mna2by30/static/staticfile
? --------
+ /var/folders/hh/k8ll_mqs7xvcm_6d0k01xq6h0000gn/T/testcontent.mna2by30/static/staticfile
```    

Perhaps this is caused because `/var` on macOS is a symlink to `/private/var`?

Anyone willing to fix the test so it passes on macOS? (cc: @adeverteuil | Refs #1982)

thomasjpfan:
1. Fixes #2358 by forcing both paths to resolve symlinks before comparing.
2. Adds test to make sure `self.endfile` is a symlink since `fallback_to_symlink=True`.

### DIFF
--- a/pelican/tests/test_generators.py
+++ b/pelican/tests/test_generators.py
@@ -1024,7 +1024,9 @@ class TestStaticGenerator(unittest.TestCase):
         self.generator.fallback_to_symlinks = True
         self.generator.generate_context()
         self.generator.generate_output(None)
-        self.assertEqual(os.path.realpath(self.endfile), self.startfile)
+        self.assertTrue(os.path.islink(self.endfile))
+        self.assertEqual(os.path.realpath(self.endfile),
+                         os.path.realpath(self.startfile))
 
     def test_delete_existing_file_before_mkdir(self):
         with open(self.startfile, "w") as f:


justinmayer:
Many thanks for working on this, Thomas! I haven't had a chance to test it yet, but will do so as soon as I can.

@adeverteuil: Could you have a look and comment here with any thoughts you might have?

justinmayer:
Merged 6e0ed07446695f54406a28987e7cdcc4d5283213 via dc465d5c5f0eca1fd1ec7eca1f4877acebd9321b. Thanks again for the fix, Thomas! 🚀

Issue #2336, opened 2018-04-23
piwik is now called matomo


geniusupgrader:
the renaming should be reflected in the variables and everywhere in pelican

jorgesumle:
Here are the strings where we use the word Piwik (case-insensitive), I used [GNU grep](https://www.gnu.org/software/grep): 

```
$ grep -Ri 'piwik'
pelican/themes/notmyidea/templates/analytics.html:{% if PIWIK_URL and PIWIK_SITE_ID %}
pelican/themes/notmyidea/templates/analytics.html:    {% if PIWIK_SSL_URL %}
pelican/themes/notmyidea/templates/analytics.html:        var pkBaseURL = "{{ PIWIK_SSL_URL }}";
pelican/themes/notmyidea/templates/analytics.html:        var pkBaseURL = "{{ PIWIK_URL }}";
pelican/themes/notmyidea/templates/analytics.html:        _paq.push(["setTrackerUrl", u+"piwik.php"]);
pelican/themes/notmyidea/templates/analytics.html:        _paq.push(["setSiteId", "{{ PIWIK_SITE_ID }}"]);
pelican/themes/notmyidea/templates/analytics.html:        g.defer=true; g.async=true; g.src=u+"piwik.js"; s.parentNode.insertBefore(g,s);
docs/_build/settings.pot:msgid "URL to your Piwik server - without 'http://' at the beginning."
docs/_build/settings.pot:msgid "If the SSL-URL differs from the normal Piwik-URL you have to include this setting too. (optional)"
docs/_build/settings.pot:msgid "ID for the monitored website. You can find the ID in the Piwik admin interface > Settings > Websites."
Coincidencia en el archivo binario docs/_build/.doctrees/environment.pickle
Coincidencia en el archivo binario docs/_build/.doctrees/settings.doctree
docs/settings.rst:.. data:: PIWIK_URL
docs/settings.rst:   URL to your Piwik server - without 'http://' at the beginning.
docs/settings.rst:.. data:: PIWIK_SSL_URL
docs/settings.rst:   If the SSL-URL differs from the normal Piwik-URL you have to include this
docs/settings.rst:.. data:: PIWIK_SITE_ID
docs/settings.rst:   ID for the monitored website. You can find the ID in the Piwik admin
Coincidencia en el archivo binario .git/objects/pack/pack-2bc14f9e82881a87287af788066c0624a27efd68.pack
```

admiralspark:
[...]

PR #2369, opened 2018-06-29
Replaced references to Piwik with Matomo

admiralspark:
SSIA. Reference Issue #2336

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -1165,18 +1165,18 @@ free to use them in your themes as well.
    A list of tuples (Title, URL) for additional menu items to appear at the
    beginning of the main menu.
 
-.. data:: PIWIK_URL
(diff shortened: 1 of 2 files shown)


jorgesumle:
The script files are still called piwik (`piwik.php` and `piwik.js`, lines 39 and 42 from `pelican/themes/notmyidea/templates/analytics.html`) according to the [documentation](https://developer.matomo.org/guides/tracking-javascript-guide) and to the [source code](https://github.com/matomo-org/matomo); if you rename that, it won't work.

justinmayer:
Thanks for your contribution, Kris. Would you care to address the problem Jorge mentioned so we can consider merging your pull request?

admiralspark:
Sorry, I had long since abandoned working on this :)

I'll close this PR as za had done a better job on it than I.

Issue #456, opened 2012-08-05
status:draft for pages


andreasio:
This is very much related to the upcomming 3.0 feature `hidden`. But I think `hidden` and `drafts` mean two different things. This seams to be confirmed here: https://github.com/getpelican/pelican/issues/380

But I find it inlogical that I can not draft a page? So I propose to add `status:draft` for pages as well as for articles.

justinmayer:
This issue was finally addressed by @oulenz. Thank you! 🙏🏻

PR #2371, opened 2018-07-03
allows pages to have draft status, like articles

oulenz:
Fixes https://github.com/getpelican/pelican/issues/456, fixes https://github.com/getpelican/pelican/issues/1620

This appears to be an old issue. There was general agreement in https://github.com/getpelican/pelican/issues/456 that this should be done, but nothing materialised (see pr https://github.com/getpelican/pelican/pull/626).

In addition to the reasons mentioned in the issues, the wordpress importer (unsurprisingly) imports draft pages as pages with `draft` status, causing an error when trying to run pelican on the output. Imho, the wordpress import process should run as smooth as possible for newcomers.


justinmayer:
This has indeed been a long-standing issue. Many thanks to @oulenz for the enhancement and to @avaris for reviewing. 🎉

PR #2373, opened 2018-07-04
Fixing some warnings and errors in the sample content.

marespiaut:
The current sample content is not up-to-date with the current Pelican mechanism.

This will help new comers to understand better how Pelican works.

* More valid articles.
* More translations.
* Images are now correctly displayed.


avaris:
Looks like you need to [regenerate the sample output](http://docs.getpelican.com/en/stable/contribute.html#running-the-test-suite).

PS: For future, it's best to run tests locally before pushing.

marespiaut:
Alright, I didn't knew that. I'll try to fix it.

justinmayer:
Hi @marespiaut. Would you be so kind as to re-generate the sample output and include it in your PR so we can get your contribution merged?

justinmayer:
Thank you for the impetus to fix this, @marespiaut. Very much appreciated! 🙇

PR #2374, opened 2018-07-04
Add hreflang attribute to default themes

jorgesumle:
Related to https://github.com/getpelican/pelican-plugins/issues/726


justinmayer:
Many thanks for the enhancement, Jorge!

Issue #2155, opened 2017-05-23
Category & Tag Slug must be a folder


itsonlybarney:
When I set CATEGORY_URL and TAG_URL to have {slug}.html as the name, the writing of the files fails. I get an error stating that */category/index.html is to be overwritten.

I'm using the examples from the [Pelican Docs](http://docs.getpelican.com/en/stable/settings.html#url-settings).

This is the actual Error message I get:
```
PS C:\Users\user\Downloads\_development\git\itsonlybarney.github.io> pelican
CRITICAL: RuntimeError: File C:\Users\user\Downloads\_development\git\itsonlybarney.github.io\output\blog/category/index.html is to be overwritten
```

The URL settings in my pelicanconf.py file looks as follows:
```
RELATIVE_URLS = True
# Post and Pages path
ARTICLE_URL = 'blog/{slug}.html'
ARTICLE_SAVE_AS = 'blog/{slug}.html'
PAGE_URL = 'pages/{slug}.html'
PAGE_SAVE_AS = 'pages/{slug}.html'
YEAR_ARCHIVE_SAVE_AS = 'blog/{date:%Y}/index.html'
MONTH_ARCHIVE_SAVE_AS = 'blog/{date:%Y}/{date:%m}/index.html'
# Tags and Category path
CATEGORY_URL = 'blog/category/{slug}/'
CATEGORY_SAVE_AS = 'blog/category/{slug}/index.html'
CATEGORIES_SAVE_AS = 'blog/categories.html'
TAG_URL = 'blog/tag/{slug}/'
TAG_SAVE_AS = 'blog/tag/{slug}/index.html'
TAGS_SAVE_AS = 'blog/tags.html'
# Author
AUTHOR_URL = 'author/{slug}.html'
AUTHOR_SAVE_AS = 'author/{slug}.html'
AUTHORS_SAVE_AS = 'authors.html'
# Pagination
DEFAULT_PAGINATION = 3
PAGINATION_PATTERNS = (
    (1, '{base_name}/', '{base_name}/index.html'),
    (2, '{base_name}/page/{number}/', '{base_name}/page/{number}/index.html'),
)
```
I'm not sure if the PAGINATION_PATTERNS command has anything to do with it or not.

bandantonio:
Yeah, the same error. I was unable to make this `{slug}.html` work :(

PR #1956, opened 2016-05-22
Documentation for avoiding Runtime error with pagination_patterns

jonathan-s:
This PR updates the documentation that details how you should avoid getting a `RuntimeError` when using pagination patterns. Closes #1615

PR #2375, opened 2018-07-05
Tweak paginator to accommodate {slug}.html patterns

oulenz:
Fixes https://github.com/getpelican/pelican/issues/1137, fixes https://github.com/getpelican/pelican/issues/1615, fixes https://github.com/getpelican/pelican/issues/2155, fixes .
Makes pr https://github.com/getpelican/pelican/pull/1956 redundant.

At present, pagination produces critical errors with the default `*_SAVE_AS` settings for authors, categories and tags, because the `{base_name}` variable strips `{slug}.html` from `foo/{slug}.html`. This fixes that by ensuring that `{base_name}` only strips `.html` and `/index.html`


justinmayer:

docs/settings.rst
```diff
@@ -1000,33 +1000,39 @@ You can use the following settings to configure the pagination.
...
+By default, pages subsequent to ``.../foo.html`` are created as
+``.../foo2.html``, etc. The ``PAGINATION_PATTERNS`` setting can be used to
+change this. It takes a sequence of triples, where each triple consists of::
 
   (minimum page, URL setting, SAVE_AS setting,)
 
-For example, if you wanted the first page to just be ``/``, and the
-second (and subsequent) pages to be ``/page/2/``, you would set
-``PAGINATION_PATTERNS`` as follows::
+For ``URL setting`` and ``SAVE_AS setting``, you may use a number of variables.
```
avaris:
```
For ``URL`` setting and ``SAVE_AS`` setting ...
```

pelican/paginator.py
```diff
@@ -67,8 +67,11 @@ def _get_page_range(self):
 
 class Page(object):
     def __init__(self, name, url, object_list, number, paginator, settings):
+        self.full_name = name
         self.name, self.extension = os.path.splitext(name)
-        self.base_url = url
+        dirname, filename = os.path.split(name)
+        self.base_name = dirname if filename == 'index.html' else self.name
+        self.full_url = url
```
avaris: Someone might be using `base_url` in a template. No need to break this with no reason.

pelican/paginator.py
```diff
@@ -67,8 +67,11 @@ def _get_page_range(self):
 
 class Page(object):
     def __init__(self, name, url, object_list, number, paginator, settings):
+        self.full_name = name
         self.name, self.extension = os.path.splitext(name)
-        self.base_url = url
+        dirname, filename = os.path.split(name)
+        self.base_name = dirname if filename == 'index.html' else self.name
```
avaris: `index.htm` 🤔

pelican/paginator.py
oulenz: If I parse this method (`_from_settings`) correctly, it's the values in `context` that get inserted into the patterns. I renamed the attribute from `base_url` to `full_url`, but I didn't rename the corresponding value in `context`, which is still `url`. So this shouldn't break anything, right?

docs/settings.rst
oulenz: Thanks, renamed those variables

pelican/paginator.py
oulenz: Thanks, added `index.htm`

pelican/paginator.py
avaris: `Page` object gets passed to the template. Someone might be using `articles_page.base_url` in their templates... for something...

oulenz: Ah ok! I think the `base_` part is misleading, but I guess we're stuck with it. I've reverted to `base_url`.

avaris: Joys of backward compatibility :)

justinmayer:
Sincere thanks to @oulenz for the fix and to @avaris for reviewing! 😁

PR #2376, opened 2018-07-05
Tweaks to simple theme on listing pages

strburst:
This has a few incremental tweaks to the simple template (see commit messages for detailed descriptions). The only content change is in 0385a9b.

I tested this with `python -Wd -m unittest discover`. Also, here is a diff of the output of generating the theme with

```shell
$ git checkout master
$ pelican -s samples/pelican.conf.py -t pelican/themes/simple -o /tmp/master_output samples/content
$ git checkout patch
$ pelican -s samples/pelican.conf.py -t pelican/themes/simple -o /tmp/patch_output samples/content
```

showing that everything is well and good.

```diff
diff -ur /tmp/master_output/authors.html /tmp/patch_output/authors.html
--- /tmp/master_output/authors.html     2018-07-05 10:30:19.553874336 -0400
+++ /tmp/patch_output/authors.html      2018-07-05 10:30:07.479809581 -0400
@@ -17,8 +17,8 @@
             <li><a href="./pages/this-is-a-test-page.html">This is a test page</a></li>
         </ul></nav><!-- /#menu -->
     <h1>Authors on Alexis' log</h1>
-
-    <ul>        <li><a href="./author/alexis-metaireau.html">Alexis Métaireau</a> (10)</li>
+    <ul>
+        <li><a href="./author/alexis-metaireau.html">Alexis Métaireau</a> (10)</li>
     </ul>
         <footer id="contentinfo" class="body">
                 <address id="about" class="vcard body">
diff -ur /tmp/master_output/categories.html /tmp/patch_output/categories.html
--- /tmp/master_output/categories.html  2018-07-05 10:30:19.550874319 -0400
+++ /tmp/patch_output/categories.html   2018-07-05 10:30:07.477809570 -0400
@@ -16,12 +16,13 @@
             <li><a href="./override/">Override url/save_as</a></li>
             <li><a href="./pages/this-is-a-test-page.html">This is a test page</a></li>
         </ul></nav><!-- /#menu -->
-<ul>
-    <li><a href="./category/bar.html">bar</a> (1)</li>
-    <li><a href="./category/cat1.html">cat1</a> (4)</li>
-    <li><a href="./category/misc.html">misc</a> (4)</li>
-    <li><a href="./category/yeah.html">yeah</a> (1)</li>
-</ul>
+    <h1>Categories on Alexis' log</h1>
+    <ul>
+        <li><a href="./category/bar.html">bar</a> (1)</li>
[...]

PR #2377, opened 2018-07-05
Show article category/tags even if no authors (in the simple theme)

strburst:
From the commit message:

> Before this commit, the behavior of the simple theme is to hide the category and tags if there is no author defined, even if the category/tags are present. This appears to be unintentional, due to a misplaced `endif`.
>
> Fix this by moving the endif to the right place, so that the other if blocks aren't nested.

I tested this by diffing on the output `samples/content` and found no diffs. (I used the same commands that I used in #2376.)


justinmayer:
Thanks to @strburst for the fix and to @avaris for reviewing. 👍

PR #2378, opened 2018-07-07
Avoid escaping dollar signs in Makefile template

davidag:
If dollars are escaped, the commands don't work as expected (i.e. doing variable substitution). 

Maybe @Lucas-C can review this as the changes were introduced in his commit #2176.


Lucas-C:
Indeed, this was a bug introduced by my commit.
The fix LGTM

justinmayer:
Thanks for the fix, David!

PR #2379, opened 2018-07-08
Normalize formatting of sample theme Jinja templates

strburst:
## Description

This PR normalizes the formatting of the Jinja templates in the `simple` and `notmyidea` themes, making them easier to read and indentation rules clearer for future contributors. The diff is huge, but the changes are entirely mechanical/whitespace-related. As far as I'm aware, only the following changes are applied.

1. 4-space indentation (there was a small amount of 2-space indentation). No "exceptions" like content inside `<body>` or `<section>`.
1. Conform to the conventions that are used in the examples in the [official Jinja template documentation](http://jinja.pocoo.org/docs/2.10/templates/). I have listed the rules as I understand them in the commit message of 59750b3.
1. Multiline nested tags like
    ```html
    <tag1><tag2>
        ... nested content ...
    </tag2></tag1>
    ```
    are broken up into
    ```html
    <tag1>
        <tag2>
            ... nested content ...
        </tag2>
    </tag1>
    ```

    but nested tags on one line like
    ```html
    <li><a href="...">...</a></li>
    ```

    are untouched/allowed.

Other stuff like CSS is untouched.

To maintain your sanity, I would recommend focusing on just the templates in the themes themselves (grouped together at the bottom of the diffs and also separated into one commit for each theme) and not the html output, then doing a pass with `git diff -w` or checking "Hide whitepsace changes" on the Github diff to ignore whitespace changes (which will only show changes like rule 3).

Note that this doesn't make the output html pretty at all due to the fact that nested Jinja templates are not aware of indentation level at all.

## Testing

I tested this by generating the output for both the simple and notmyidea themes and running the diff tests, then also inspecting the output in a browser to make sure everything looks the same.

## Implementation note

Unfortunately, there doesn't seem to be a good standalone Jinja template formatter (the most recommended option I found when searching was [atom-beautify](https://atom.io/packages/atom-beautify), which is part of Atom, of course). I used the tool [htmlbeautifier](https://github.com/threedaymonk/htmlbeautifier), but had to apply a lot of Jinja-specific rules manually on top.

### DIFF
--- a/pelican/tests/output/basic/a-markdown-powered-article.html
+++ b/pelican/tests/output/basic/a-markdown-powered-article.html
@@ -1,65 +1,68 @@
 <!DOCTYPE html>
 <html lang="en">
-<head>
+    <head>
         <meta charset="utf-8" />
         <title>A markdown powered article</title>
         <link rel="stylesheet" href="/theme/css/main.css" />
         <link href="/feeds/all.atom.xml" type="application/atom+xml" rel="alternate" title="A Pelican Blog Atom Feed" />
-</head>
+    </head>
 
-<body id="index" class="home">
+    <body id="index" class="home">
         <header id="banner" class="body">
-                <h1><a href="/">A Pelican Blog </a></h1>
-                <nav><ul>
-                    <li><a href="/tag/oh.html">Oh Oh Oh</a></li>
-                    <li><a href="/override/">Override url/save_as</a></li>
-                    <li><a href="/pages/this-is-a-test-page.html">This is a test page</a></li>
-                    <li><a href="/category/bar.html">bar</a></li>
-                    <li class="active"><a href="/category/cat1.html">cat1</a></li>
-                    <li><a href="/category/misc.html">misc</a></li>
-                    <li><a href="/category/yeah.html">yeah</a></li>
-                </ul></nav>
+            <h1><a href="/">A Pelican Blog </a></h1>
+            <nav>
+                <ul>
+                            <li><a href="/tag/oh.html">Oh Oh Oh</a></li>
+                            <li><a href="/override/">Override url/save_as</a></li>
+                            <li><a href="/pages/this-is-a-test-page.html">This is a test page</a></li>
+                            <li><a href="/category/bar.html">bar</a></li>
+                            <li class="active"><a href="/category/cat1.html">cat1</a></li>
+                            <li><a href="/category/misc.html">misc</a></li>
+                            <li><a href="/category/yeah.html">yeah</a></li>
+                </ul>
+            </nav>
         </header><!-- /#banner -->
 <section id="content" class="body">
-  <article>
-    <header>
-      <h1 class="entry-title">
-        <a href="/a-markdown-powered-article.html" rel="bookmark"
-           title="Permalink to A markdown powered article">A markdown powered article</a></h1>
-    </header>
+    <article>
+        <header>
+            <h1 class="entry-title">
+                <a href="/a-markdown-powered-article.html" rel="bookmark"
+                   title="Permalink to A markdown powered article">A markdown powered article</a>
+            </h1>
+        </header>
 
-    <div class="entry-content">
+        <div class="entry-content">
 <footer class="post-info">
-        <abbr class="published" title="2011-04-20T00:00:00+00:00">
-                Published: Wed 20 April 2011
-        </abbr>
+    <abbr class="published" title="2011-04-20T00:00:00+00:00">
+        Published: Wed 20 April 2011
+    </abbr>
 
-<p>In <a href="/category/cat1.html">cat1</a>.</p>
+    <p>In <a href="/category/cat1.html">cat1</a>.</p>
 
-</footer><!-- /.post-info -->      <p>You're mutually oblivious.</p>
+</footer><!-- /.post-info -->            <p>You're mutually oblivious.</p>
 <p><a href="/unbelievable.html">a root-relative link to unbelievable</a>
 <a href="/unbelievable.html">a file-relative link to unbelievable</a></p>
-    </div><!-- /.entry-content -->
+        </div><!-- /.entry-content -->
 
-  </article>
+    </article>
 </section>
         <section id="extras" class="body">
                 <div class="social">
-                        <h2>social</h2>
-                        <ul>
+                    <h2>social</h2>
+                    <ul>
                             <li><a href="/feeds/all.atom.xml" type="application/atom+xml" rel="alternate">atom feed</a></li>
 
-                        </ul>
+                    </ul>
                 </div><!-- /.social -->
         </section><!-- /#extras -->
 
         <footer id="contentinfo" class="body">
-                <address id="about" class="vcard body">
+            <address id="about" class="vcard body">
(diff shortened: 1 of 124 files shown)


strburst:
If this PR isn't isomorphic, I will eat my own hat (metaphorically). 🎩

strburst:
Hi @justinmayer, sorry for the delay; I've rebased everything as requested. Additionally, I also took a closer look at the [official Jinja template documentation](http://jinja.pocoo.org/docs/2.10/templates/) and realized there were some indentation rules I didn't see before, so I've fixed this. I put a complete list of indentation rules in the commit message of 59750b3.

pelican/tests/output/basic/a-markdown-powered-article.html
```diff
@@ -1,65 +1,68 @@
...
-                    <li><a href="/pages/this-is-a-test-page.html">This is a test page</a></li>
-                    <li><a href="/category/bar.html">bar</a></li>
-                    <li class="active"><a href="/category/cat1.html">cat1</a></li>
-                    <li><a href="/category/misc.html">misc</a></li>
-                    <li><a href="/category/yeah.html">yeah</a></li>
-                </ul></nav>
+            <h1><a href="/">A Pelican Blog </a></h1>
+            <nav>
+                <ul>
+                            <li><a href="/tag/oh.html">Oh Oh Oh</a></li>
```
mosra:
```suggestion
                        <li><a href="/tag/oh.html">Oh Oh Oh</a></li>
```

If we're going for an OCD-correct indentation, this list should go four spaces back I think ;)

pelican/tests/output/basic/a-markdown-powered-article.html
```diff
@@ -1,65 +1,68 @@
...
-        </abbr>
+    <abbr class="published" title="2011-04-20T00:00:00+00:00">
+        Published: Wed 20 April 2011
+    </abbr>
 
-<p>In <a href="/category/cat1.html">cat1</a>.</p>
+    <p>In <a href="/category/cat1.html">cat1</a>.</p>
 
-</footer><!-- /.post-info -->      <p>You're mutually oblivious.</p>
+</footer><!-- /.post-info -->            <p>You're mutually oblivious.</p>
```
mosra: Maybe a newline should go here as well? Not sure about the indent tho.

```suggestion
</footer><!-- /.post-info -->
            <p>You're mutually oblivious.</p>
```

mosra:
*(Disclaimer: I'm an OCD freak when it comes to formatting. Take with a grain of salt :laughing: )*

In my themes I'm not indenting the `<head>` / `<body>` tags to avoid excessive indentation. Additionally for markup I'm personally using only 2 spaces so it then ends up [looking like this](https://github.com/mosra/m.css/blob/master/pelican-theme/test/layout/index.html). (Just an opinion, not demanding to use two-space indentation here as well ;).)

The indentation mostly makes sense, except for the two cases I mentioned above.  The [generated output](https://github.com/getpelican/pelican/blob/59750b3a8554f69f5ae632c4b98a9b63d95ea077/pelican/tests/output/basic/a-markdown-powered-article.html) still has quite a lot of (what seem to be rather random) empty lines, not sure what to do about that. Maybe remove them all? Some things coming from `{% block %}` also seem to be not indented -- there's the `indent()` Jinja2 filter that could help with that, but note that the indentation affects content of `<pre>` tags, so better not do that for the main page content. 

My thinking is that if we're going to do such a huge change in the formatting, we might as well finish it to the extreme to avoid doing similar work again in the future. (Because this always results in many coflicts for everyone who's modifying the themes as well.) It also makes testing easier, since with clean markup it's clearer to see what a particular change did to the theme.

strburst:
Hi @mosra, thanks for your comments! It makes me feel like someone cares about all the indentation work I had to do, haha.

First, it's important to note that we basically can't edit the generated output directly; it has to exactly match what is generated, because there are some tests that assert this (there's a little more info [here](http://docs.getpelican.com/en/stable/contribute.html#running-the-test-suite)). So sadly, I can't apply any of your suggested changes to `pelican/tests/output/*`. Looking at the templates that generated them, they do seem to be pretty reasonable already, but Jinja seems to like to join sub-blocks to the previous line in certain circumstances.

Second, it would be nice to do, but I have pretty much given up trying to make the HTML output look pretty; last time I looked at it, it seems like Jinja doesn't have a concept of indenting content substituted in from blocks and includes. For example,

```jinja
{# template_parent.html #}
<tag1>
    <tag2>
        <tag3>
            <p>From parent</p>
            {% block content %}
            {% endblock %}
        </tag3>
    </tag2>
</tag1>
```

```jinja
{# template_child.html #}
{% extends template_parent.html %}
{% block content %}
Hi everyone!
{% endblock %}
```

```html
<!-- Output -->
<tag1>
    <tag2>
        <tag3>
            <p>From parent</p>
Hi everyone!
        </tag3>
    </tag2>
</tag1>
```

Note that the child template is basically just substituted in line-wise, with no awareness of indentation. And of course, it would be a Bad Thing to couple child templates to how much indentation is in their parents. Hopefully, people will be looking at the templates that generated the output much more often rather than the output itself.

But if there were some sort of setting that adds the same indentation that the template was on into the file, that might be really nice. We could also run it through a prettifier/minifier, but that is only something the [assets plugin](https://github.com/getpelican/pelican-plugins/tree/master/assets) can do, and not native Pelican.

I believe changing indentation from 4 spaces to 2 spaces would actually be fairly simple; it'll just a `sed` command. I do think 2 spaces is more trendy lately in web development (especially Javascript); I only chose 4 spaces because the Jinja doc examples use it. So I could be easily convinced either way.

mosra:
> First, it's important to note that we basically can't edit the generated output directly;

Of course ;) To make myself clear, I was looking at the output to see what's wrong in the templates (and so all my suggestions are meant to be "this output looks wrong and the source template should be fixed").

> We could also run it through a prettifier/minifier

No, please not, that's a one-way ticket to hell. I know too many projects that started applying minifiers to their test results instead of fixing the root causes and it went only downhill from there.

> Jinja seems to like to join sub-blocks to the previous line in certain circumstances.

This can be ~~fixed~~ worked around by putting an explicit empty line in the template (yes, I know, it's less than ideal). Sometimes also things like `{%+ ` or `+%}` help to preserve the space before / after. You have to try though, I don't have a guide that would work 100%.

For the blocks I'm afraid I don't have a solution, I was always explicitly indenting the content of `{% block %}`. I don't think it would be a Bad Thing, since it's always clear how much the child block needs to be indented. Also, again, you **shouldn't** indent the main content because it will break `<pre>` tags. I'm doing something like this in that case, to denote where the unindented content is pasted:

```html+jinja
<tag1>
    <tag2>
        <tag3>
            <p>From parent</p>
<!-- content -->
{% block content %}
{% endblock %}
<!-- /content -->
        </tag3>
    </tag2>
</tag1>
```

For includes (which *can* happen at multiple places and so the indentation is not always clear), I have a workaround -- been using this myself on a few places:

```html+jinja
    <footer class="m-container">
      <div class="m-row">
        <div class="m-col-m-10 m-push-m-1 m-nopadb">
          {% macro footer() %}{% include "article_footer.html" %}{% endmacro %}
          {{ footer()|indent(10) }}
        …
```

PR #2380, opened 2018-07-09
importer: update links to attachments if --wp-attach

oulenz:
The `--wp-attach` flag of the importer downloads attached files to the content directory, but doesn't update links to these files within posts. This fixes that.


justinmayer:
Thank you for the enhancement, @oulenz.

PR #2381, opened 2018-07-11
automatically copy linked static files

oulenz:
This makes it so that linked static files are automatically copied to the output directory, removing the need to include the path in `STATIC_PATHS`. I chose to introduce a new `{static}` label so that it is now possible to link both to the generated `article.html` and `page.html` (with `{filename}`) and to their source (with `{static}`).

Includes documentation and tests.


pelican/tests/test_generators.py
```diff
@@ -698,21 +713,43 @@ def test_tag_and_category_links_on_generated_pages(self):
...
         settings['CACHE_PATH'] = self.temp_cache
         settings['DEFAULT_DATE'] = (1970, 1, 1)
+        context = get_context(settings)
 
         generator = PagesGenerator(
-            context=settings.copy(), settings=settings,
+            context=context, settings=settings,
             path=CUR_DIR, theme=settings['THEME'], output_path=None)
         generator.generate_context()
-        pages_by_title = {p.title: p.content for p in generator.pages}
```
oulenz: This line (pre-existing) in combination with the `{static}` and `{attach}` labels in `pelican/tests/TestPages/page_with_static_links.md` (introduced in this commit) caused `test_basic_generation_works` in `test_pelican.py` to fail on line 110, with a message that the log message is only found once instead of 3 times.

I spent hours on trying to find out how on earth this should be possible but didn't manage to narrow it down further than this. It looks like somehow the cache or the logs are polluted from this test, but I am absolutely befuddled how evaluating two extra static links here should reduce the number of log messages about three completely different static links in a test running on a  _different subtree of the project_. It's a complete mystery to me.

avaris:
I didn't look at this as closely as I'd like, so I might be wrong, but looks likes this relies very much on content generators (`Article` and `Page`) being processed *before* `Static` generator. That's certainly an implementation detail and not very reliable.

oulenz:
@avaris You're right that it relies on the `StaticGenerator` coming last, but that's very much already a feature of the current code base. Quoting from `__init__`:

> StaticGenerator must run last, so it can identify files that were skipped by the other generators, and so static files can have their output paths overridden by the {attach} link syntax.

In particular, there's a test that ensures the `StaticGenerator` is the last one.

justinmayer:

pelican/generators.py
```diff
@@ -175,28 +175,36 @@ def get_files(self, paths, exclude=[], extensions=None):
...
         return files
 
-    def add_source_path(self, content):
+    def add_source_path(self, content, static=False):
         """Record a source file path that a Generator found and processed.
         Store a reference to its Content object, for url lookups later.
         """
         location = content.get_relative_source_path()
-        self.context['filenames'][location] = content
+        key = '{}_content'.format('static' if static else 'generated')
```
avaris: This is a bit nit-picking but I find this much more readable:
```python
key = 'static_content' if static else 'generated_content'
```

pelican/contents.py
```diff
@@ -262,22 +263,34 @@ def _link_replacer(self, siteurl, m):
...
+            linked_content = _get_linked_content(key, path)
             if linked_content:
                 if what == 'attach':
-                    if isinstance(linked_content, Static):
-                        linked_content.attach_to(self)
-                    else:
-                        logger.warning(
-                            "%s used {attach} link syntax on a "
-                            "non-static file. Use {filename} instead.",
-                            self.get_relative_source_path())
```
avaris: So... What happened to this warning?

If my understanding is correct, you get a `Replacement indicator 'attach' not recognized` instead (which is very confusing).

pelican/generators.py
```diff
@@ -749,9 +759,12 @@ def __init__(self, *args, **kwargs):
...
         self.staticfiles = []
-        for f in self.get_files(self.settings['STATIC_PATHS'],
-                                exclude=self.settings['STATIC_EXCLUDES'],
-                                extensions=False):
+        linked_files = [os.path.join(self.path, path)
+                        for path in self.context['static_links']]
+        found_files = self.get_files(self.settings['STATIC_PATHS'],
+                                     exclude=self.settings['STATIC_EXCLUDES'],
+                                     extensions=False)
+        for f in chain(linked_files, found_files):
```
avaris: There seems to be a possibility of processing same file twice (thus emitting signals twice etc).

pelican/generators.py
oulenz: I agree...

pelican/contents.py
oulenz: Why would you expect to see that? I can only find `Replacement indicator %s not recognized` in the `else` branch starting in line 310, but at this point we've already entered the `if` branch.

I think I removed the warning because `isinstance(linked_content, Static)` will always be true: if you use `{attach}` on e.g. the source file of an article then that source file will be attached as a static file.

pelican/tests/test_contents.py
```diff
@@ -341,23 +342,25 @@ def test_intrasite_link_more(self):
         cls_name = '_DummyAsset' if six.PY3 else b'_DummyAsset'
 
         args = self.page_kwargs.copy()
-        args['settings'] = get_settings()
+        args['settings'] = get_settings(STATIC_PATHS=[])
```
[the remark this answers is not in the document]
oulenz: By default `STATIC_PATHS == ['images']`, but we want to test whether `'{static}'` also works for files that wouldn't have been copied over otherwise.

pelican/generators.py
oulenz: You're right, let me look at that.

pelican/contents.py
avaris: Hmm... That is tricky. 

`STATIC_EXCLUDE_SOURCES` is now useless then? At least for the auto-magically added static files? Might need to check for that.

I was assuming based on that. If you `{attach}` a source file and that file is not in `static_content`, or at least it's not supposed to be, then you get `None` for `linked_content` and fall through to the else branch in 310.

pelican/tests/test_contents.py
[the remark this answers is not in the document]
avaris: Sure, but you're filling out the `static_content` and `generated_content` manually just below. You won't go through the process of discovering them.

pelican/tests/test_contents.py
```diff
@@ -341,23 +342,25 @@ def test_intrasite_link_more(self):
...
+            'reference.rst':
+                type(cls_name, (object,), {'url': 'reference.html'}),
         }
 
         # video.poster
         args['content'] = (
             'There is a video with poster '
-            '<video controls poster="{filename}/images/poster.jpg">'
+            '<video controls poster="{static}/images/poster.jpg">'
             '<source src="|filename|/assets/video.mp4" type="video/mp4">'
```
avaris: This should be `|static|` as well.

pelican/contents.py
oulenz: The idea is if you're linking to a source file with `{static}`, that's an explicit confirmation that you do want that file. So `STATIC_EXCLUDE_SOURCES` only affects `STATIC_PATHS`. You might want to set it to false if you want to publish source files from an entire directory without explicitly linking to them (or in the template rather than in the input files), although I guess there is also `OUTPUT_SOURCES` so you may be right that it's not super necessary anymore.

pelican/tests/test_contents.py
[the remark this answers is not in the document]
oulenz: Yes, you're right of course, not sure what I was thinking at the time...

pelican/tests/test_contents.py
oulenz: Thanks, I've replaced a bunch of other occurrences of `|filename|` by `|static|` in the test material.

pelican/contents.py
avaris: OK, I'm fine with the fact that explicit `{static}` linking means you *want* the source. `{attach}` is sketchy though. Maybe check if `path` is in `generated_content` for `{attach}` and issue a warning?

oulenz: Do you think people might think they can change the path where a generated file ends up through `{attach}`, and thus mistakenly use it for that purpose?

I was thinking that when people want to include a link to a source file, they're about as likely as not to want that file to sit in the same folder as the file they're linking from, and thus choose `{attach}` over `{static}`. Then they would get this warning which they can't do anything about.

Moreover, I think the documentation is pretty clear about the fact that `{attach}` only works for static content.

avaris: I was thinking more about people accidentally using `{attach}` in place of `{filename}`, but you're right. There is a legitimate reason to use `{attach}` on a source file. I guess people need to be careful now :).

justinmayer:
Bravo, @oulenz! This is a really nice enhancement to Pelican. Thanks also to @avaris for the review and guidance. 🎉

Issue #2322, opened 2018-04-05
pelican-import getting error with pandoc@2.1.3


TheMushrr00m:
Trying to import all of my content from wordpress i am getting this error:

![screenshot_20180405_042617](https://user-images.githubusercontent.com/6392644/38357952-9f3540b2-3889-11e8-8bbb-7a0fef477fb2.png)

I am using the command:
`pelican-import --wpfile -o output laesporadelhongo.wordpress.2018-04-05.xml --dir-page --dir-cat -m markdown
`
The error it's related with the most recent pandoc's version 2.1.3 in Arch Linux.

TheMushrr00m:
I think that changing in [this line](https://github.com/getpelican/pelican/blob/master/pelican/tools/pelican_import.py#L731) how `pelican-import` invokes to `pandoc` the errors will be resolved.

sagamusix:
I changed the invocation of pandoc in `pelican_import.py` like this, it works with recent pandoc releases then:
```
            parse_raw = '-raw_html' if not strip_raw else ''
            cmd = ('pandoc  --from=html'
                   ' --to={1}{0} -o "{2}" "{3}"')
```

TheMushrr00m:
@sagamusix your solution works for me.
Thanks!

sagamusix:
Note that there is also a pending pull request which might be a bit more comprehensive and correct than my quick hack: #2289

TheMushrr00m:
I see, but your "quick hack" was what I needed at the moment and with that I was able to resolve my situation. Thanks again.

PR #2383, opened 2018-07-12
Convert FEED setting from `%s` to `{slug}`-style strings

### LINKED ISSUE #2106: `TAG_FEED_ATOM = 'label.{slug}.atom.xml'` breaks Pelican
When defining *_URL and *_SAVE_AS settings, `{slug}` replacements are valid. That is to say the following are valid:

```
TAG_URL = 'label/{slug}/'
TAG_SAVE_AS = 'label/{slug}/index.html'
```

However, if the same form is tried on RSS/Atom feed name settings, Pelican breaks with a critial error. That is to say, the following (which would follow from the above) is NOT valid:

```
TAG_FEED_ATOM = 'label.{slug}.atom.xml'
```

Instead, `%s` substitution must be used. That is to say, the following is valid:

```
TAG_FEED_ATOM = 'label.%s.atom.xml'
```

This seems to be the case for all RSS/Atom feed name settings.

Is it possible to get feed name settings to support `{}` substitution, to match *_URL and *_SAVE_AS settings?

MinchinWeb:
Closes #2106


justinmayer:
@MinchinWeb: It appears Travis is reporting test failures. Could you take a look?

MinchinWeb:
Success!

Flake8 issues have been fixed.

I have Pandoc v2.1.3 install locally and so some test fail due to that. See issues #2255 and #2322, and WIP PR #2289.

There are several other tests that fail locally (on Windows) for various reasons; some are things I haven't touched, so I'm leaving alone; some are due to 'symbolic link privilege not help'; some are due to Windows filepaths being written different than POSIX paths (these I skip now on Windows). These, as a whole, I'm ignoring.

The other broken test was a simple typo. Fixed now!

Python 3.7 has been released, but was failing to download on Travis; it can be added back at a later time.

justinmayer:
Would any @getpelican/reviewers be willing to take a moment and review this pull request?

mosra:
Yay, I personally welcome the change to `{}`-style! (Was also wondering myself why feeds use `%s` while everything else is `{}`-style.) But I think this should be done in a backwards-compatible way. With your proposed change, all setting that had the old-style 

```py
CATEGORY_FEED_ATOM = 'feeds/%s.atom.xml'
```

would now result in `feeds/%s.atom.xml` put *verbatim* into the generated output, **without any warning** that something's wrong. Discovering such breakage in each affected site can then take months and that's **bad**. So instead just calling `format()`, I'd propose something like this:

```py
self.settings['CATEGORY_FEED_ATOM']).format(slug=category.slug).replace('%s', category.slug)
```

And similarly elsewhere (maybe `replace()` first? not sure). That'll work for both approaches, with the `replace()` being removed some years later once the backwards compatibility is not needed anymore.

Besides that, the diff contains a lot of unnecessary formatting changes that make it *very hard* to review. Would it be possible to clean up the diff to not have them? Also, unrelated changes such as `posix_path()` or Windows test fixes could be better discussed in a separate PR.

MinchinWeb:
@mosra : I like you suggestion of how to support the two styles. I searched for examples of how to do this, and never did find anything useful...

Regarding the "extra" pieces, some of the reformatting is from the different lengths of `{slug}` vs `%s`, some was for line-length consistency. The second group could be removed, but there is no automated way to do it, I'd basically have to completely rebuilding the patch. The `posix_path()` changes again were for consistency. If you really want these two pieces out, I could maybe pull them out to a new PR, and then rebase this one on top of it. Is is worth the extra effort?

The Windows testing issues are similar (they could be pulled to a separate PR, and this one rebased on to it), but they were needed to get this far: when I first ran the test suite locally (on my Windows machine), a huge swatch of the tests failed, and so it was very hard to know what, if any, tests I had broken. Making the tests behave more sanely was the faster alternative to spamming Travis and crossing my fingers.

MinchinWeb:
@justinmayer @mosra : both old-style and new-style text substitutes are now supported!

```
self.settings['CATEGORY_FEED_ATOM'].replace('%s', '{slug}')
                                                       .format(slug=cat.slug),
```

pelican/generators.py
```diff
@@ -400,7 +402,8 @@ def generate_feeds(self, writer):
                     writer.write_feed(
                         arts,
                         self.context,
-                        self.settings['TAG_FEED_ATOM'].format(tag.slug),
+                        self.settings['TAG_FEED_ATOM'].replace('%s', '{slug}')
+                                                      .format(tag.slug),
```
mosra: Why not `.format(slug=tag.slug)` here (and the two cases below)?

MinchinWeb: Good catch. It's been fixed!

avaris:
Sorry for late-ish response but, I don't really like the `%s` replacement. 

For one, it'd fail if you have `{}`s around `%s`:
```python
>>> '%s'.replace('%s', '{slug}').format(slug='foo')
'foo'
>>> '{%s}'.replace('%s', '{slug}').format(slug='foo')
'{slug}'
>>> '{%s'.replace('%s', '{slug}').format(slug='foo')
  File "<stdin>", line 1, in <module>
ValueError: Single '}' encountered in format string
```

Second, I'd rather have these handled when the settings file is read (in `settings.py`).

And I'd propose throwing a warning and falling back to the default in case `%s` is in these settings.

MinchinWeb:
@avaris I like the idea of moving the conversion to the `settings.py` file.

I'm not sure what to do about the errors you mention, and I haven't found an example online support both string styles at the same time. Do we throw an error on `%s` strings, and make the users upgrade? Do we throw a warning, try the conversion, and hope for the best?

avaris:
A check in `settings.py`. If `'%s'` is in the string, then throw a warning about deprecation, replace it from the `DEFAULT_CONFIG`. Something like:
```python
for KEY in list_of_feed_setting_keys:
    if '%s' in setting[KEY]:
        logger.warning('%%s usage in %s is deprecated, use {name} instead. Falling back to default'. KEY)
        setting[KEY] = DEFAULT_CONFIG[KEY]
```
Very similar to [this](https://github.com/getpelican/pelican/blob/master/pelican/settings.py#L421-L426).

MinchinWeb:
@avaris @mosra : I've moved to avaris' suggestion of complaining of old style settings when the settings are first loaded and falling back to the default in this case. I'm of slightly mixed feeling as it requires the user to actively update their settings to work with the new Pelican version, and doesn't support backwards-compatibility, but on the other hand it avoids a bunch of odd corner cases that are hard to detect and resolve automatically. But "explicit is better than implicit", so let's go with it.

Does something in the documentation need to be added to explain this change?

The failing test is due to a change in how Markdown v3 outputs HTML for footnotes. The fix has been submitted separately as PR #2417.

avaris:
> Does something in the documentation need to be added to explain this change?

[Changelog](https://github.com/getpelican/pelican/blob/master/docs/changelog.rst) is usually the place for such stuff.

justinmayer:
... and the release announcement post. 😊

MinchinWeb:
Changelog updated!

justinmayer:
@avaris / @mosra: Any further comments on @MinchinWeb's pull request?

avaris:
Stuff related to this PR is OK, but there are a lot of unrelated stuff (some unwarranted formatting changes, tox?? etc) attached to it. Is it possible to clean this up a bit?

justinmayer:
@MinchinWeb: I agree it's best to keep pull requests as discrete and digestible as possible. Could you take a look and perhaps move anything not directly related to issue 2106 into separate pull requests?

avaris:
It's more about what it removes, which is the unrelated stuff reviewers have to go over otherwise.

MinchinWeb:
7 Pull Requests and 5 hours later, the pieces of this have all been split apart. #2432 is the core functionality I set out to add. I hope the various pull requests can be merged quickly and cleanly.

- #2428 -- add Windows and Python 3.7 support to Travis-CI
- #2429 -- reflow text to 79 character line limit
- #2431 -- build all paths using `posix_join()`
- #2432 -- covert all FEED settings from `%s` to `{slug}` (or `{lang}`)

Issue #735, opened 2013-02-20
Pagination for archive.html Vs A separate blog.html


amanjeev:
Is it a good idea to introduce pagination for archives? Pelican's default behavior is to be a blog so the pagination is only available for index.html which is the home page. I was wondering if there should be or is planned feature where you can 

Change the blog page from index.html to say blog.html?

Or 

Have pagination for the archive.html?

Update: So I thought to start developing this myself yesterday and got the development environment running so far. I am trying to go through and understand the source. Where is it preferred to ask questions if I am stuck, here, freenode or somewhere else?

justinmayer:
Addressed via the aforementioned pull request. Well done, @oulenz!

PR #2384, opened 2018-07-12
Control pagination per template

oulenz:
Fixes https://github.com/getpelican/pelican/issues/735

* Enables different pagination values per template
* Enables pagination for period archives (useful when `period_archives.html` inherits from `index.html`)
* Enables control over pagination for tags, categories and authors independently of `index.html`

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -755,10 +755,6 @@ Template pages
    ``DIRECT_TEMPLATES`` are searched for over paths maintained in
    ``THEME_TEMPLATES_OVERRIDES``.
 
-.. data:: PAGINATED_DIRECT_TEMPLATES = ['index']
-
-   Provides the direct templates that should be paginated.
-
 
 Metadata
 ========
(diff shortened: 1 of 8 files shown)


justinmayer:

mosra:
Looks good to me (though I didn't try it out myself) :)

Is there a way to explicitly say that a template shouldn't be paginated (instead of omitting it)? For example the `archives` template is not paginated by default, yet the documentation is not explicit about that and I personally struggled to find out why. What about this (`0` is no pagination, `None` is default):

```py
PAGINATED_TEMPLATES = {
    'index': None, 
    'tag': None, 
    'category': None, 
    'author': None,
    'archives': 0
}
```

Looking at the settings, there's the `DEFAULT_ORPHANS` setting. Would it make sense to have it per-template as well or is that just an unnecessary complication?

oulenz:
Thank you for having a look!

I'm hesitant about interpreting 0 as 'no pagination' since it wouldn't add any functionality but merely introduce a second way to achieve this. Calling this setting `PAGINATED_TEMPLATES` and the updated  description rather strongly imply that templates not included therein will not be paginated. But this is purely a personal preference and I can add this if others agree.

I considered `DEFAULT_ORPHANS`, but struggled to imagine a scenario where you'd want to have different values per template. I say let's leave it alone and reconsider the issue if someone explicitly asks for it.

mosra:
Okay! :)

justinmayer:
@mosra: Thank you for the extra pair of eyes!  👀 

Nice work, @oulenz. Everything looks good to me, including the deprecated setting warning. Many thanks for your work on this (and your patience). Much appreciated! 👍

PR #2389, opened 2018-07-25
Fix DeprecationWarning in tests about collections.abc

farseerfc:
This is to avoid DeprecationWarning about ABCs moved to collections.abc when running tests on python3.7


justinmayer:
Many thanks for the enhancement, Jiachen!

MinchinWeb:
c.f. #2383

PR #2366, opened 2018-06-26
Add pandoc2 support to pelican-import. Fix #2255

davidag:
Tests are not passing. I am working on it.

PR #2394, opened 2018-08-03
Add pandoc2 support to travis/tox

davidag:
- Install pandoc2 in travis and decide whether to use it or not in tox.
- A symbolic link is created in tox, due to not being able to make
  setenv work like this: `setenv = PATH={toxinidir}:{env:PATH}`
- Install pandoc2 only on pyxx-pandoc2
- Add new inclusion to travis matrix combination
- A build failed with InsecurePlatformWarning on Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,11 +7,15 @@ env:
   - TOX_ENV=py27
   - TOX_ENV=py34
   - TOX_ENV=py35
+  - TOX_ENV=py27-pandoc2
+  - TOX_ENV=py34-pandoc2
+  - TOX_ENV=py35-pandoc2
 matrix:
   include:
     - python: 3.6
-      env:
-      - TOX_ENV=py36
+      env: TOX_ENV=py36
+    - python: 3.6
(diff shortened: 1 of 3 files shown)


davidag:
When pr #2366 gets merged these changes should pass.

justinmayer:
Thanks for your work on this, David. I'm curious whether it makes sense to merge this, or instead wait until after potentially switching our CI/CD to GitHub Actions (#2726). What do you think?

davidag:
Given that the code is in part outdated and any effort would have to be redone later, maybe it's better to wait until the switch is completed. After that, I'll be happy to update this PR to support GitHub Actions.

justinmayer:
Thanks for submitting this, David. We have since switched Pelican's CI environment from Travis to GitHub Actions, and Pandoc installation is already part of the new CI process, so I think this pull request can be closed. If you find ways to improve and build upon the current CI environment, by all means please file an issue and we can talk about it.  (^_^)

Issue #2225, opened 2017-10-12
img's style tags output onto page


stuaxo:
Using the GUM theme, I imported my wordpress site.

Some of the images end up with a line of their styles output afterwards like this:

![screenshot from 2017-10-12 02-00-20](https://user-images.githubusercontent.com/179677/31474399-67f8cc2a-aef2-11e7-877c-69b02f4336f1.png)

Markdown looks like this
```
[![](http://www.stuartaxon.com/wp-content/uploads/2008/09/warn-300x294.png "warn"){.alignnone
.size-medium .wp-image-88 width="300"
height="294"}](wp-content/uploads/2008/09/warn.png)
```

And a fragment of the html
```
<a href="wp-content/uploads/2008/09/warn.png"><img alt="" src="http://www.stuartaxon.com/wp-content/uploads/2008/09/warn-300x294.png" title="warn">{.alignnone
.size-medium .wp-image-88 width="300"
height="294"}</a>
```

Any idea what's going on - do I need to configure markdown somehow, or is it a bug ?

stuaxo:
This looks like it would work with the markdown attr extension ... if it output ```{:``` instead of ```{```

justinmayer:
Were you able to find a solution to your problem here? If you need further assistance with this, please reach out via the Pelican IRC channel, as noted in the [How to Get Help](http://docs.getpelican.com/en/stable/contribute.html#how-to-get-help) section of the documentation. Thanks!

stuaxo:
Cheers, I wasn't able to find a solution

http://stuartaxon.com/2008/09/15/warn-there-is-another-system

Next time I look at this I'll try the irc channel, it'll be a while since I only work on stuff like this between paying gigs, and there are so many projects to fill free time with :D

oulenz:
I think I got to the bottom of this. Markdown will correctly parse these attributes with the `attr_list` extension, as @stuaxo suggested. The attribute list doesn't have to start with `{:`, it will take `{` just fine. 

There is another issue though: the attribute list must not include new line characters, see https://github.com/Python-Markdown/markdown/issues/482 for background. However, newline characters are introduced by pandoc, due to its default `--wrap=auto` setting.

The solution should be to add `--wrap=none` to the pandoc command. I've tested this and it seems to work fine. I would be happy to write a pr, but https://github.com/getpelican/pelican/pull/2366 should be merged first. @justinmayer

PR #2396, opened 2018-08-07
importer: don't wrap, because it breaks HTML attributes

oulenz:
Fixes https://github.com/getpelican/pelican/issues/2225


justinmayer:
Thank you for the enhancement, @oulenz!

Issue #2390, opened 2018-07-29
Unclear error message running pelican.server 


ryanfox:
Hello,

I recently upgraded from 3.7.1 to master. After building my site, I tried to run the server via `python -m pelican.server`, as previously. I got a new message:

    server.py: error: the following arguments are required: path

Ok, cool. I don't have to cd into output/ any more to run the server. Running `python -m pelican.server outupt/`:

    TypeError: __init__() missing 1 required positional argument: 'RequestHandlerClass'

That is... less than helpful. Googling doesn't have any pertinent info. After a little digging, I found the master branch docs already specify the new `pelican --listen` and that resolved it.

It took me a little bit to figure out what was going on - I wasn't expecting the command line UI to change on a minor version, and the message ended up being totally unrelated to what had actually happened.

I think it would be helpful for people upgrading from previous versions to give a clearer error message, maybe 'The pelican server should be run via `pelican --listen`'.

Thanks for all the work so far!

justinmayer:
Hi Ryan. I'm all for helpful error messages, so thanks for raising this issue. Thoughts in no particular order:

* When released, the Pelican 3.8 announcement post will include "how to upgrade" info, as usual.
* Rather than Google, solution might've been more easily found via the [dev docs](http://docs.getpelican.com/en/latest/) or the [closed milestone PRs](https://github.com/getpelican/pelican/milestone/12?closed=1).
* This CLI UI change will be introduced in version 3.8, which from my perspective isn't minor.

But as I said, I'm all for helpful error messages. Would you like to help us implement a better error message for this condition, in order to help upgraders who haven't read the announcement post? (cc:  @josch @Lucas-C)

oulenz:
@ryanfox While it is true that the devserver has been integrated into Pelican, the old way of doing things (running `python -m pelican.server`) was still supposed to work. The reason it didn't is because two bugs were accidentally introduced. I've fixed this in the attached PR and added a proper deprecation warning.

PR #2393, opened 2018-07-31
Clarify error message for deprecated dev server command

ryanfox:
… Fixes #2390

PR #2351, opened 2018-06-11
Remove develop_server.sh in favour of pelican serving static files itself - close #2176

Lucas-C:
Competing static site generators integrate the functionality of regenerating
content and serving it into their main executable. In pelican this
functionality used to be in an external script `develop_server.sh` which
resides in the blog base directory. This has the disadvantage that changes in
pelican can break the `develop_server.sh` scripts which will not automatically
be upgraded together with pelican by package managers. Thus, pelican should
integrate this functionality into its main executable.

To this end, this commit removes `develop_server.sh` and adds three command
line options to the pelican executable:

 * `-l/--listen` starts the HTTP server (`-s/--serve` was already taken)
 * `-p/--port` specifies the port to listen at
 * `-b/--bind` specifies the IP to bind to

`--listen` and `--autoreload` can be used together to achieve the same
effect that other static site generators offer: Serve files via HTTP
while at the same time auto-generating the content.

Since the `develop_server.sh` script was removed, pelican-quickstart looses the
`develop` option.

Since the `develop_server.sh` script was removed, the Makefile looses the
`stopserver` target and the `devserver` target is replaced by running `pelican
-l` in the foreground.

Since pelican now offers the `--listen` option, the fabfile uses that instead
of starting the socketserver itself.

PR #2397, opened 2018-08-08
Fix pelican.server __main__ code, add deprecation warning

oulenz:
PR https://github.com/getpelican/pelican/pull/2351 integrated the devserver into the main pelican module, replacing `pelican.server`. In the process, it accidentally broke `pelican.server` in two places. This fixes that and adds a deprecation warning.

Fixes https://github.com/getpelican/pelican/issues/2390 , which was closed with (but not fixed by) PR https://github.com/getpelican/pelican/pull/2393

@justinmayer


justinmayer:
Thank you, @oulenz! (P.S. You don't need to @ mention me... I get notifications for all repo activity.)

PR #2406, opened 2018-09-12
Don't process metadata links before all files are present

mosra:
@justinmayer @avaris I finally got back to this (sorry for the delay) and submitting a fix proposed in the discussion at https://github.com/getpelican/pelican/pull/2288/files#r204337359. Since #2288 caused regressions (and this is fixing them), I think it makes sense to have this included in the 3.8 release.

Description of the change is below, is just a summary of the discussion linked above.

- - - 

The `refresh_metadata_intersite_links()` is called again later, so this bit is only causing everything to be processed twice. Besides that, it makes the processing dependent on file order -- in particular, when metadata references a file that was not parsed yet, it reports an "Unable to find" warning. But everything is found in the second pass, so this only causes a superflous false warning and no change to the output.

This also brings back the `_summary` field, because some plugins might have depended on that, even though it's internal.

### DIFF
--- a/pelican/contents.py
+++ b/pelican/contents.py
@@ -142,6 +142,10 @@ class Content(object):
         if not hasattr(self, 'status'):
             self.status = getattr(self, 'default_status', None)
 
+        # store the summary metadata if it is set
+        if 'summary' in metadata:
+            self._summary = metadata['summary']
+
         signals.content_object_init.send(self)
 
     def __str__(self):
@@ -465,13 +469,23 @@ class Content(object):
 
     def refresh_metadata_intersite_links(self):
         for key in self.settings['FORMATTED_FIELDS']:
-            if key in self.metadata:
+            if key in self.metadata and key != 'summary':
                 value = self._update_content(
                     self.metadata[key],
                     self.get_siteurl()
                 )
                 self.metadata[key] = value
                 setattr(self, key.lower(), value)
+
+        # _summary is an internal variable that some plugins may be writing to,
+        # so ensure changes to it are picked up
+        if ('summary' in self.settings['FORMATTED_FIELDS'] and
+                'summary' in self.metadata):
+            self._summary = self._update_content(
+                self._summary,
+                self.get_siteurl()
+            )
+            self.metadata['summary'] = self._summary
 
 
 class Page(Content):
--- a/pelican/contents.py
+++ b/pelican/contents.py
@@ -142,10 +142,6 @@ class Content(object):
         if not hasattr(self, 'status'):
             self.status = getattr(self, 'default_status', None)
 
-        if (len(self._context.get('generated_content', [])) > 0 or
-                len(self._context.get('static_content', [])) > 0):
-            self.refresh_metadata_intersite_links()
-
         signals.content_object_init.send(self)
 
     def __str__(self):
--- a/pelican/tests/cyclic_intersite_links/first-article.rst
+++ b/pelican/tests/cyclic_intersite_links/first-article.rst
@@ -1,0 +1,7 @@
+First article
+#############
+
+:date: 2018-11-10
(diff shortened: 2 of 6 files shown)


mosra:
Turns out I had to call `refresh_metadata_intersite_links()` explicitly in the test because it had no chance of being called otherwise (it's done from the generators or `Pelican.run()` which are on a higher level than the `Page` object in the tests). To me that makes sense, but you should check if that's really a correct approach.

The tests now all pass.

pelican/contents.py
```diff
@@ -380,6 +377,23 @@ def summary(self, value):
...
+        logger.warning('_summary has been deprecated since 3.8.0. '
+                       'Use the summary decorator instead')
+        return self.summary
+
+    @_summary.setter
+    def _summary(self, value):
+        """another deprecated function to access summary"""
+
+        logger.warning('_summary has been deprecated since 3.8.0. '
+                       'Use the summary decorator instead')
```
avaris: summary *decorator*? You mean property ('ie. `.summary`)? or `metadata['summary']`?

First is a no go. The reason for someone using `._summary` is (or at least should be) because they can't use `.summary` for various reasons like intrasite links etc. 

And this is getting out of hand :). `._summary` was an internal variable for `.summary` property. Now, it's becoming property.

Also, I don't get why you need this. You don't need `refresh_metadata_intersite_links()` working on `summary` because it already has that handled via the `.summary` property. If you want to treat `summary` as any other metadata and rely on `refresh_metadata_intersite_links()` then the `.summary` getter need to be changed to not do that as well. Or if you want to keep things as is, `summary` can be excluded from `refresh_metadata_intersite_links()`.

Either way, there is some redundancy here.

mosra: > And this is getting out of hand :)

yes, sorry :see_no_evil: 

Hm, okay, so the only other option I see is to put back the `_summary`  internal variable and related parts in `__init__()` / `get_summary()` and in addition also explicitly process it in `refresh_metadata_intersite_links()`:

```diff
diff --git a/pelican/contents.py b/pelican/contents.py
index 09d388d4..3cae9acf 100644
--- a/pelican/contents.py
+++ b/pelican/contents.py
@@ -447,7 +447,7 @@ class Content(object):
 
     def refresh_metadata_intersite_links(self):
         for key in self.settings['FORMATTED_FIELDS']:
-            if key in self.metadata:
+            if key in self.metadata and key != 'summary':
                 value = self._update_content(
                     self.metadata[key],
                     self.get_siteurl()
@@ -455,6 +455,14 @@ class Content(object):
                 self.metadata[key] = value
                 setattr(self, key.lower(), value)
 
+        # _summary is an internal variable that some plugins may be writing to,
+        # so ensure changes to it are picked up
+        if 'summary' in self.settings['FORMATTED_FIELDS'] and 'summary' in self.metadata:
+            self._summary = self._update_content(
+                self._summary,
+                self.get_siteurl()
+            )
+            self.metadata['summary'] = value
 
 class Page(Content):
     mandatory_properties = ('title',)

```

Does this look better to you?

mosra:
@avaris I just rebased this on master, added an explicit repro for the false-positive warnings (so it's really clear what this PR is fixing -- I get like 10 of such false-positives for my [moderately sized site](https://magnum.graphics/)) and reimplemented the `_summary` "compat" using the patch above. Would be great if you could have a look :)

avaris:
Oops, sorry. I forgot I didn't finish looking at this. Tomorrow, promise ;).

mosra:
@avaris yay, thank you for the review! and sorry it took so long until i arrived with an acceptable solution :)

justinmayer:
Many thanks to @mosra and @avaris for working on this fix. Much appreciated! 👍

Issue #1243, opened 2014-01-25
Confusing error when docutils is not installed


stephane-caron:
The error returned by Python when docutils is not installed is not explicit: it says that HTMLTranslator is not defined (needed by FeedGenerator and such), and one needs to go into readers.py to figure out that this happens because the "import docutils" failed.

Could you fix this, so that either it complains or the project is able to run when docutils is not installed?

PR #2408, opened 2018-09-18
Make the markdown package requirement explicit.

haridas:
Similar changes had been done on this
fix. (https://github.com/getpelican/pelican/issues/1243)
When using markdown documents currently the error messages doesn't say
the actual error, which forces the developer to debug the pelican code
to spot the bug. This fix will explicitly enforce the required dependency.

### DIFF
--- a/pelican/readers.py
+++ b/pelican/readers.py
@@ -21,10 +21,7 @@ from pelican.contents import Author, Category, Page, Tag
 from pelican.utils import SafeDatetime, escape_html, get_date, pelican_open, \
     posixize_path
 
-try:
-    from markdown import Markdown
-except ImportError:
-    Markdown = False  # NOQA
+from markdown import Markdown # NOQA
 
 # Metadata processors have no way to discard an unwanted value, so we have
 # them return this value instead to signal that it should be discarded later.
--- a/setup.py
+++ b/setup.py
@@ -7,7 +7,7 @@ from os.path import join, relpath
 from setuptools import setup
 
 requires = ['feedgenerator >= 1.9', 'jinja2 >= 2.7', 'pygments', 'docutils',
-            'pytz >= 0a', 'blinker', 'unidecode', 'six >= 1.4',
+            'pytz >= 0a', 'blinker', 'unidecode', 'six >= 1.4', 'markdown',
             'python-dateutil']
 
 entry_points = {


avaris:
What error message? If you don't have `markdown` pelican should skip those files.

haridas:
It does skip that particular article, but the error it spit out isn't clear about what's happening. It leads into debugging the library code to locate this issue. I faced this two times and done debugging with `pdb` to locate this minor packaging missing issue.

The main reason for this PR is, the error should be clearly mention the actual problem.

```
ERROR: Could not process articles/kmean-clustering-with-anomaly-detection.ipynb
  | TypeError: 'bool' object is not callable
Done: Processed 22 articles, 0 drafts, 1 page and 0 hidden pages in 1.59 seconds.
Done
```

avaris:
Sure, but I'm fairly certain that the error you pasted comes from a plugin, not pelican itself.

haridas:
If you checked the file changes in this PR it comes in `pelican/readers.py` module. The `markdown` package is getting imported there and exposing the `MarkdownReader` class in it. There is a flag this class expose `enabled` to check the Reader is enabled or not, but it doesn't say how to enable if required.

Also it's expected that pelican will support the standard files like `rst` and `md` by default.

Regarding the `--debug` flag, which doesn't point to the actual issue, which shows the error like this, 

```
DEBUG: Read file articles/what-i-like-about-j2ee.rst -> Article
DEBUG: Read file articles/vim-upgrade-2017.rst -> Article
DEBUG: Read file articles/kmean-clustering-with-anomaly-detection.ipynb -> Article
ERROR: Could not process articles/kmean-clustering-with-anomaly-detection.ipynb
  | TypeError: 'bool' object is not callable
  |___
  | Traceback (most recent call last):
  |   File "/Users/haridas/ENV/lib/python2.7/site-packages/pelican/generators.py", line 523, in generate_context
  |     context_sender=self)
  |   File "/Users/haridas/ENV/lib/python2.7/site-packages/pelican/readers.py", line 526, in read_file
  |     content, reader_metadata = reader.read(path)
  |   File "/Users/haridas/Projects/personal/github/haridas.github.com/./plugins/ipynb/markup.py", line 63, in read
  |     _content, metadata = md_reader.read(metadata_filepath)
  |   File "/Users/haridas/ENV/lib/python2.7/site-packages/pelican/readers.py", line 292, in read
  |     self._md = Markdown(**self.settings['MARKDOWN'])
  | TypeError: 'bool' object is not callable
```
Error comes at this line `self._md = Markdown(**self.settings['MARKDOWN'])`.

IMO If we say that the plugin has to explicitly check whether this `MarkdownReader` is enabled or not is kinda pushing this dependency check outside the core module.

This same kind of issue happened for the `docutils` some time back, if you check this https://github.com/getpelican/pelican/pull/1283/files PR you can see that.

avaris:
Leaving the discussion about whether or not `markdown` should be hard requirement aside...

Yes, a plugin should check the enabled status of a reader before attempting to use it. That mechanism exists for a reason. If the plugin *relies* on that reader to be enabled, it can helpfully provide the appropriate warning if otherwise.

haridas:
Ok @avaris, I understood the intention behind it. I will push the change to plugin code.

My initial idea was, `docutils` and `markdown` are kinda doing similar jobs for `rst` and `md` formats, and I was expecting `md` to be supported as default with pelican.

Issue #2412, opened 2018-10-02
mtime issue on Debian stretch make files copy quasi systematic even if STATIC_CHECK_IF_MODIFIED set to True


jfbaillette:
Issue on commit 9e526a825ee308c9e982bf0eca427ea4fdfc0f8d.

> python copy instruction doesn't preserve mtime under Debian/stretch (up to date).

This results in lot of files are copied to output even if not changed.

**Proposed solution in pelican/pelican/generators.py, line 823:**
        #return s_mtime > d_mtime
        return s_mtime - d_mtime > 0.000001

Exemples from stat:

| Source | Output (created by copy) |
| ----------------------------------- | ----------------------------------- |
| 2018-09-26 16:16:23.298790930 +0200 | 2018-09-26 16:16:23.298790000 +0200 |
| 2018-09-26 16:00:53.379527955 +0200 | 2018-09-26 16:00:53.379528000 +0200 |
| 2018-09-26 16:05:24.019213477 +0200 | 2018-09-26 16:05:24.019213000 +0200 |
| 2018-09-26 16:14:52.660719249 +0200 | 2018-09-26 16:14:52.660719000 +0200 |
| 2018-09-26 16:05:53.790765421 +0200 | 2018-09-26 16:05:53.790765000 +0200 |
| 2018-09-26 17:00:02.930290005 +0200 | 2018-09-26 17:00:02.930289000 +0200 |
| 2018-09-26 17:01:30.525954070 +0200 | 2018-09-26 17:01:30.525954000 +0200 |
| 2018-09-26 15:58:24.850144752 +0200 | 2018-09-26 15:58:24.850144000 +0200 |

After this patch, my pelican ran from 300s to ... less than one second...

avaris:
Well, that's annoying. Looks like a truncation issue. 

Your solution sounds reasonable to me. Want to turn it into a PR :) ?

jfbaillette:
Sounds reasonable. PR #2414 proposed.

PR #2414, opened 2018-10-03
Resolving mtime truncation (issue #2412)

jfbaillette:
Issue #2412 : python copy instruction doesn't preserve mtime under Debian/stretch (up to date).
This results in lot of files are copied to output even if not changed (and a lot of wasted time).

Proposed solution in pelican/pelican/generators.py, line 823:
#return s_mtime > d_mtime
return s_mtime - d_mtime > 0.000001

### DIFF
--- a/pelican/generators.py
+++ b/pelican/generators.py
@@ -820,7 +820,7 @@ class StaticGenerator(Generator):
         save_as = os.path.join(self.output_path, staticfile.save_as)
         s_mtime = os.path.getmtime(source_path)
         d_mtime = os.path.getmtime(save_as)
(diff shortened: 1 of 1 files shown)


avaris:
Brilliant! python-markdown [broke](https://python-markdown.github.io/change_log/release-3.0/#output_formats-simplified-to-html-and-xhtml) [itself](https://github.com/Python-Markdown/markdown/blob/master/markdown/extensions/footnotes.py#L144)...

avaris:
@justinmayer I just did.

justinmayer:
Fantastic. Thank you! 😊

PR #2415, opened 2018-10-03
Remove Atom summaries when they are unnecessary

### LINKED ISSUE #2403: Full Atom feeds contain empty summaries
When providing the full content of articles in an Atom feed, the generated file contains an empty `summary` in each entry. It should only contain a `content` and no `summary`.

According to the [Atom RFC](https://tools.ietf.org/html/rfc4287):

> It is advisable that each atom:entry element contain a non-empty atom:title element, a non-empty atom:content element when that element is present, and a non-empty atom:summary element when the entry contains no atom:content element.

The [W3C validator](https://validator.w3.org/feed/) raises a warning when it finds such empty summaries.

Sample configuration:

```python
FEED_DOMAIN = SITEURL
FEED_ALL_ATOM = 'feeds/atom.xml'
FEED_ALL_RSS = 'feeds/rss.xml'
ATOM_FEED_SUMMARY_ONLY = False
RSS_FEED_SUMMARY_ONLY = False
CATEGORY_FEED_ATOM = None
TRANSLATION_FEED_ATOM = None
AUTHOR_FEED_ATOM = None
AUTHOR_FEED_RSS = None
SUMMARY_MAX_LENGTH = 0
```

Excerpt of the generated Atom feed:
```xml
    <entry>
        <title>Foo</title>
        <link href="https://foo.com/bar.html" rel="alternate"></link>
        <published>2018-03-13T18:27:00+01:00</published>
        <updated>2018-03-13T18:27:00+01:00</updated>
        <author>
            <name>Foo</name>
        </author>
        <id>https://foo.com/bar.html</id>
        <summary type="html"></summary>
        <content type="html">This is the content.</content>
    </entry>
```

NicolasLM:
Do not include a summary tag to Atom entries when:
- the summary is empty
- the summary and the full content are the same

Fixes #2403


NicolasLM:
Tests are failing because when I regenerated the output it did not only replace the Atom files but a lot of unrelated HTML, so I believe something is wrong with my setup. Some help would be appreciated.

avaris:
Code looks OK but you didn't push the regenerated output?

PS: Ignore the `test_article_with_footnote` failure, for now. That's not your fault.

NicolasLM:
I pushed the regenerated output but I think that something is off. I probably don't have the correct locale installed.

NicolasLM:
I installed the French locale and regenerated the output, it looks better.

NicolasLM:
Code rebased and test files updated. Bump @avaris :)

justinmayer:
Many thanks to @NicolasLM for all the hard work on this endeavor, and kudos to @avaris for reviewing it! 🚀

PR #2416, opened 2018-10-03
Make relative source path available for themes

jfbaillette:
| Item | Explanation |
| --- | --- |
| Need | Figure out in themes the **relative path name** from the file (md or rst) which is generated |
| Typical use case  | Users may **directly modify their article** from a link opening GitHub in edition mode |
| Proposal | Make a **article.orig_relpath** available for themes article.html template |

# Example

In *YOURTHEME*/template/article.html:
``` html
...
<button onclick="window.location.href='https://github.com/getpelican/pelican/edit/master/{{ article.orig_relpath }}'">
  Modify this article under GitHub
</button>
...
```

### DIFF
--- a/pelican/generators.py
+++ b/pelican/generators.py
@@ -595,6 +595,7 @@ class ArticlesGenerator(CachingGenerator):
                 all_articles.append(article)
             elif article.status == "draft":
                 all_drafts.append(article)
+            article.orig_relpath  = f;
             self.add_source_path(article)
 
         self.articles, self.translations = process_translations(all_articles)


avaris:
Umm... You mean `article.source_path`? See [here](https://github.com/getpelican/pelican/blob/master/pelican/contents.py#L37) and [here](https://github.com/getpelican/pelican/blob/master/pelican/contents.py#L110).

jfbaillette:
1. `article.source_path` gives an *absolute path* to the doc. Ex: `/home/project/dev/content/explanations.md`.
1. `article.orig_relpath` gives the *relative path* to the doc. Ex: `content/explanations.md`.

The second may be used to forge URLs enabling access in GitHub via a button as proposed in the PR?

avaris:
Fair enough, but this should be handled in the `Content` similar to `source_path` so that other contents (like pages) can also use it.

And for consistency, I'd suggest names like `relative_source_path` or `rel_source_path`.

justinmayer:
@jfbaillette: There are also some code style violations, as you can see by the failing Flake8 test here. The best way to resolve that is to run Flake8 locally and fix any problems before making+pushing commits.

PR #1984, opened 2016-08-01
add the --dump-settings command line argument

leotrs:
Addresses [#1762](https://github.com/getpelican/pelican/issues/1762).

Issue #1762, opened 2015-06-18
make clean should empty cache


danmackinlay:
#1742 and #1722 both mention the confusing behaviour of corrupted caches.

It is perhaps counterintuitive that the default makefile does not empty the cache, just the environment.

For the sake of intuitiveness, it would be wise to clear the cache too when one does a `make clean`

howthebodyworks:
FWIW, I know that this cache thing is regarded as an upgrade issue, but I get occasional AttributeErrors like #1722 from non-reproducible transient cache corruption; easing cache nuking looks like an easy win compared to fixing transient cache errors, esp given #1759.

justinmayer:
Hi Dan. Thanks for your comments. Is something you would like to work on?

ingwinlu:
the cache directory is configurable (`CACHE_PATH`), so 'guessing' a directory in the Makefile is a no-go for me. To realize that we would need to add a command line argument to pelican itself which parses the configuration for the cache directory and deletes it.

leotrs:
@ingwinlu if there's still interest in this (it sounds good to me), I could try to prepare a PR at some point (no promises).

justinmayer:
> it would be wise to clear the cache too when one does a make clean

Seems sensible enough to me.

@leotrs: Would welcome a PR to that effect if you can spare the time.  (^_^)

leotrs:
As per @ingwinlu's comment, the cleanest way I can see this happening is asking pelican to parse the config file and outputting the `CACHE_PATH`, so make clean can just run ```rm `pelican --output-cache-path```` or some such. Is that the way we want to go?

justinmayer:
@leotrs: Seems like a fine approach to me.

leotrs:
While we're at it, we can have a `pelican --read-config-file` and dump the whole final settings dict. This could be useful in debugging. Then, `make clean` can just `grep` the output and do its thing.

leotrs:
@justinmayer, [I added the](https://github.com/getpelican/pelican/pull/1984) `--dump-settings` command line argument, with the aforementioned functionality. Run from the `pelican/tests` folder in the main repository:
```
tests $ pelican -s default_conf.py --dump-settings 'GITHUB_URL'
'http://github.com/ametaireau/'
```

Then I tried to edit the Makefile as requested, but my bash is failing me. I tried to do the following, without success.
```
clean:
	[ ! -d $(OUTPUTDIR) ] || rm -rf $(OUTPUTDIR)
[...]

PR #2422, opened 2018-10-23
Add CLI argument to print current settings

### LINKED ISSUE #2275: Dump variables on debug
Having a double verbose debug which dump variables both CONF variables and variables used in templates would help a lot while debugging :)

justinmayer:
This is a reworked version of PR #1984 that, among other changes, implements the improvements that @Scheirle suggested.

Any further comments, from @Scheirle or other @getpelican/reviewers?

Fixes #2275
Refs #1762

### DIFF
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -7,11 +7,11 @@ import locale
 import logging
 import multiprocessing
 import os
+import pprint
 import re
 import sys
 import time
 import traceback
-from pprint import pprint
 
 import six
 
@@ -270,6 +270,32 @@ class Pelican(object):
             return writer(self.output_path, settings=self.settings)
 
 
+class PrintSettings(argparse.Action):
+    def __call__(self, parser, namespace, values, option_string):
+        instance, settings = get_instance(namespace)
+
+        if values:
+            # One or more arguments provided, so only print those settings
+            for setting in values:
+                if setting in settings:
+                    # Only add newline between setting name and value if dict
+                    if isinstance(settings[setting], dict):
+                        setting_format = '\n{}:\n{}'
+                    else:
+                        setting_format = '\n{}: {}'
+                    print(setting_format.format(
+                        setting,
+                        pprint.pformat(settings[setting])))
+                else:
+                    print('\n{} is not a recognized setting.'.format(setting))
+                    break
+        else:
+            # No argument was given to --print-settings, so print all settings
+            pprint.pprint(settings)
+
+        parser.exit()
+
+
 def parse_arguments():
     parser = argparse.ArgumentParser(
         description='A tool to generate a static blog, '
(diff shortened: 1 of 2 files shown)


pelican/__init__.py
```diff
@@ -527,6 +533,23 @@ def main():
 
     try:
         pelican, settings = get_instance(args)
+
+        if args.print_settings:
+            from pprint import pprint
```
MinchinWeb: Is there a reason not to have this at the top of the file with the other import statements?

justinmayer: No reason. I agree it's better at top. Fixed. Thanks!

pelican/__init__.py
```diff
@@ -319,6 +320,12 @@ def parse_arguments():
                         help='Relaunch pelican each time a modification occurs'
                         ' on the content files.')
 
+    parser.add_argument('--print-settings', dest='print_settings',
+                        action='append', nargs='?', metavar='SETTING_NAME',
+                        help='Print current configuration settings and exit. '
+                        'Append a setting name as argument to see the value '
+                        'for that specific setting only.')
+
```
avaris: Isn't `nargs='*'` with default `action='store'` more appropriate?

Logic afterwards will change though.
```shell
pelican # no --print-settings
# -> args.print_settings == None

pelican --print-settings
# -> args.print_settings == []

pelican --print-settings FOO BAR
# -> args.print_settings == ['FOO', 'BAR']
```

And if you want to get fancy and avoid putting logic inside `main()`, you can subclass [`Action`](https://docs.python.org/3/library/argparse.html#argparse.Action) then do stuff in `__call__` and at the end call `parser.exit()` to terminate. In that case `action=MyActionSubclass`

avaris: If I'm not mistaken, with the current setup you have to do
```shell
pelican --print-settings FOO --print-settings BAR # ...
```
in order to specify multiple options.

justinmayer: Thanks for the input. I hadn't planned on supporting multiple setting arguments, but I added a commit that should now support invocations such as:

    pelican --print-settings MARKDOWN ARTICLE_SAVE_AS

I agree that putting logic inside `main()` is less than ideal, but when I tried subclassing Action and putting the logic inside `__call__` instead, it wasn't clear to me how to access the various variables that are plainly available in the `main()` context. It started to feel like a slippery rabbit hole, so I aborted. If you can provide more detail as to how the logic would actually fit into `__call__`, I'm certainly willing to try again.

justinmayer: That did indeed point me in the right direction. Thanks! I pushed another commit that moves the logic out of `main()`. Any other thoughts on this before it's merged?

avaris: Looks good :)

justinmayer:
Many thanks to @MinchinWeb and @avaris for the assistance with this! 😁

PR #2424, opened 2018-10-29
Enable both pages & categories on Simple theme menu. Fixes #2404

### LINKED ISSUE #2404: Simple theme places pages or categories, but not both, on nav bar.
There is a logic error in the base.html file of the simple theme. It prevents Pelican from placing links to both static pages and categories on the menu bar.  You get one  or the other, not both.

I know how to fix this. How would you like me to go about it?

greeder59:
I changed the logic of the nav bar creation by removing the else statement that prevented categories from being added to the nav bar if pages are present. And, of course, I cleaned up the indent levels and end tags.


justinmayer:
Many thanks to Gordon for the contribution and to Deniz for reviewing. 👏

PR #2426, opened 2018-10-31
Add current date when publishing via ghp and Invoke

jagmoreira:
Re-submission of #2083 based on Invoke's task.py.


justinmayer:
Thank you for the contribution, Joao.

PR #2427, opened 2018-11-02
Add Python 3.7 support to Travis and Tox

MinchinWeb:
This seems to be the official way (c.f. https://github.com/travis-ci/travis-ci/issues/9815#issue-336465122)

Is there a reason not to update the doc generate and flake8 test to Python 3.7 as well?


MinchinWeb:
c.f. #2383.

PR #2428 is a more complete solution that includes adding Windows testing, and is probably the long term preferred solution, but that said, that should not hold back merging this in as an interim fix to test Python 3.7.

justinmayer:
Many thanks to @MinchinWeb for the enhancement and to @avaris for reviewing.

PR #2428, opened 2018-11-03
Travis: Add Windows testing support

MinchinWeb:
Experimental Windows support has been enabled on Travis-CI. The PR implements testing Pelican on Windows for Python 2.7 and 3.4 to 3.7. It also adds testing for Python 3.7 on Linux.

Docs building and flake8 tests are not run on Windows.

Python installations on Windows (at least on Travis) are not recognized by tox, so the nosetest command is run directly.

The testing suite on does not currently pass (and generates over 1,000 lines of output to tell you so), and so are classified as "allowable failures" on Travis. The hope is that this is one of the first steps to getting the test suite to pass cleanly on Windows.

Closes #2427. c.f. #2383.


.travis.yml
```diff
@@ -1,26 +1,200 @@
...
+      - sudo apt-get update
+      - sudo apt-get install python3.7
+      - python -m pip install virtualenv --user
+      - virtualenv venv
+      - source venv/bin/activate
+      - python -m pip install --upgrade pip wheel
+      - pip --version
+      - python -m pip install --upgrade pip wheel
+      - python -m pip install -r requirements/test.pip
+      - python -m pip install nose nose-cov coveralls pygments==2.1.3
```
avaris: These should be `python3` and `pip3` I think. It's using 2.7 right now:
```
pip 18.1 from /home/travis/build/getpelican/pelican/venv/local/lib/python2.7/site-packages/pip (python 2.7)
```

MinchinWeb: Now using the Python 3 version installed.

Turns out getting the Linux testing working again has been more involved that getting the Windows tests to work...

avaris:
Looks like `pandoc` isn't installed in the linux environments:
```
SKIP: [u'pandoc', u'--version'] executable not found
```

.travis.yml
```diff
@@ -1,26 +1,212 @@
...
+      - sudo apt-get update
+      - sudo apt-get install python3.7
+      - curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py
+      - python3.7 get-pip.py --user
+      - python3.7 -m pip install virtualenv --user
+      - virtualenv venv
+      - source venv/bin/activate
+      - python -m pip install --upgrade pip wheel
+      - pip --version
+      - python -m pip install --upgrade pip wheel
```
avaris: Doing this twice.

avaris:
There should be a way to reduce the duplication...

.travis.yml
MinchinWeb: second one dropped!

MinchinWeb:
@avaris : Duplication can be reduced, but that requires a bunch of Bash scripting, and working in Python on Windows. I've started the work [here](https://github.com/MinchinWeb/pelican/blob/windows-tests-3/.travis.yml), but my Bash skills are falling short.

@justinmayer : The current set up is working. Is it ready to merge?

justinmayer:
Thanks for your work on this, @MinchinWeb. I'm curious whether it makes sense to merge this, or instead wait until after potentially switching our CI/CD to GitHub Actions (#2726). What do you think?

MinchinWeb:
I'm always leery of putting things, especially open source work, off until "some day". (By way of personal example, I started moving my Pelican plugins over to the common repo/organization, but it's been stalled for maybe a year?)

My thought would be to do this now, assuming this still works. It has the benefit of making sure anything added to Pelican before the move to GitHub Actions doesn't break Windows (or break it more). As well, it makes it clear that to reach feature parity, the GitHub Actions' CI pipeline should support Windows as well.

But I guess that also depends on how fast the switch to GitHub Actions will take. Is it a week long project, or a six month project? The longer the transition is likely to take, the more value there is in merging it now.

For what it's worth, I choose Python because Pelican supported Windows. Jekyll and Ruby was the other option I was considering at the time, and one of the packages Jekyll depended on had explicitly decided not to support Windows.

Just my two cents.

MinchinWeb:
Whatever is the solution, the sooner it gets done, the less work it will be. I just submitted #2739 because the project's `tasks.py` was hardcorded to Linux and so failed hard trying to get set up for Windows. Also, working on #2738, I get 32 failing tests, but it's not obvious which were failing because of Windows, and which because my pull request had broken things. (I think only 12 of them are due to Windows, and most of those due to the changes in #2738...)

avaris:
I was playing with GitHub Actions today and thought `pelican` would be a good playground for this :). It's very much WIP and there are parts I don't quite like, but [here is the progress](https://github.com/avaris/pelican/actions/runs/86142382). And, yes, situation on Windows isn't great. I didn't look into errors in detail yet, but some of those could be actual bugs in `pelican`.

MinchinWeb:
@avaris I'm actually really excited for Windows testing on Github Actions! Travis would [ruyour tests](https://docs.travis-ci.com/user/reference/windows/) in Git Bash, rather than PowerShell. The end result is that the tests are run in an environment that doesn't really match what end users will typically have set up, and I've had tests that will pass on Travis (on bash) but fail on locally (on PowerShell). I'm hoping GitHub Actions, by running the tests directly in PowerShel, will do away with those issues!

PR #2430, opened 2018-11-03
Docs: ghp-import works on Windows now

MinchinWeb:
see https://github.com/davisp/ghp-import/pull/25


justinmayer:
Thank you, @MinchinWeb!

Issue #1704, opened 2015-04-21
Default notmyidea theme is not mobile-friendly and will depress SEO


georgevreilly:
The notmyidea theme is not mobile-friendly, according to [Google's Site Tester](https://www.google.com/webmasters/tools/mobile-friendly/?url=http%3A%2F%2Fblog.getpelican.com%2F). Google Search is now [penalizing the search rankings of non-mobile-friendly sites](http://searchengineland.com/tomorrow-is-mobilegeddon-are-you-ready-for-the-google-mobile-friendly-update-21929). Pelican users who care about SEO will likely see a drop in their rankings.

qingkaikong:
Yes, I found this not mobile friendly as well, do we have a solution?

mosra:
Sorry to revive an old thread, but I have something that might interest you: http://mcss.mosra.cz

It's a lightweight, self-contained CSS framework *and* Pelican theme that is designed to be responsive and mobile-friendly. Currently it's a standalone thing, but I'm happy to help integrating it (or parts of it) into Pelican.

jorgesumle:
It's a nice theme. I like that it doesn't require an JavaScript. @mosra, if you make a pull request, please remove the external dependencies like Google Fonts. Basically, default themes should not breach users' privacy #2178

mosra:
It's not *that* easy; the theme itself depends on a plugin that adds a bunch of Jinja2 filters and also fixes the HTML output to be proper HTML5 ([details](http://mcss.mosra.cz/plugins/htmlsanity/)). I'm slowly trying to integrate the important parts of the plugin into Pelican itself (see e.g. #2163 or #2164), but that takes time and I'm not sure if all the things that the plugin does would be welcome :)

mosra:
@justinmayer any thoughts on this? 

I think we can all agree that the default notmyidea theme is not sufficient anymore and might scare away potential new users. With [m.css](http://mcss.mosra.cz) I have a complete package for publishing content with Pelican that's lightweight, mobile-friendly and self-contained. Currently it's a fully standalone thing, however I think it's sufficiently minimal to be included as one of the builtin themes and make Pelican look better out-of-the-box. But:

- The CSS is based around modern HTML5 and what the builtin docutils reStructuredText HTML writer produces is *far from that*. I'm fixing it [with a plugin](http://mcss.mosra.cz/plugins/htmlsanity/) that's also providing some more aids to the theme (which were originally part of #2164, but I eventually moved them out to this plugin).
[...]

Issue #1280, opened 2014-03-03
Add a new default theme which is responsive and widely reusable


naturallymitchell:
In this issue I would like to solve:
- #668 (A mobile aware theme) where there is interest in adding a theme that works across display sizes, aka responsive.
- #516 (theme notmyidea should use the simple theme) where there is agreement that Pelican's default theme(s) should reuse its own inheritable features.
- The visions of [justmay - August 6th 2013](https://botbot.me/freenode/pelican/msg/5020211/) and [kylef - 7 months ago](https://github.com/getpelican/pelican/issues/516#issuecomment-21623430) for a new theme in core that other themes can reuse.

This issue builds on #1092 (Inherit from arbitrary theme) which makes `!simple`'s inheritance functionality available to themes such as this one. 

I don't have a PR yet, but here's the code:
- [`base`](https://github.com/mitchtbaum/base) theme - a highly organized, highly reusable starter theme which makes great use of Jinja's blocks functionality
- [`fixed_sidebar`](https://github.com/mitchtbaum/fixed_sidebar) theme - a common blog pattern
- [`top_welcomebar`](https://github.com/mitchtbaum/top_welcomebar) theme - a basic Ghost theme
- [`corkboard`](https://github.com/mitchtbaum/corkboard) theme - a template targeted for collaboration sites

These three examples demonstrate the base theme, and while not absolutely necessary, I believe that these three example themes, or some other ones, would be very useful for demonstrating the foundation code and driving adoption.  Not just that..  Having more options out-of-the-box would improve usability, where users can choose how they'd like to begin using Pelican.  Also, it would increase the diversity of looks-n-feels of Pelican.

I'm not an HTML/CSS expert, so before these themes are ready, they will need some care.

leotrs:
What happened to those themes? Themes already live in a different repo.

leotrs:
@jsonbecker @mitchtbaum ^

justinmayer:
I would like to choose a Pelican theme as a test case for community-maintained themes, much like we have done under the [Pelican Plugins](https://github.com/pelican-plugins) organizational banner.

@DragonSerw mentioned [Papyrus](https://github.com/aleylara/Papyrus), which could be a potential starting point.

[...]

PR #2436, opened 2018-11-09
Make notmyidea theme more "mobile friendly"

ajorg:
Adds a viewport specification and makes most of the `width` properties
relative, with the old values set as max-width. Also adds a max-width of
100% to img so that images that are too wide for a mobile screen will be
shrunk to fit.

This isn't perfect - on my own blog it wraps the `SITESUBTITLE` awkwardly over the `SITENAME`, and the banner doesn't wrap. Also text goes right to the edge - it needs a little margin. But it passes Google's [Mobile-Friendly Test](https://search.google.com/test/mobile-friendly) which should improve page rank, and it improves readability significantly on mobile devices without changing the appearance (that I can tell) from desktop browsers.

Mostly addresses #1704, but doesn't quite hit the bar set for #1280.

### DIFF
--- a/pelican/tests/output/basic/a-markdown-powered-article.html
+++ b/pelican/tests/output/basic/a-markdown-powered-article.html
@@ -2,6 +2,7 @@
 <html lang="en">
 <head>
         <meta charset="utf-8" />
+        <meta name="viewport" content="width=device-width, initial-scale=1">
         <title>A markdown powered article</title>
         <link rel="stylesheet" href="/theme/css/main.css" />
         <link href="/feeds/all.atom.xml" type="application/atom+xml" rel="alternate" title="A Pelican Blog Atom Feed" />
--- a/pelican/tests/output/basic/archives.html
+++ b/pelican/tests/output/basic/archives.html
@@ -2,6 +2,7 @@
 <html lang="en">
 <head>
         <meta charset="utf-8" />
+        <meta name="viewport" content="width=device-width, initial-scale=1">
         <title>A Pelican Blog</title>
         <link rel="stylesheet" href="/theme/css/main.css" />
         <link href="/feeds/all.atom.xml" type="application/atom+xml" rel="alternate" title="A Pelican Blog Atom Feed" />
--- a/pelican/tests/output/basic/article-1.html
+++ b/pelican/tests/output/basic/article-1.html
@@ -2,6 +2,7 @@
 <html lang="en">
 <head>
         <meta charset="utf-8" />
+        <meta name="viewport" content="width=device-width, initial-scale=1">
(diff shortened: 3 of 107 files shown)


avaris:
You need to [regenerate test outputs](http://docs.getpelican.com/en/stable/contribute.html#running-the-test-suite).

PS: Running test locally before pushing helps a lot.

ajorg:
Woops, yes, I'll do that.

ajorg:
Anyone knows how to make the banner wrap properly?

ajorg:
Ah dang, updated something and forgot to update the test output again. Ugh.

justinmayer:
@ajorg: Thank you for the contribution. This has been a long-requested enhancement. Any comments on this from @getpelican/reviewers before it is merged?

ajorg:
There are definitely some things still not quite right about it. I'm eager to hear some feedback from someone who has a better grasp of CSS than I.

Google even notified me that it's still not quite right. Weird because if you test it manually with the tool they tell you to test with, it will tell you it's great, but when their bot crawls it, it still finds problems.

It's a good start though, and demonstrates that good enough is not far away. It's worth spending more time on.

justinmayer:
Many thanks for your work on this, Andrew. I took some time to review this PR's generated output on mobile Safari on iOS 12.1, and I have to agree that something isn't quite right. The index view only looks slightly different than current master, and the article view appears "zoomed-in", with no way to view the entire page-width in the viewport. (I can post screenshots later if that's helpful.)

Given that further iteration will be needed before this endeavor can be integrated, I'm deferring this until a subsequent release. I like the idea of improving mobile support, however, and look forward to more progress on this topic.

justinmayer:
Hi Andrew. Just checking in regarding your progress on this pull request. Do you think you could take another look and see what improvements might be possible so we can get your work merged?

ajorg:
Hi Justin,

I don't think I have bandwidth to learn what I need to learn about CSS to make this work correctly. Perhaps someone who already has that knowledge could pick this up?

BenSturmfels:
I'd be happy to pick this one up. I've got a set of changes that are very similar, but also deal with the subtitle and such. Am I best to create a separate merge request?

ajorg:
> I'd be happy to pick this one up. I've got a set of changes that are very similar, but also deal with the subtitle and such. Am I best to create a separate merge request?

That works for me. I'm unlikely to pick it back up myself at this point. Thank you!

BenSturmfels:
I've now added #2914 which should supercede this pull request. This one can be closed now. Thanks for the head-start @ajorg!

justinmayer:
This has been addressed by #2914.

PR #2437, opened 2018-11-09
Replace %s rather than fall back to defaults

ajorg:
3a0add4b6e66fe08e9f5710f98235491c09e4f81 caused existing configs to fall back to defaults. But since we know exactly how to fix the user config so that the behavior doesn't change, we should do so, while still warning that use of `%s` is deprecated.

Also fixes a bug where we tried to look for `%s` in `None`.

### DIFF
--- a/pelican/settings.py
+++ b/pelican/settings.py
@@ -245,6 +245,22 @@ def get_jinja_environment(settings):
     return settings
 
 
+def _printf_s_to_format_field(printf_string, format_field):
+    """Tries to replace %s with {format_field} in the provided printf_string.
+    Raises ValueError in case of failure.
+    """
+    TEST_STRING = 'PELICAN_PRINTF_S_DEPRECATION'
+    expected = printf_string % TEST_STRING
+
+    result = printf_string.replace('{', '{{').replace('}', '}}') \
+        % '{{{}}}'.format(format_field)
+    if result.format(**{format_field: TEST_STRING}) != expected:
+        raise ValueError('Failed to safely replace %s with {{{}}}'.format(
+            format_field))
+
+    return result
+
+
 def handle_deprecated_settings(settings):
     """Converts deprecated settings and issues warnings. Issues an exception
     if both old and new setting is specified.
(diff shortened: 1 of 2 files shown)


avaris:
See: https://github.com/getpelican/pelican/pull/2383#issuecomment-426409071

ajorg:
> See: [#2383 (comment)](https://github.com/getpelican/pelican/pull/2383#issuecomment-426409071)

Thanks @avaris, that makes sense. I've pushed an update to address it.

pelican/settings.py
```diff
@@ -389,21 +405,31 @@ def handle_deprecated_settings(settings):
...
         for key in ['AUTHOR_FEED_ATOM',
                     'AUTHOR_FEED_RSS',
                     'CATEGORY_FEED_ATOM',
                     'CATEGORY_FEED_RSS',
                     'TAG_FEED_ATOM',
                     'TAG_FEED_RSS',
                     ]:
-            if key in settings and '%s' in settings[key]:
-                logger.warning('%%s usage in %s is deprecated, use {slug} '
-                               'instead. Falling back to default.', key)
```
avaris: You removed the deprecation warning.

pelican/settings.py
```diff
@@ -245,6 +245,22 @@ def get_jinja_environment(settings):
...
 
+def _printf_s_to_format_field(printf_string, format_field):
+    """Tries to replace %s with {format_field} in the provided printf_string.
+    Raises ValueError in case of failure.
+    """
+    TEST_STRING = 'PELICAN_PRINTF_S_DEPRECATION'
+    expected = printf_string % TEST_STRING
+
+    result = printf_string.replace('{', '{{').replace('}', '}}') \
+        % '{{{}}}'.format(format_field)
```
avaris: I think it's cleaner and more understandable to pass `format_field` with `{}`. Then you won't need this and it makes a bit more sense in my opinion:
```python
_printf_s_to_format_field(foo, '{slug}')
# instead of
_printf_s_to_format_field(foo, 'slug')
```

avaris: Oh, that's true. This is better in that case.

justinmayer:
Does the deprecation warning need to be added back to this PR? If there's anything left to do here, we only have about 24 hours for it to be included in the next release.

ajorg:
> @ajorg: In that case, could you please rebase on current master, resolve any merge conflicts, and [squash into a single commit](https://github.com/getpelican/pelican/wiki/Git-Tips#squashing-commits)?

Done so. Thanks!

justinmayer:
Many thanks to @ajorg for the contribution and to @avaris for reviewing!

PR #2438, opened 2018-11-10
Gracefully handle reST parse errors

mosra:
Contains the solution suggested by @avaris in https://github.com/getpelican/pelican/pull/2047#issuecomment-430757553 together with a test, so we ensure neither the hard exit described in #1759 nor the false success described in #1224 ever happens again.

I was struggling a bit where to put the test because the actual conversion of an exception to an `ERROR` is done in `pelican.generators` and not in the reST reader class, hope that's okay.

### DIFF
--- a/pelican/log.py
+++ b/pelican/log.py
@@ -24,6 +24,9 @@ class BaseFormatter(logging.Formatter):
         record.__dict__['customlevelname'] = customlevel
         # format multiline messages 'nicely' to make it clear they are together
         record.msg = record.msg.replace('\n', '\n  | ')
+        record.args = tuple(arg.replace('\n', '\n  | ') if
+                            isinstance(arg, six.string_types) else
+                            arg for arg in record.args)
         return super(BaseFormatter, self).format(record)
 
     def formatException(self, ei):
--- a/pelican/readers.py
+++ b/pelican/readers.py
@@ -12,6 +12,7 @@ import docutils.io
 from docutils.writers.html4css1 import HTMLTranslator, Writer
 
 import six
+from six import StringIO
 from six.moves.html_parser import HTMLParser
 
 from pelican import rstdirectives  # NOQA
@@ -256,7 +257,9 @@ class RstReader(BaseReader):
                         'syntax_highlight': 'short',
                         'input_encoding': 'utf-8',
                         'language_code': self.settings.get('DEFAULT_LANG'),
-                        'exit_status_level': 2,
+                        'halt_level': 2,
+                        'traceback': True,
+                        'warning_stream': StringIO(),
                         'embed_stylesheet': False}
         user_params = self.settings.get('DOCUTILS_SETTINGS')
         if user_params:
@@ -269,7 +272,7 @@ class RstReader(BaseReader):
         pub.set_components('standalone', 'restructuredtext', 'html')
         pub.process_programmatic_settings(None, extra_params, None)
         pub.set_source(source_path=source_path)
-        pub.publish(enable_exit_status=True)
+        pub.publish()
         return pub
 
     def read(self, source_path):
--- a/pelican/tests/parse_error/parse_error.rst
+++ b/pelican/tests/parse_error/parse_error.rst
@@ -1,0 +1,4 @@
+Page with a parse error
+#############
+
+The underline is too short.
--- a/pelican/tests/test_pelican.py
+++ b/pelican/tests/test_pelican.py
@@ -225,3 +225,18 @@ class TestPelican(LoggedTestCase):
             count=1,
             msg="MD_EXTENSIONS is deprecated use MARKDOWN instead.",
             level=logging.WARNING)
+
+    def test_parse_errors(self):
+        # Verify that just an error is printed and the application doesn't
+        # abort, exit or something.
+        settings = read_settings(path=None, override={
+            'PATH': os.path.abspath(os.path.join(CURRENT_DIR, 'parse_error')),
+            'OUTPUT_PATH': self.temp_path,
+            'CACHE_PATH': self.temp_cache,
+        })
+        pelican = Pelican(settings=settings)
+        mute(True)(pelican.run)()
+        self.assertLogCountEqual(
+            count=1,
+            msg="Could not process .*parse_error.rst",
+            level=logging.ERROR)
--- a/pelican/tests/test_readers.py
+++ b/pelican/tests/test_readers.py
@@ -410,6 +410,12 @@ class RstReaderTest(ReaderTest):
         self.assertEqual(tuple_date.metadata['date'],
                          string_date.metadata['date'])
 
+    def test_parse_error(self):
+        # Verify that it raises an Exception, not nothing and not SystemExit or
+        # some such
+        with six.assertRaisesRegex(self, Exception, "underline too short"):
+            self.read_file(path='../parse_error/parse_error.rst')
+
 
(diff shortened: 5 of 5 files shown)


pelican/tests/test_generators.py
```diff
@@ -174,6 +176,25 @@ def test_bad_prefix(self):
                           '!UNKNOWN/authors')
 
 
+class TestParseErrors(LoggedTestCase):
+
+    def test(self):
+        # Verify that just an error is printed and the application doesn't
+        # abort, exit or something. Testing here and not in the reST reader
+        # test as only the page/article generators have the final exception
+        # catching routine.
```
avaris: Given that you expect an `Exception` and nothing else, you can test using `assertRaises`.

mosra: I wanted to put it here so it actually verifies that a logger error comes out (and not that the exception is uncaught, like it was before).

Using `assertRaises()` in `test_readers.py` was my first attempt, but that ultimately felt "not good enough", as it doesn't test anything about the exception-to-error conversion in the generators and so it could still result in the exception causing an abort in case the code in `generators.py` changes in some way.

(But if you think `assertRaises()` in `test_readers.py` is better, sure, I can go back to that.)

avaris: Ah, OK. The comment made it look like you *had to* test like this. This looks somewhat out of place currently :).

How about this:
* One test explicitly in `RstReaderTest`
* And... 
    * We have a few "full" `PageGenerator` tests down below (Like [this](https://github.com/getpelican/pelican/blob/c3aaa97a13dae32aa0d13e78df94ccf25045adb1/pelican/tests/test_generators.py#L635)). Place the page somewhere this can pick up and assert for the error log in addition to others.
    * Or... alternatively. You move this to `test_pelican`, which does similar log counts already.

mosra: Done :)

justinmayer:
@mosra & @avaris: Many thanks for all your work on this. Much appreciated!

cpitclaudel:
This PR makes turns RST warnings into errors and interrupts processing for the corresponding article; wouldn't it work better to turn RST warnings into Pelican warnings and continue processing?

justinmayer:
To what end? Presumably you’d want to fix the problem and re-generate. If you’re going to suggest continuing anyway is “better,” I think it’s incumbent upon you to elaborate and explain why and how, in detail. 😉

cpitclaudel:
> If you’re going to suggest continuing anyway is “better,” I think it’s incumbent upon you to elaborate and explain why and how, in detail. wink

Challenge accepted :) Does "for consistency", aka "for whichever reason Pelican continues running after issuing its own warnings."?

Currently if a post is missing a title Pelican will issue a warning but won't stop processing the post, while if one of the titles in the post doesn't have the right underlines it will issue a warning *and stop*.

Other reasons: 

- The formatting is odd: `ERROR: <file-name> SystemMessage: <file-name> (WARNING/2) …`
- The current behavior is inconsistent with what other reST systems like Sphinx or rst2html, so it will break the assumptions of extensions writers (warnings are not supposed to be fatal by default)
- There's no easy way in reST to selectively disable warnings, so false positives are annoying to deal with;
- Currently Pelican stops after a single warning, so I usually run rst2pseudoxml to get all warnings first and then feed the fixed-up code to pelican (seeing multiple warnings at once is more convenient than seeing them one by one)
- It makes editing documents with warnings easier (for example when I'm editing a post in devserver mode I often have warnings that I intend to fix later but not right away).

Thanks for the hard work; Pelican is a very neat piece of software.

cpitclaudel:
Oh, I guess I should add:

- It's not so easy to find out how to make warnings non-fatal, currently (my first intuition was to pass `--fatal errors`, which of course didn't work)

avaris:
> Currently if a post is missing a title Pelican will issue a warning but won't stop processing the post

Can you explain this? If a post is missing title, a warning is after it is read and that content *is skipped*. I'm not sure what "won't stop processing" means in this context.

> The formatting is odd: ERROR: <file-name> SystemMessage: <file-name> (WARNING/2) …

That's what `docutils` provide as error while parsing, so that's what `pelican` shows. `pelican` does not generate that.

--

`--fatal errors` means `pelican` will *stop* after it encounters an error, like the one you'd get from RST parsing. Normally, these are not fatal. They are logged, skipped and `pelican` will go on with other content. Unless you're giving that, errors in a single RST file *should not* stop `pelican`. If that's the case, that would be bug and it would be nice to have an example reproducing that.

avaris:
Also, I have to add, this PR makes RST warnings *less* obtrusive if anything. Previously, they were much more fatal than this :).

avaris:
Oh, maybe I misunderstood. Wording is confusing :).

Currently, when a warning in RST is encountered that is treated as error and `docutils` stops processing that file. And you want to see all warnings from that file (if there are multiple)? That's reasonable.

For what is worth, you can achieve this with configuration:

```python
DOCUTILS_SETTINGS = {
    'halt_level': 4,
    'warning_stream': None}
```

This should (in theory) print out all the warnings[*] and should keep parsing the file unless a serious error happens.

[*] They would be direct `docutils` warnings and will not be not as nicely formatted, but :man_shrugging:

cpitclaudel:
> If a post is missing title, a warning is after it is read and that content is skipped. I'm not sure what "won't stop processing" means in this context.

I'm not sure what "skipped" means, so sorry if I misunderstood.  What I mean is that when a post contains a pelican warning (like when it's missing a title) an HTML page is still generated.  On the other hand, if there's a reST warning, even a trivial one, no HTML is generated.

> They are logged, skipped and pelican will go on with other content. Unless you're giving that, errors in a single RST file should not stop pelican

Indeed, reST warnings do not stop Pelican; but they stop the generation of an HTML file for the content that has the reST warning.

> That's what docutils provides as error while parsing, so that's what pelican shows. pelican does not generate that.

Well, it's Pelican that adds the `ERROR` header to a message that then says `WARNING` :)

> Currently, when a warning in RST is encountered that is treated as error and docutils stops processing that file. And you want to see all warnings from that file (if there are multiple)? That's reasonable.

Exactly right.  Also, I want to see the generated HTML.

> For what it's worth, you can achieve this with configuration:

Yup, that's what I do currently.

> Also, I have to add, this PR makes RST warnings less obtrusive if anything. Previously, they were much more fatal than this :).

Of course; this PR is great. I'm just suggesting a further improvement.

avaris:
> I'm not sure what "skipped" means, so sorry if I misunderstood. What I mean is that when a post contains a pelican warning (like when it's missing a title) an HTML page is still generated. On the other hand, if there's a reST warning, even a trivial one, no HTML is generated.

I see. I tested and yes, you get a title with empty string (`''`), which produces an `output/.html` file... That shouldn't be valid and that seems like an oversight. It should be an error. Same as you'd get if you define a Markdown without title.

```
ERROR: Skipping /path/to/article.rst: could not find information about 'title'
```

cpitclaudel:
> It should be an error. 

Not always, for example not if `FILENAME_METADATA = r'(?P<slug>.*)'`

> Would you like to submit a PR? :)

I'd love to.  I can't promise anything in the next few weeks though, I'm swamped :/

avaris:
> Not always, for example not if `FILENAME_METADATA = r'(?P<slug>.*)'`

Sure, it doesn't have to be in the `.rst` file and can be defined in other ways. But your example doesn't set `title`, only sets `slug`. `title` is required. You would/should still get error.

PR #2243, opened 2017-10-31
Refactor URLWrapper subclasses to make slugs settable again (#2242)

zackw:
URLWrapper slugs are intended to be settable, but because all of the concrete URLWrapper subclasses override the _getter_ for .slug using `@property`, the setter in the superclass becomes inaccessible.  Fix this by refactoring out the code that the subclasses actually need to override to a normal helper method.

Please note the `???` comment in Author._slug_substitutions; is that another bug that should be fixed while we're in here?

PR #2440, opened 2018-11-11
Make URLWrapper.slug settable again

### LINKED ISSUE #2242: Category/Tag/Author slugs are not settable
URLWrapper objects have a setter for their 'slug' property, but all of the concrete URLWrapper subclasses override the _getter_ for 'slug', which, because of the way Python's property accessors work, makes the setter inaccessible.  This breaks the 'category_meta' plugin and probably other things as well.

avaris:
Fixes #2242, Supersedes #2243

### DIFF
--- a/pelican/urlwrappers.py
+++ b/pelican/urlwrappers.py
@@ -36,9 +36,17 @@ class URLWrapper(object):
     @property
     def slug(self):
         if self._slug is None:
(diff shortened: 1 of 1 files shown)


justinmayer:
Many thanks for taking care of this, @avaris! 😁👍

PR #2446, opened 2018-11-15
pelican-quickstart - auto generate requirements.txt

stuaxo:


stuaxo:
I'll see if there's something reasonable I can add to the quickstart doc instead.

PR #2449, opened 2018-11-15
Make importer work again. Fixes #2448

### LINKED ISSUE #2448: pelican-importer: pandoc check **eats all content** before it is written
fields2pelican is passed a generator called 'fields'
https://github.com/getpelican/pelican/blob/master/pelican/tools/pelican_import.py#L768

The check to see if pandoc is needed uses the 'in' operator consuming this
https://github.com/getpelican/pelican/blob/master/pelican/tools/pelican_import.py#L744

The loop afterwards is passed the now empty generator and so won't write any files
https://github.com/getpelican/pelican/blob/master/pelican/tools/pelican_import.py#L784-L785

stuaxo:
pelican-import: Move pandoc check inside loop, fixing #2448

### DIFF
--- a/pelican/tests/test_importer.py
+++ b/pelican/tests/test_importer.py
@@ -44,7 +44,7 @@ class TestBloggerXmlImporter(unittest.TestCase):
     def setUp(self):
         self.old_locale = locale.setlocale(locale.LC_ALL)
         locale.setlocale(locale.LC_ALL, str('C'))
-        self.posts = list(blogger2fields(BLOGGER_XML_SAMPLE))
+        self.posts = blogger2fields(BLOGGER_XML_SAMPLE)
 
     def tearDown(self):
         locale.setlocale(locale.LC_ALL, self.old_locale)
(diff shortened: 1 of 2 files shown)


avaris:
My only issue is that this will quit in the middle of conversion. Perhaps it's better to warn the user that `pandoc` required to convert a post, skip that content and continue.

stuaxo:
I changed this to output a list of files that require pandoc.

It doesn't run though, because there is some code in fields2pelican that fails when pandoc is not present -

https://github.com/getpelican/pelican/blob/721b18310ec62060b04e2db73d6e979373c08f8b/pelican/tools/pelican_import.py#L828-L851

Running this without pandoc installed:
 - checks pandoc_version, which is an empty tuple here.
 - doesn't handle this condition and builds ```cmd``` containing pandoc anyway
 - runs it
 - gets errorlevel > 0 and then exits

stuaxo:
OK, I've implemented that - here is some sample output when running without Pandoc.

```
ERROR: Pandoc must be installed to import the following posts:
  |   about
  |   hello-interwebs
```

justinmayer:
Many thanks to @stuaxo for the fix and to @avaris for reviewing. 😁

PR #2451, opened 2018-11-16
Handle PORT in Makefile for serve-global

msoucy:


justinmayer:
Thanks for the contribution, Matt. Given that there were merge conflicts, I added similar functionality via e7ef5466. Feel free to submit another PR if you feel there is more work to be done here.

PR #2452, opened 2018-11-16
Importer: Avoid downloading duplicate post attachments

stuaxo:
When importing from wordpress it's possible to have posts with the same attachment multiple times, this creates a set per post of attachments avoiding this.

### DIFF
--- a/pelican/tests/test_importer.py
+++ b/pelican/tests/test_importer.py
@@ -417,20 +417,22 @@ class TestWordpressXMLAttachements(unittest.TestCase):
         self.assertTrue(self.attachments)
         for post in self.attachments.keys():
             if post is None:
-                expected = ('https://upload.wikimedia.org/wikipedia/commons/'
-                            'thumb/2/2c/Pelican_lakes_entrance02.jpg/'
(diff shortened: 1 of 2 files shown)


pelican/tests/test_importer.py
```diff
@@ -420,17 +420,17 @@ def test_attachments_associated_with_correct_post(self):
                 expected = ('https://upload.wikimedia.org/wikipedia/commons/'
                             'thumb/2/2c/Pelican_lakes_entrance02.jpg/'
                             '240px-Pelican_lakes_entrance02.jpg')
-                self.assertEqual(self.attachments[post][0], expected)
+                self.assertEqual(self.attachments[post], {expected})
```
oulenz: Minor quibble: since the variable is called `expected`, I would prefer if it already were the expected set, i.e. `expected = {'https://...'}`.

oulenz:
Looks good to me

pelican/tests/test_importer.py
stuaxo: Sorted

justinmayer:
Many thanks to @stuaxo for the fix and to @oulenz for reviewing. 😄

Issue #2444, opened 2018-11-14
All static files copied ontop of eachother


dns2utf8:
Hi all

I upgraded to 4.0.0 today and noticed that all my static files are processed as "draft":
```
...
-> Copying /home/user/website/content/mindtuxmap/img/lang/ch.png to drafts/pages/.html
-> Copying /home/user/website/content/images/breeze_viewpdf.svg to drafts/pages/.html
...
```

Do I have to configure something special? It worked like a charm with 3.7.1.

Cheers

dns2utf8:
I noticed, I can include png's or pdf's if I use `{attach}` instead of `{static}`.
I was under the impression that this should only affect the path on the generated site not the resolution itself, is this a bug?

Best

justinmayer:
Hi Stefan. I believe @oulenz, and maybe @mosra, made some improvements for Pelican 4.0 that might relate to the issue you described. Perhaps they might have an idea what's causing that.

Keep in mind that they may need more information than you've provided. As mentioned in the [How to Get Help](http://docs.getpelican.com/en/stable/contribute.html#how-to-get-help) section of the documentation, it's hard to troubleshoot without knowing more about your environment (settings file contents, et cetera).

oulenz:
I am traveling without access to a computer, tomorrow evening is the
earliest I can look at this. I wonder whether it might be because the
Static and Page types inherit from each other.

Den tor. 15. nov. 2018, 13.40 skrev Justin Mayer <notifications@github.com:

dns2utf8:
Okay, what do you need to know? I am building with a gitlab runner inside a debian docker image.
I will try to use the git version tomorrow.

Currently I solved it by pinning the version to 3.7.1 for the CI.
So it is not urgent. @oulenz if you are on vacation, please enjoy it.

avaris:
A repo that reproduces the problem would be ideal.

oulenz:
@dns2utf8 My theory is that your settings include

```python
DEFAULT_METADATA = {
    'status': 'draft',
}
```

...or something similar to the effect that your pages have default draft status. Is this correct?

If yes, the problem is that because the `Static` class inherits from `Page`, your static content takes the metadata values of a draft page. I've written a PR that fixes that.

Otherwise we have to keep looking, but the PR is probably still pertinent.

bryanbrattlof:
I think @oulenz has it.

I've just ran into this issue myself, and it does look like the static content is being handed over to the draft page.

[...]

PR #2455, opened 2018-11-16
Disinherit Page from Static so default draft status does not affect save_as

oulenz:
Fixes https://github.com/getpelican/pelican/issues/2444 (hopefully)

Hitherto, the `Static` class inherited from the `Page` class. In particular it inherited its `_expand_settings` method, meaning its `save_as` (and friends) metadata values would be determined by the `DRAFT_PAGE_SAVE_AS` setting (and friends) if `DEFAULT_METADATA` included `'status': 'draft'`. I've changed that by letting `Static` inherit from `Content` directly.

Is there anything that relied on `Static` inheriting from `Page` that the tests don't pick up on?

(I'm assuming we don't want draft static content. Still, I've not set `allowed_statuses = ('published',)` because at the moment, default metadata does still affect static content. Then again, `is_valid` isn't actually called when static content is generated, so `allowed_statuses` wouldn't have been tested for anyway.)

### DIFF
--- a/pelican/contents.py
+++ b/pelican/contents.py
@@ -527,7 +527,11 @@ class Article(Content):
 
 
 @python_2_unicode_compatible
-class Static(Page):
+class Static(Content):
+    mandatory_properties = ('title',)
+    default_status = 'published'
+    default_template = None
+
     def __init__(self, *args, **kwargs):
         super(Static, self).__init__(*args, **kwargs)
         self._output_location_referenced = False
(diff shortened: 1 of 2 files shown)


justinmayer:
@avaris: Any thoughts regarding this pull request?

justinmayer:
Many thanks to @oulenz for the fix and to @avaris for reviewing.

Issue #2456, opened 2018-11-16
ComplexHTTPRequestHandler is doing excessive amount of work for non-complex cases


mosra:
Spinning up a Pelican server after #2324 was merged, every static request prints the following in the log:

```
-> Tried to find `/static/ship.jpg.html`, but it doesn't exist.
-> Tried to find `/static/ship.jpg/index.html`, but it doesn't exist.
-> Tried to find `/static/ship.jpg/`, but it doesn't exist.
-> Found `/static/ship.jpg`.
127.0.0.1 - - [16/Nov/2018 19:16:29] "GET /static/ship.jpg HTTP/1.1" 200 -
```

And in case of pages it's like this:

```
-> Tried to find `/css/components/test.html`, but it doesn't exist.
-> Found `/css/components/test/index.html`.
127.0.0.1 - - [16/Nov/2018 19:16:32] "GET /css/components/test/ HTTP/1.1" 200 -
```

To me that seems quite excessive, since the `GET` request is already containing the correct path and so trying that first (as it was *before* #2324) makes the most sense (to me at least). In other words, I have the layout done in a way that doesn't require any complex handling -- the files are just where a simple `GET` expects them to be.

Seeing the amount of PRs submitted just for this alone I realize this is a mine-field and every fix breaks someone's else workflow. Since I didn't find a clear reason in the PR descirption, my question is, @oulenz (since you submitted the aforementioned PR), what exactly was the reason to not try the `''` suffix first (the ideal case, basically) and what would break if it did try that first?

(Looks like this is my first issue for this project ever, interesting.)

oulenz:
To answer your last question: if you try `''` first, then looking for `'/index.html'` later is pointless because if `'foo/index.html'` exists, then so does `'foo'`.

I am not convinced the 'unnecessary' attempts to get a resource are problematic in terms of performance, especially since this is just a test server used during development.

However, one thing we could do is to simplify to just checking `''` if the requested path already contains an extension. That would only negatively affect directory names with internal `.`, but that's an edge case I would be happy to ignore.

mosra:
> if you try '' first, then looking for '/index.html' later is pointless

Thinking outside of the box: do we actually need to look for `/index.html`? As far as my experience goes with plain `python -m http.server`, it was never a problem. Assuming I have a file at `css/index.html`:

- if I ask for `http://localhost:8000/css`, the log of `python -m http.server` shows

    ```
[...]

PR #2457, opened 2018-11-16
Server: limit logging when resolving get request

oulenz:
Fixes https://github.com/getpelican/pelican/issues/2456 (hopefully)

### DIFF
--- a/pelican/server.py
+++ b/pelican/server.py
@@ -42,8 +42,6 @@ def parse_arguments():
 
 
 class ComplexHTTPRequestHandler(srvmod.SimpleHTTPRequestHandler):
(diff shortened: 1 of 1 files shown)


justinmayer:
Many thanks, @oulenz!

Any comments from @getpelican/reviewers on this proposed fix?

justinmayer:
@oulenz: Do you feel like anything from this PR is still relevant now that the one @avaris submitted has been merged?

mosra:
I *think* this could still reduce the logging on failure (the list of attempted tries would be shorter for paths that are directories), but it's such a rare corner case that I'm not sure is worth the extra complexity.

oulenz:
As far as I am concerned this is now purely cosmetic, so I'll close.

Issue #2441, opened 2018-11-12
Generate requirements.txt in site skeleton


stuaxo:
I came back to pelican after over a year and had to look at the docs when setting up a python environment, it would be good if the site skeleton added a requirements.txt

justinmayer:
Hey Stuart. Glad to have you back. That said, I'm confused.. The Quickstart docs make it pretty clear what's needed to install Pelican, along with Markdown support if needed. What exactly is it that you lacked?

Also, not that I think it's particularly relevant, but... https://github.com/getpelican/pelican/tree/master/requirements

avaris:
Requirements for what...?

stuaxo:
requirements.txt for running pelican for somebodies site.

So, just the same requirements as the docs, mine looks like this:

```
beautifulsoup4==4.6.3
fabric==1.14.0
feedparser==5.2.1
lxml==4.2.5
Markdown==3.0.1
pandoc==1.0.2
pelican==3.7.1
```

Obviously markdown and some of the others might not be known to start with (but maybe can be worked out during the question/answer part of setting up the site).


Since my free time is in chunks with long gaps, I like not having to think too much and having a requirements.txt auto-added would have helped.

I could probably have a look at adding something like this, but I'm not sure where to start.

justinmayer:
After installing Pelican, run:

    pip freeze > requirements.txt

Does that not more-or-less accomplish your goal? Or am I still missing something?

stuaxo:
It definitely would do it, however for whatever reason I didn't do it at the time ... possibly, because my brain turns off a bit when doing a tutorial or quickstart.

Here is a PR that generates a requirements.txt in the project skeleton

https://github.com/getpelican/pelican/pull/2445

I'd like to think its presence would have reminded me to run pip freeze after installing markdown, so I'm on the fence about adding that bit to the docs.

LouisJackman:
`pelican-quickstart` gives a good starting point and already assumes `pelican` is installed. However, having dependencies specified in the generated scaffolding, whether it's `pyproject.toml` or `requirements.txt`, could still be useful for workflows like `git pull`ing a website to a new development machine and getting the whole thing working with a `.venv/bin/activate && pip3 install -r requirements.txt` invocation or the `pyproject.toml` equivalent. Python developers might reasonably expect that to work in a site generated by a Python site generator.

stuaxo:
[...]

PR #2461, opened 2018-11-19
Update quickstart.rst to add requirements.txt generation.

stuaxo:
Here's a docs solution to #2441 - I don't have a strong opinion if this is goes in or not though.


justinmayer:
@stuaxo: Given the concerns that @avaris raised in #2441, as well as the fact that some folks might prefer the newer `pyproject.toml` format instead of `requirements.txt` files, I'm inclined not to include this in Pelican at this time. Thanks for your understanding.

Issue #2467, opened 2018-11-20
Bad MD file but DEBUG doesn't tell me which file the WARNING is for...


egberts:
In effort to chase down the error message, it is painful to rely on Python debugger just to find out where the error occur at:

```
DEBUG: Read file pages/404.md -> Page
DEBUG: Read file pages/uses.md -> Page
WARNING: There is no "date" attribute in the item metadata. Defaulting to slug order.
DEBUG: Read file extra/favicon.ico -> Static
```
Please note that the file 'page/about.md' is missing in above output.

The above error message was generated using:

```bash
pelican -v -v -v -v -D -s pelicanconf.py
```
And examining the `pages/uses.md` file shows that it has a perfectly filled out 'Date:' value.

```markdown
    Title: uses
    Date: 2018-11-20 12:51
   ...
```

After much debugging of Python, I find out that the `about.md` was missing a `Date:` metadata tag and its value.  And nothing to do with that `uses.md` file.

Notice that `about.md` file name was not mentioned in the terminal output?

Shouldn't that debug message be printing  "DEBUG: Reading file pages/about.md" BEFORE the warning message.

bryanbrattlof:
Hi @egberts, I'm sure someone with more experience than me will chime in here, but I had a few things pop into my head when I read this. 

I know we are talking about having the logs elaborate more, but would using a `grep` command work to find the problem files?
```bash
grep -Lr "Date:" path/to/your/content | grep ".md$"
```

If not, then someone with more knowledge of the code might have some better ideas.

My suggestion would be to loop through the `AttributeError` in `utils.ordered_content` until we find the Pages/Articles in the `content_list` that causes the problem
```python
...

try:
    content_list.sort(key=attrgetter(order_by),
                      reverse=order_reversed)
except AttributeError:
    for content in content_list:
        if hasattr(content, order_by):
            continue
        logger.warning(
            "There is no '{attr}' attribute in the '{path}' "
            "file. Defaulting to slug order.".format(
                attr=order_by,
                path=content.relative_dir
        ))
...

```

This would let all the pages that are missing the `order-by` attribute emit a `WARNING:`, but this might be a little noisy for some.
```bash
WARNING: There is no 'date' attribute in the 'pages/first-page' file. Defaulting to slug order.
WARNING: There is no 'date' attribute in the 'pages/second-page' file. Defaulting to slug order.
```

egberts:
[...]

PR #2473, opened 2018-11-24
List files missing tag needed to order content

bryanbrattlof:
This should fix: #2467

### DIFF
--- a/pelican/utils.py
+++ b/pelican/utils.py
@@ -763,9 +763,19 @@ def order_content(content_list, order_by='slug'):
                     content_list.sort(key=attrgetter(order_by),
                                       reverse=order_reversed)
                 except AttributeError:
-                    logger.warning(
-                        'There is no "%s" attribute in the item '
-                        'metadata. Defaulting to slug order.', order_by)
+                    for content in content_list:
+                        try:
+                            getattr(content, order_by)
+                        except AttributeError:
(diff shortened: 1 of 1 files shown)


pelican/utils.py
```diff
@@ -763,9 +763,18 @@ def order_content(content_list, order_by='slug'):
                     content_list.sort(key=attrgetter(order_by),
                                       reverse=order_reversed)
                 except AttributeError:
-                    logger.warning(
-                        'There is no "%s" attribute in the item '
-                        'metadata. Defaulting to slug order.', order_by)
+                    for content in content_list:
+                        if getattr(content, order_by, None) is not None:
```
avaris: This assumes `getattr(content, order_by)` will not be `None`.

bryanbrattlof:
Thanks @avaris, I was trying to avoid having to indent so much, and assumed that everything we sort by would return something other than `None`.

I changed to a `try/except AttributeError` to get rid of the assumption.

avaris:
> assumed that everything we sort by would return something other than None.

Generally true, for the core at least, but with plugins and such you never know :).

justinmayer:
Many thanks to @bryanbrattlof for the enhancement and to @avaris for the review. Sincere apologies for the long delay in getting to this, folks.

PR #2476, opened 2018-11-24
Try importing from collections.abc for compatibility with Python 3.8

oulenz:
This changes the remaining imports not covered by https://github.com/getpelican/pelican/pull/2389. The try/catch is necessary as long as we support Python 2.7.

### DIFF
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -2,7 +2,10 @@
 from __future__ import print_function, unicode_literals
 
 import argparse
-import collections
+try:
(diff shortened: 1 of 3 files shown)


Hanaasagi:
I think there is another need to be modified.

https://github.com/getpelican/pelican/blob/e967988eff613ec6cdb74b3f0140d3937322c059/pelican/__init__.py#L245

justinmayer:
Many thanks to @oulenz for the compatibility fix and to @avaris for reviewing.

Issue #2470, opened 2018-11-22
Typo of Article Markdown 'Tag:' (should be 'Tags:') gives cryptic error message


egberts:
Using an `articles/example.md` file, when mistyping in the metadata `Tag:` (instead of `Tags:) resulted in a cryptic error message as shown:

```erlang
CRITICAL: AttributeError: 'unicode' object has no attribute 'slug'
```

Probably should say something like:
```erlang
CRITICAL: 'Tags:' metadata is missing from "example.md" file
```

# Crash dump
```python
CRITICAL: AttributeError: 'unicode' object has no attribute 'slug'
Traceback (most recent call last):
  File "/usr/local/bin/pelican", line 11, in <module>
    load_entry_point('pelican==4.0.0', 'console_scripts', 'pelican')()
  File "/usr/local/lib/python2.7/dist-packages/pelican-4.0.0-py2.7.egg/pelican/__init__.py", line 623, in main
    pelican.run()
  File "/usr/local/lib/python2.7/dist-packages/pelican-4.0.0-py2.7.egg/pelican/__init__.py", line 178, in run
    p.generate_context()
  File "/usr/local/lib/python2.7/dist-packages/pelican-4.0.0-py2.7.egg/pelican/generators.py", line 633, in generate_context
    if not article.is_valid():
  File "/usr/local/lib/python2.7/dist-packages/pelican-4.0.0-py2.7.egg/pelican/contents.py", line 202, in is_valid
    self._has_valid_save_as(),
  File "/usr/local/lib/python2.7/dist-packages/pelican-4.0.0-py2.7.egg/pelican/contents.py", line 174, in _has_valid_save_as
    sanitised_join(output_path, self.save_as)
  File "/usr/local/lib/python2.7/dist-packages/pelican-4.0.0-py2.7.egg/pelican/contents.py", line 434, in save_as
    return self.get_url_setting('save_as')
  File "/usr/local/lib/python2.7/dist-packages/pelican-4.0.0-py2.7.egg/pelican/contents.py", line 231, in get_url_setting
    return self._expand_settings(key)
  File "/usr/local/lib/python2.7/dist-packages/pelican-4.0.0-py2.7.egg/pelican/contents.py", line 526, in _expand_settings
    return super(Article, self)._expand_settings(key, klass)
  File "/usr/local/lib/python2.7/dist-packages/pelican-4.0.0-py2.7.egg/pelican/contents.py", line 225, in _expand_settings
[...]

PR #2477, opened 2018-11-25
Remove non-existent tag attribute handling from Content.url_format

oulenz:
Fixes https://github.com/getpelican/pelican/issues/2470. Removes code that was added by https://github.com/getpelican/pelican/pull/1926 without motivation and that has no effect other than causing a confusing error message when a user accidentally uses `Tag:` rather than `Tags:` in an article.


justinmayer:
Many thanks to @oulenz for the fix and to @avaris for reviewing. 🎊

PR #2479, opened 2018-11-25
Use language_code if it is supported

### LINKED ISSUE #2475: Unsupported language codes passed to Docutils
Even though https://github.com/getpelican/pelican-plugins/issues/1087 was moved from here to pelican-plugins, if I understand it correctly the original error was due to https://github.com/getpelican/pelican/pull/2256 , and @avaris suggested something may need to be done about it here. (@mosra)

avaris:
Fixes #2475 

Not completely sure about the warning text, open for suggestions :).

### DIFF
--- a/pelican/readers.py
+++ b/pelican/readers.py
@@ -9,6 +9,7 @@ from collections import OrderedDict
 import docutils
 import docutils.core
 import docutils.io
+from docutils.parsers.rst.languages import get_language as get_docutils_lang
 from docutils.writers.html4css1 import HTMLTranslator, Writer
 
 import six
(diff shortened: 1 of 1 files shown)


mosra:
:+1: 

Looks good, tested locally and it behaves as expected. Turns out setting `DEFAULT_LANG` to `th` causes *quite a lot* of issues in my plugins as well (and also in [Pyphen](http://pyphen.org/)), looks like I have some more work to do :)

justinmayer:
Many thanks to @avaris for the fix and to @mosra for reviewing. 🕺

Issue #2482, opened 2018-11-28
ssh-upload command in Makefile is concatenated with rsync-upload


saiwal:
The Makefile generated by default is combining the lines of rsync-upload with that of ssh-upload which needs manual correction. (line 80 of Makefile).

`ssh_upload: publish
	scp -P $(SSH_PORT) -r $(OUTPUTDIR)/* $(SSH_USER)@$(SSH_HOST):$(SSH_TARGET_DIR)rsync_upload: publish
	rsync -e "ssh -p $(SSH_PORT)" -P -rvzc --cvs-exclude --delete $(OUTPUTDIR)/ $(SSH_USER)@$(SSH_HOST):$(SSH_TARGET_DIR)`

which should look like two separate lines

`ssh_upload: publish
	scp -P $(SSH_PORT) -r $(OUTPUTDIR)/* $(SSH_USER)@$(SSH_HOST):$(SSH_TARGET_DIR)`

`rsync_upload: publish
	rsync -e "ssh -p $(SSH_PORT)" -P -rvzc --cvs-exclude --delete $(OUTPUTDIR)/ $(SSH_USER)@$(SSH_HOST):$(SSH_TARGET_DIR)`

PR #2483, opened 2018-11-28
Remove unnecessary whitespace stripping in Makefile template

bryanbrattlof:
This fixes: #2482 for me. It was just some extra white-space controls that didn't need to be in the makefile template.


justinmayer:
Many thanks to @saiwal for reporting, @bryanbrattlof for the fix, and @avaris and @mosra for reviewing. Love the teamwork! 👏

Issue #2484, opened 2018-11-30
Need to insert HTML comment of original content filename


egberts:
This `:Slug:` else `:Title:` makes it harder to determine what the original content file name is.  So, when you want to go back and edit the file, the output file name may not be similar enough with the content filename and the end-user is often forced to use `grep` tool to locate the original/content file.

The objective is to update each page with an HTML comment notation where the original content file is:
```html
<title="slugified title">
<!-- content filename: article/original-filename.md -->
```

What is the variable name that has the original content filename so that it can be written into the base.html in form of

bryanbrattlof:
Hi'ya @egberts, I was just doing this myself. 

This isn't documented anywhere, nor is it setup like the other variables for use in the templates, so I guess it's a *"use at your own risk"* kind of thing, but [```article.get_relative_source_path()```](https://github.com/getpelican/pelican/blob/49c77d7412673c38306fd9a69a073544302148ba/pelican/contents.py#L441) works for me.

It's a part of the base ```Content``` object so ```Page```, ```Article```, and ```Static``` have it when building.

Hope that helps.

egberts:
Yep, that does it, @bryanbrattlof .

I inserted the following to the beginning of my template file, in this example `article.html`:

```html
<!-- articles.html begins -->
<!-- source file: {{ article.get_relative_source_path() }} -->
{% extends "base.html" %}
{% block title %}{{ article.title }}{% endblock %}
```
And got the resultant output file:
```html
<!-- articles.html begins -->
<!-- source file: articles/api-security-checklist.md -->
<!-- base.html begins, seeded by blog/articles/api-security-checklist.html  -->
<!DOCTYPE html>
```

Closing this as solved.

avaris:
There is `article.source_path` which is [documented](http://docs.getpelican.com/en/stable/themes.html#article).

egberts:
I too went with Bryan because I also emulated GitLab's funky top-level directory that results in a longer SITEURL for our blog's top-level directory.

Simple example supported is:

    SITEURL=https://example.net/

The ideal but complex examples that I am trying to get supported are:

    SITEURL=https://example.net/blog
    SITEURL=https://gitlab.com/builds/egberts/<repo>

As a result, I've had to make an extension to the SITEURL:

    SITEURL = 'https://example.net'
    SITE_SUBPATH = 'blog'
    RELATIVE_URLS = True if developing_site else False
    SITE_DIR = ''
[...]

PR #2486, opened 2018-12-01
Add relative_source_path atrtibute to content

bryanbrattlof:
I wanted to see what the consensus was for adding a ```relative_source_path``` attribute to the ```Content``` object was. 

@egberts, and I in #2484 where talking about using an unpublished way to add a relative path as a comment into the generated html for ```page``` and ```article``` for better debugging using continuous integration pipelines like ```gitlab-ci```.

I also saw the PR (#2416) wanted the ability to generate urls to edit his content in github, but it seems like it died there. 

Hope everyone's weekend it going well.


avaris:
This doesn't need to be a property.

avaris:
For debugging purposes, I see very little benefit in this over `source_path` but one can use this to have a link to the source. So, I'm not opposed to it.

egberts:
Awesome!

pedrohdz:
I am open to making the change myself. Just do not want to step on anyone's toes.

I am looking for the ability to add an _edit in github_ button. This would do it.

bryanbrattlof:
Thanks @pedrohdz for the suggestion. I've added the tests. :)

PR #2492, opened 2018-12-12
Remove stopserver from .PHONY prereqs in Makefile template

ixjlyons:
Just a minor oversight from when the devserver functionality was integrated into the pelican command and the stopserver target was removed. This shouldn't really affect much -- I just noticed in shell completion output while upgrading to 4.x :)


justinmayer:
Many thanks to @ixjlyons for the fix and to @avaris for reviewing.

Issue #2487, opened 2018-12-06
On fresh install: "simple" theme build error using publishconf.py


julien-h:
```
❯ pelican --version
4.0.1
```

If I use the "simple" built-in theme:

```
THEME = 'simple'
```

Then the following command works:

```
pelican content
```
But not:

```
pelican content -s publishconf.py
```

Outputs this error:
```
CRITICAL: TypeError: not all arguments converted during string formatting
```

But not if I use the "notmyidea" built-in theme.

EDIT: Same problem with a fresh install using `pelican-quickstart`.

rask004:
Is anyone else able to replicate this? I tried with a quickstarted project but using publishconf.py worked.

Additionally, I cannot find the mentioned error when searching in pelican.

rask004:
@julien-h What plugins are you using? Also, what settings are different between your pelicanconf.py and publishconf.py?

Julian:
Similar issue here, I'm on pelican `4.2.0`, site works fine with the default theme but not simple:

```
⊙  rm -rf output && pelican -t notmyidea .                                                                                                                                                        julian@Air
Done: Processed 32 articles, 0 drafts, 0 pages, 0 hidden pages and 0 draft pages in 2.71 seconds.

~/Development/til is a git repository on master
⊙  rm -rf output && pelican -t simple .                                                                                                                                                           julian@Air
CRITICAL: TypeError: not all arguments converted during string formatting
```

Not sure from looking at #2442 quickly what an end user was supposed to do there, but putting this here anyhow in case anyone else finds this ticket, will try to update it if I figure out where the TypeError comes from and how to fix it.

PR #2494, opened 2018-12-16
Fall back to the old feed settings if using an older theme

bryanbrattlof:
I wanted to take a stab at reducing the upgrade pains for the older themes still in circulation, and hopefully reduce the hair pulling during development.

All this does is catch the ```TypeError``` that ```template.render``` throws when using the new atom/rss feed settings format, and tries to render the template using the old format. 

- If it works, this will emit a warning (not sure I like the wording of it just yet).

- If ```template.render``` fails again, this will catch and emit an error with the stack trace to make absolutely sure the trace is emitted.

### DIFF
--- a/pelican/writers.py
+++ b/pelican/writers.py
@@ -178,13 +178,62 @@ class Writer(object):
             # other stuff, just return for now
             return
 
+        def _old_theme_fix(template, localcontext, path):
+            """
+            To give themes time to update, change atom/feed variables
+            back to using %s and try again. If it fails again return None
+            """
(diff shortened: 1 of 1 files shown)


bryanbrattlof:
Going through the open issues, this should fix #2487

bryanbrattlof:
Hi'ya everyone,

It seems like most of the themes have addressed the problem this PR fixes, (or most people know how to fix them). Either way this doesn't seem to be an issue anymore, so I'm going to close this.

PR #2501, opened 2019-01-02
Fix warning message

jorgesumle:
Add space between words 'static' and 'content'.


justinmayer:
Thanks to @jorgesumle for the fix (and patience), and to @mosra for reviewing.

PR #2503, opened 2019-01-04
Pagination: the lstrip() was way too hungry, eating absolute URLs

mosra:
The PR #2375 by @oulenz replaced stripping of a single leading slash with a call to `lstrip()`. While it might have looked innocent at first, it eats all leading slashes it can find, thus breaking absolute URLs that start with just `//` (i.e., not `https://`, without the scheme). Funnily enough even the test which was meant to verify this was changed and thus the error went unnoticed (and I wasn't around when the PR got reviewed and merged, sigh). I wasn't aware of this until I got a bug report yesterday during a high-traffic spike: https://github.com/mosra/magnum-website/issues/9

This reverts back to how pagination worked for the `{url}` placeholder as I did it in 182fb11c80935f2ef1661beaf1151969a24e833f (PR https://github.com/getpelican/pelican/pull/2235). I also added extra comments to this piece of code (*and* the test) to make it less likely that someone breaks this again in the future ;)

Thanks for merging.

### DIFF
--- a/pelican/paginator.py
+++ b/pelican/paginator.py
@@ -146,7 +146,17 @@ class Page(object):
         }
 
         ret = prop_value.format(**context)
-        ret = ret.lstrip('/')
+        # Remove a single leading slash, if any. This is done for backwards
+        # compatibility reasons. If a leading slash is needed (for URLs
+        # relative to server root or absolute URLs without the scheme such as
+        # //blog.my.site/), it can be worked around by prefixing the pagination
+        # pattern by an additional slash (which then gets removed, preserving
+        # the other slashes). This also means the following code *can't* be
+        # changed to lstrip() because that would remove all leading slashes and
+        # thus make the workaround impossible. See
+        # test_custom_pagination_pattern() for a verification of this.
+        if ret[0] == '/':
+            ret = ret[1:]
         return ret
 
     url = property(functools.partial(_from_settings, key='URL'))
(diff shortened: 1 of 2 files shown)


oulenz:
I apologise for breaking your production environment.

I still don't quite understand the reasoning behind your implementation. If I parse the code correctly, the site in the test has the url `//blog.my.site/`. The first `/` is stripped away, and to compensate for that the `PAGINATION_PATTERN` has `/{url}`. Is that correct? If so, couldn't you compensate for the use of `lstrip` by letting the `PAGINATION_PATTERN` have `//{url}`?

(Or alternatively, we could consider whether we need to strip `/` at all.)

mosra:
`PAGINATION_PATTERN` having `//{url}` (or `///{url}`) wouldn't help -- it'll result in just `{url}` again, since `lstrip()` is called on the result of the template string expansion (if I'm reading the code correctly). Besides that, the url could be `//blog.my.site/` in one case but e.g. `http://another.domain/archives/` in another (convoluted case, I know) and thus having `//{url}` would result in (incorrect) `/http://another.domain/archives/`.

> (Or alternatively, we could consider whether we need to strip / at all.)

I kept the `/` stripping there to avoid breaking existing setups. I didn't know what was the original reason to do it, but it was easy to workaround and so I went with the path of least resistance :)

oulenz:
Ah yes, true.

I think the explanation you added could be clearer. It sort of implies that this automatically preserves double `//`, which of course it can't. What it does do is give users the opportunity to opt-out of the `/` removal by adding a dummy `/` in the `PAGINATION_PATTERN` that takes the heat, so to speak.

I'm leaning towards not documenting this is as a feature under `PAGINATION_PATTERN`, in the hope that we can maybe remove the `/` stripping altogether one day.

mosra:
Heh, looks like I can't speak English today. I reworded the comment, hopefully better now? :)

justinmayer:
Many thanks to @mosra for the fix and to @oulenz for reviewing.

Issue #2442, opened 2018-11-14
pelican is not compatible with jinja2


amane-katagiri:
Pelican 4.0.0 replaces `%s` in feed settings to `{slug}` (on #2432), but `format` filter in Jinja2 2.10 cannot process the formatted representation (`{slug}`). How can I fix it?

# show versions

```
$ pip freeze
blinker==1.4
docutils==0.14
feedgenerator==1.9
Jinja2==2.10
Markdown==3.0.1
MarkupSafe==1.1.0
pelican==4.0.0
pkg-resources==0.0.0
Pygments==2.2.0
python-dateutil==2.7.5
pytz==2018.7
six==1.11.0
Unidecode==1.0.22
```

# make project

```
$ pelican-quickstart
Welcome to pelican-quickstart v4.0.0.

This script will help you create a new Pelican-based website.

Please answer the following questions so this script can generate the files
needed by Pelican.


> Where do you want to create your new web site? [.]
> What will be the title of this web site? HOGE
> Who will be the author of this web site? HOGE
> What will be the default language of this web site? [en] ja
> Do you want to specify a URL prefix? e.g., https://example.com   (Y/n) y
> What is your URL prefix? (see above example; no trailing slash) http://example.com
> Do you want to enable article pagination? (Y/n) n
> What is your time zone? [Europe/Paris] Asia/Tokyo
> Do you want to generate a tasks.py/Makefile to automate generation and publishing? (Y/n) y
> Do you want to upload your website using FTP? (y/N) n
> Do you want to upload your website using SSH? (y/N) n
> Do you want to upload your website using Dropbox? (y/N) n
> Do you want to upload your website using S3? (y/N) n
> Do you want to upload your website using Rackspace Cloud Files? (y/N) n
> Do you want to upload your website using GitHub Pages? (y/N) n
Done. Your new project is available at /dev/shm/hoge
$ echo "THEME = 'simple'" >> pelicanconf.py
$ cat > content/hoge.md <<EOS
> Title: HOGE
> Category: hoge
> Date: 2018-11-14 11:00
> Slug: hoge
>
> HOGE
> EOS
```

# make site

```
$ make publish
pelican /dev/shm/hoge/content -o /dev/shm/hoge/output -s /dev/shm/hoge/publishconf.py -D
(* snip *)
CRITICAL: TypeError: not all arguments converted during string formatting
Traceback (most recent call last):
(* snip *)
[...]

PR #2507, opened 2019-01-05
Fix incorrect update note in changelog

jorgesumle:
See [explanation](https://github.com/getpelican/pelican/issues/2442#issuecomment-446303114)


justinmayer:
Thank you for the fix, Jorge.

PR #2522, opened 2019-01-23
Simplify _input_compat and str_compat

cclauss:
The __six__ module provides us these same constructs using different names.


justinmayer:
Many thanks to @cclauss for the enhancement and to @avaris for reviewing.

Issue #2534, opened 2019-02-22
IndexError when ARTICLE_TRANSLATION_ID


rminderhoud:
Hello, I upgraded my project to Pelican 4.x and I have a very minimal config. It revealed the following error when the `ARTICLE_TRANSLATION_ID` setting is not defined in a config file

```
CRITICAL: IndexError: tuple index out of range
Traceback (most recent call last):
  File "/home/ralph/.local/share/virtualenvs/ram-F3HMn9Ws/bin/pelican", line 11, in <module>
    sys.exit(main())
  File "/home/ralph/.local/share/virtualenvs/ram-F3HMn9Ws/lib/python3.6/site-packages/pelican/__init__.py", line 623, in main
    pelican.run()
  File "/home/ralph/.local/share/virtualenvs/ram-F3HMn9Ws/lib/python3.6/site-packages/pelican/__init__.py", line 178, in run
    p.generate_context()
  File "/home/ralph/.local/share/virtualenvs/ram-F3HMn9Ws/lib/python3.6/site-packages/pelican/generators.py", line 652, in generate_context
    self.articles, self.translations = _process(all_articles)
  File "/home/ralph/.local/share/virtualenvs/ram-F3HMn9Ws/lib/python3.6/site-packages/pelican/generators.py", line 648, in _process
    arts, translation_id=self.settings['ARTICLE_TRANSLATION_ID'])
  File "/home/ralph/.local/share/virtualenvs/ram-F3HMn9Ws/lib/python3.6/site-packages/pelican/utils.py", line 682, in process_translations
    .format(*translation_id).format(*id_vals)
IndexError: tuple index out of range
```

**Version**: Pelican 4.0.1
**Steps**:
1. Remove `ARTICLE_TRANSLATION_ID`  from your pelican file
2. Run `pelican -s conf.py --debug`

oulenz:
It took me a long time to come up with a scenario that would cause that error message on that line, but I think I've got it.

Do you have one or more articles with an empty slug?

rminderhoud:
Ah, yes I do. I have a few articles where I don’t use the slug tag but let
it generate a slug from the file name. I apologize, didn’t realize that was
also part of the issue.

oulenz:
If you don't set `ARTICLE_TRANSLATION_ID`, it defaults to using the slug. So if in fact you're not using translations on your website, you should be able to fix this by setting `ARTICLE_TRANSLATION_ID = None`.

[...]

PR #2535, opened 2019-02-24
Fix bug in process_translations warning message generation that caused error with empty slugs

oulenz:
Fixes https://github.com/getpelican/pelican/issues/2534

Instances of `operator.attrgetter` return a tuple if initialised with multiple arguments, but not if initialised with just one argument. Nevertheless, we would unpack that return value into a `format` call on our warning message. With most strings, this would just cause malformed warning messages as the first character was inserted and the rest ignored. But with empty strings, it caused a tuple index out of range error.


justinmayer:
Thank you for the fix, @oulenz.

Issue #2489, opened 2018-12-07
Determine where TypeError: not all arguments converted during string formatting is occuring?


AWegnerGitHub:
With Pelican 4.0 there was a change made regarding position argument formatting: 

> Some user-submitted themes use positional argument formatting on object-related feed URLs, which will cause sites to fail to build with: "TypeError: not all arguments converted during string formatting". In that case, the theme needs to be updated.

The one that I fixed was `CATEGORY_FEED_ATOM = 'feeds/%s.atom.xml'` to `CATEGORY_FEED_ATOM = 'feeds/{slug}.atom.xml'`

My problem is that I am still encountering this `TypeError` and I'm not sure where. This is my verbose build output:

```
$ pelican content --output output --settings publishconf.py --verbose
-> Writing /home/andy/my_repos/test_pelican/output/feeds/all.atom.xml
-> Writing /home/andy/my_repos/test_pelican/output/feeds/review.atom.xml
-> Writing /home/andy/my_repos/test_pelican/output/hello-world.html
-> Writing /home/andy/my_repos/test_pelican/output/index.html
-> Writing /home/andy/my_repos/test_pelican/output/tags.html
CRITICAL: TypeError: not all arguments converted during string formatting
```

If I run --print-settings, the only setting with a `%` in it is `DEFAULT_DATE_FORMAT` which is unchanged from the default value of `'%a %d %B %Y'` set in Pelican's [settings.py on line 118](https://github.com/getpelican/pelican/blob/71650b408cefc5d29fe81a91984b6ecd02e1d513/pelican/settings.py#L118)

How can I determine where or what is causing the TypeError?

julien-h:
This error occurs even with the builtin theme "simple". I opened an issue about this here: https://github.com/getpelican/pelican/issues/2487

bryanbrattlof:
Hi'ya @julien-h and @AWegnerGitHub

I've been messing around with this for a little bit, but I'm unsure if the error I'm getting is the same as yours.  The `TypeError` I'm getting is related to the `base.html` template using the old format to generate the urls.
```
CATEGORY_FEED_ATOM|format(category.slug)
TAG_FEED_ATOM|format(tag.slug)
```
If that's the same error then we just need to use the keyword argument
```
CATEGORY_FEED_ATOM|format(slug=category.slug)
TAG_FEED_ATOM|format(slug=tag.slug)
```
[...]

PR #2515, opened 2019-01-15
Update 'simple' theme to use new Atom/RSS settings

bryanbrattlof:
While, I'm working on #2513, I thought it might be easier for the maintainers to add this fix and worry about adding features in a later release. It should fix #2506 and fix #2487 

Going off the [release notes](https://blog.getpelican.com/pelican-4.0-released.html) for 4.0, I replaced the ```|format(``` with ```.format(slug=```

PR #2538, opened 2019-03-01
themes: Fix format issue in `simple` template

mrshu:
* The `simple` template that is bundled with Pelican has been shown to
  suffer from the problem described in #2489. This has also been
  reported as a bug in the Debian Bug tracer:
  https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=919636

* This commit fixes the problem by updating the `simple` template to the
  new way of interpolating variables via the Jinja2 `format` helper
  function.

Signed-off-by: mr.Shu <mr@shu.io>


mrshu:
Closing this as a better fix can be found in #2515

PR #2542, opened 2019-03-11
Quote Makefile paths, to handle spaces. Fixes #2066

### LINKED ISSUE #2066: Weird make problem

bberberov:


justinmayer:
@bberberov: Did you actually encounter the problem mentioned in the referenced issue with a Makefile generated by Pelican 4.0.1?

bberberov:
Yes, `pelican --version` is 4.0.1.

I'm not sure what your platform is, but you can reproduce it on Linux:

1. `mkdir "space test"`
2. `cd "space test"`
3. `pelican-quickstart`
4. `make publish`

I believe the problem is caused by `$(CURDIR)` not escaping the necessary characters in the default configuration.

bberberov:
@oulenz Only when you use a non-standard Pelican installation.  For example:

```Makefile
PELICAN?=/usr/opt/My Pelican/pelican
```

`$(PELICAN)` should always be a single token so it is OK to always quote it, IMHO.

___

There might be some issues with the `_upload` targets.  I'm not 100% sure `"$(OUTPUTDIR)"/` and `"$(OUTPUTDIR)"/*` are the is the safest constructs.  I was having some potential issues yesterday.  **I will try to put this back to `WIP` and review.**

I don't have AWS so I'm not sure if you can have have spaces in `S3_BUCKET`.

`CLOUDFILES_USERNAME`, `CLOUDFILES_API_KEY`, `CLOUDFILES_CONTAINER` `GITHUB_PAGES_BRANCH` can probably also be quoted.
___

Is it OK to make `SSH_PORT` optional like `PORT` is made optional earlier in the file?  My config does not require it.

oulenz:
I would have said that if users explicitly set  `PELICAN`, it's up to them to escape any spaces.

In a similar vein, would it perhaps be cleaner to escape `INPUTDIR`, `OUTPUTDIR` etc at the top of the file rather than all the places where they are used?

bberberov:
> I would have said that if users explicitly set `PELICAN`, it's up to them to escape any spaces.

It really doesn't hurt anything, unless you want to pass arguments there.  I don't think that's what the variable is for.

> In a similar vein, would it perhaps be cleaner to escape `INPUTDIR`, `OUTPUTDIR` etc when they are set, at the top of the file, rather than in all the places where they are used?

I thought about that.  Then I looked it up.  `make` doesn't really escape quoted content.  It becomes part of the variable value verbatim.  It didn't make sense to me.  Quoting is for the shell command invocation, IMHO.

bberberov:
I tested the `_upload` targets.  What is in the commit is correct and consistent with the previous `Makefile` commands.  I tried to keep it conservative.  There are some minor improvements I can suggest, but that's maybe a separate pull request, if this one goes through.

oulenz:
The way I understand it, users can do with this Makefile whatever they want. The point of `pelican-quickstart` is to present new users with something that is usable out of the box. You are right that it should still work if `CURDIR` contains spaces. Including quotes when `INPUTDIR` and friends are set would achieve that. Including these quotes when the variables are set means it is less work to remove them if a user who doesn't use spaces wants to clean up their Makefile.

The quickstart will change `PELICAN?={{pelican}}` to `PELICAN?=pelican`, which doesn't need quotes. If a user changes this to `PELICAN?=/usr/opt/My Pelican/pelican` then at that point they should include quotes.

I agree your approach doesn't do any harm, and if others agree then I am fine with it. In the end it's a matter of taste.

justinmayer:
Many thanks to @bberberov  for the contribution and to @oulenz for reviewing.

PR #2548, opened 2019-04-02
Allow empty save_as in PAGINATION_PATTERNS for capping max page

est:
I want to limit the maximum number of older pages a viewer can visit, so here's a quick fix.

Just place settings like these in your `pelicanconf.py`:

    PAGINATION_PATTERNS = (
        (1, '{name}{extension}', '{name}{extension}'),
        (2, '{name}.{number:02d}{extension}', '{name}.{number:02d}{extension}'),
        (10, '#', ''),  # allow only up to 9 pages generated
    )

And it's done.

It's consistent with the rest of `*_SAVE_AS` settings convention.

### DIFF
--- a/pelican/writers.py
+++ b/pelican/writers.py
@@ -262,8 +262,9 @@ class Writer(object):
 
                 localcontext = _get_localcontext(
                     context, page.save_as, paginated_kwargs, relative_urls)
-                _write_file(template, localcontext, self.output_path,
-                            page.save_as, override_output)
+                if page.save_as:
+                    _write_file(template, localcontext, self.output_path,
+                                page.save_as, override_output)
         else:
             # no pagination
             localcontext = _get_localcontext(


avaris:
I am very confused by this. You want to stop writing pages after a point? 

But... `*_next_page` is still there, as well as `page.has_next()` will be `True`. You'll end up with really weird links [here](https://github.com/getpelican/pelican/blob/master/pelican/themes/simple/templates/pagination.html#L7-L9).

est:
@avaris  yes I want to stop writing pages after a point.

> *_next_page is still there, as well as page.has_next() will be True

Let me take a look

avaris:
I'm still puzzled by the *"why"* of all this. Leaving that aside, abusing `PAGINATION_PATTERNS` isn't best for this. The way I'd go about this is probably: adding a new config (`MAX_NUMBER_OF_PAGES = None`) and capping it after [here](https://github.com/getpelican/pelican/blob/master/pelican/paginator.py#L54):
```python
if self.settings['MAX_NUMBER_OF_PAGES'] is not None:
    self._num_pages = min(self._num_pages, self.settings['MAX_NUMBER_OF_PAGES'])
```

est:
> I'm still puzzled by the "why" of all this

Just my personal itch. I'd like to make posts older than certain period of time *not enumerable*, but reachable via links or SERP. Just similar to many people would delete their older twitter posts or reddit posts to limit exposure.

I am also experimenting capping it by time. I'll just close this PR in case this this feature is too confusing.

PR #2580, opened 2019-06-07
Add FUNDING.yml to project

justinmayer:
This adds multiple methods for financially supporting Pelican's development. Primary motivations are to defray expenses for servers and domain name registration, as well to support and encourage Pelican's on-going maintenance and enhancement.


justinmayer:
Yay for what will hopefully be one small step toward sustainable open-source development. 🎉

PR #2526, opened 2019-02-02
Add livereload invoke task. Fixes #1326

johnfraney:
Adds a `livereload` invoke task that reloads the browser window when
content files are updated.

Usage:

```console
$ invoke livereload
[I 190202 16:28:30 server:298] Serving on http://127.0.0.1:5500
[I 190202 16:28:30 handlers:59] Start watching changes
[I 190202 16:28:30 handlers:61] Start detecting changes
[I 190202 16:28:32 handlers:132] Browser Connected: http://127.0.0.1:5500/
```

See: https://livereload.readthedocs.io/en/latest/

Issue #1326, opened 2014-04-23
Live page reloading


kylef:
Jekyll has a "live reload" feature, which extends on from `pelican --reload` and adds HTML and JavaScript magic so that it will automatically reload a HTML page in your browser when you make changes to the source. This is really nice for developing.

I'm have no idea how this would work, weather it would be better suited in a plugin etc. So chucking it up here to discuss.

leotrs:
@justinmayer What's the state of `pelican --autoreload`?

avaris:
I'd say pretty confused, since `--autoreload` doesn't have anything to do with serving pages.

leotrs:
It says 
```
 -r, --autoreload      Relaunch pelican each time a modification occurs on
                               the content files. (default: False)
```

If the point is to get the content to the browser as soon as possible, I thought relaunching pelican would have to be part of that. Am I missing something?

justinmayer:
@leotrs: `pelican -r` already re-generates output when source articles/pages change. As I understand it, @kylef is talking about subsequently triggering a *browser* refresh to see the changed output without having to manually reload the page.

As for the status, to my knowledge a pull request was never submitted, and I'm not sure that anyone has tested the commit that @notetau made in his fork. If someone were to prepare a proper pull request (with docs, tests, etc), it would be a nice feature addition.  (^_^)

leotrs:
I see. Thanks. I agree this would be desirable.

saschwarz:
FYI: this blog post gives a pretty easy way to do this: https://merlijn.vandeen.nl/2015/pelican-livereload.html I just tried it and it seems to work nicely when using port 5500 I didn't need to install any browser plugin(?). It could be a new script that is integrated in the Makefile/fabfile.

dandavison:
I'm also looking forward to having this in pelican. In case it's helpful, the `grip` markdown renderer https://github.com/joeyespo/grip has an implementation that works well.

To add to @saschwarz's comment, another (even easier) workaround for those using chrome on OS X is to use `chrome-cli reload -t <tabid>` https://github.com/prasmussen/chrome-cli. E.g. by configuring your text editor to make the call to `chrome-cli` on file modification, or by having some shell code watching for file modification.

josch:
This feature would go excellent together with the `pelican --listen` option that I implemented in pull request #2176. If anybody would accept my pull request then I volunteer to implement this feature as well.

It probably doesn't hurt to even have it enabled by default without needing an extra switch, right?

[...]

PR #2584, opened 2019-06-20
Enhance Invoke livereload task

justinmayer:
Main changes are:

* Instead of using LiveReload's default port (5500), use the existing `port` value from the task's CONFIG dictionary, which defaults to 8000.
* Add Pelican's settings file to the list of watched files
* Add theme's templates and static assets (CSS + JS) to the list of watched files

Related to #1326 and #2526. cc: @bhrutledge, @johnfraney


johnfraney:
@justinmayer Looks good! These improvements make a lot of sense.

PR #2586, opened 2019-06-28
Add pyproject.toml and poetry.lock to project

### LINKED ISSUE #2450: Define dependencies in pyproject.toml file
Python dependency resolution is vastly improved via [PEP 518 and pyproject.toml](https://www.python.org/dev/peps/pep-0518/) compared to Pip and requirements.txt. Might be nice to add a pyproject.toml file to the project to manage Pelican's dependencies.

justinmayer:
This could facilitate publishing new versions of Pelican as well as make
it easier for Poetry users to install Pelican via: `poetry install`

Closes #2450


justinmayer:
Thanks for reviewing, Kevin! :grin:

PR #2589, opened 2019-07-03
Replace os.mknod with open for macos and windows compatibility

oulenz:
Fix issue raised in https://github.com/getpelican/pelican/pull/2588

As far as I can tell, this is also supposed to work on MacOS and Windows, but I haven't tested it..


justinmayer:
The test in question now passes on macOS. Thank you, @oulenz! 🎉

PR #2590, opened 2019-07-03
Consolidate version strings in pyproject.toml

justinmayer:
Storing the current version in a single place greatly simplifies issuing new package releases. I'm working on some new automation tooling, and having the version string exist solely in `pyproject.toml` will help that endeavor significantly.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,10 +1,28 @@
 #!/usr/bin/env python
+import re
 import sys
 from io import open
 from os import walk
 from os.path import join, relpath
 
 from setuptools import setup
+
+
+def get_version():
+    VERSION_REGEX = re.compile(
+        r"^version\s*=\s*\"(?P<version>.*)\"$"
+    )
+    with open("pyproject.toml") as f:
+        for line in f:
+            match = VERSION_REGEX.match(line)
+
+            if match:
+                return match.group("version")
+
+    return None
+
+
+version = get_version()
 
 requires = ['feedgenerator >= 1.9', 'jinja2 >= 2.7', 'pygments', 'docutils',
             'pytz >= 0a', 'blinker', 'unidecode', 'six >= 1.4',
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -6,6 +6,10 @@
 
 from setuptools import setup
 
+from pelican import get_version
```
avaris: This is an issue. Importing `pelican` will try to import loads of not-yet-installed dependencies.

pip 19+ is fine because it added `pyproject.toml` but earlier versions will fail. Similarly `python setup.py install` will also fail.

justinmayer: I wondered about that. Thanks for pointing that out. At the minor expense of making it less DRY, I re-pushed with a slimmer version of the `get_version()` function added directly to `setup.py`. Do you see any issues with how things look now? Anything you think I should change?

avaris: Is the `get_version` in `__init__.py` necessary? Only way it would run successfully is to import pelican from source folder (not installed) but all the dependencies were installed. Odd condition.

avaris:
Oh, by the way, that regex won't capture stuff like `"4.0.2.dev0"`. Why not gather everything between `""`?

```python
VERSION_REGEX = re.compile(
        r"^version\s*=\s*\"(?P<version>.*)\"$"
    )
```

setup.py
justinmayer: No, I don't think the `get_version()` function is necessary. I removed it in the latest commit and moved its basic logic to the `except` block below. Does that make things better?

justinmayer:
The main rationale for that regex is that I don't intend to use those kinds of non-release versions going forward, but for now I've adopted your proposed regex in case that plan doesn't work out. :wink:

setup.py
justinmayer: I understand better now your point about it being an odd condition that doesn't need to be addressed. I removed that logic from `__init__.py`.

avaris:
:+1:

justinmayer:
Addendum: I reverted the `setup.py` portion of this PR via a52c67d4.

PR #2593, opened 2019-07-09
Use relative paths only in StaticGenerator

sharmaeklavya2:
While trying to write a plugin for post-processing static files, I saw that `context['staticfiles']` had duplicate entries for the same file. I checked `StaticGenerator.generate_context` and I found that we're using absolute paths for linked static files and relative paths for static files from `STATIC_PATHS`.

I fixed it by using relative paths for linked files.

### DIFF
--- a/pelican/generators.py
+++ b/pelican/generators.py
@@ -800,8 +800,7 @@ class StaticGenerator(Generator):
 
     def generate_context(self):
         self.staticfiles = []
-        linked_files = {os.path.join(self.path, path)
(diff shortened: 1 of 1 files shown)


justinmayer:
@oulenz: As someone who has spent considerable time with Pelican's handling of static files, do you have any thoughts regarding this change?

justinmayer:
Many thanks to @sharmaeklavya2 for the fix and to @oulenz for reviewing. 💫

PR #2599, opened 2019-07-30
Add FEED_START_DATE to shortcut initial feed entries

heliocastro:
Usually feed parsers from different systems ignores past entries from
the original migraated blog when catch a new feed,
then simply repost all entries of the newly migrated
blog/site again, making an unwanted side effect.

This option enables to decide when the newly created site should
create the feeds and provide the proper atom/rss outputs

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -931,6 +931,14 @@ the ``TAG_FEED_ATOM`` and ``TAG_FEED_RSS`` settings:
    Maximum number of items allowed in a feed. Feed item quantity is
    unrestricted by default.
 
+.. data:: FEED_START_DATE
+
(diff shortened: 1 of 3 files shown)


justinmayer:
Hi Helio. Thank you for your pull request submission. As @davidag suggested, I think this might be best suited as a Pelican plugin. If you would be interested in reformulating this functionality in plugin form, I would be happy to create a separate repository for it under the new [Pelican Plugins](https://github.com/pelican-plugins) organization. What do you think?

justinmayer:
As mentioned above, we would welcome this functionality as a plugin. If interested, please comment here so I can create a repository at the new org, as noted above.

PR #2600, opened 2019-08-01
Allowing to pass `argv` to pelican `main` entrypoint

Lucas-C:
The benefit of this is to allow to directly call pelican from `invoke` tasks, without needing to pass through system calls with `c.run` :
```python
from invoke import task
from pelican import main as pelican_main
CONFIG = ...
@task
def build(c):
    pelican_main('-s {settings_base} -o {deploy_path}'.format(**CONFIG).split(' '))
```

### DIFF
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -307,7 +307,7 @@ class PrintSettings(argparse.Action):
         parser.exit()
 
 
-def parse_arguments():
+def parse_arguments(argv=None):
     parser = argparse.ArgumentParser(
         description='A tool to generate a static blog, '
                     ' with restructured text input files.',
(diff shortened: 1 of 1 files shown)


Lucas-C:
Usage example in an existing blog:
https://github.com/Lucas-C/ludochaordic/blob/master/tasks.py#L13

justinmayer:
This seems reasonable enough to me. Any comments from other @getpelican/reviewers?

justinmayer:
Many thanks to @Lucas-C for the enhancement and to @avaris for reviewing.

justinmayer:
Thanks again for the fix, @Lucas-C. This change has been automatically published as [Pelican 4.1.1](https://github.com/getpelican/pelican/releases/tag/4.1.1), courtesy of [AutoPub](https://github.com/autopub/autopub). ✨

PR #2602, opened 2019-08-03
[Fix #2592] fix default binding address

### LINKED ISSUE #2592: Pelican binds to 0.0.0.0 instead of 127.0.0.1 when --bind is omitted
### Problem

When starting Pelican like
`$ pelican --listen`

it binds to all network interfaces. However:

```
$ pelican --help
...
  -b BIND, --bind BIND  IP to bind to when serving files via HTTP (default:
                        127.0.0.1) (default: None)
```

When ommiting --bind, it should bind to 127.0.0.1 only.

### Observation

My macOS machine presents a pop-up with the question if I want to allow '_The application "Python.app" to accept incoming network connections?_'. Typically, macOS only asks this question if a service is being bound to a public network interface and not the loop back interface.

After clicking allow, a netstat shows Pelican is bound to all available interfaces (note: one should first visit the site on the IP number of the public network interface, otherwise netstat won't show this):

```
$ netstat -an|grep '*.8000'
tcp4       0      0  *.8000                 *.*                    LISTEN

```

### Cause

When pelican is started with --listen, but --bind is ommited, default settings from pelican/settings.py are used. In the DEFAULT_CONFIG dictionary, we find: 'BIND': '',

BIND is then passed to listen(), RootedHTTPServer(), BaseHTTPServer() and finally to socket()

The socket documentation states:

> For IPv4 addresses, two special forms are accepted instead of a host address: '' represents INADDR_ANY, which is used to bind to all interfaces, and the string '<broadcast>' represents INADDR_BROADCAST.

Thus, because the default setting of BIND is '', Pelican is bound to all interfaces and not, as documented and promised, 127.0.0.1 only.

### Solution

The solution is to set BIND to '127.0.0.1' in pelican/settings.py. As a workaround, it is also possible to set BIND = '127.0.0.1' in pelicanconf.py

apoos-maximus:
Signed-off-by: apoos-maximus <apoorv.99.sachan@gmail.com>

### DIFF
--- a/pelican/settings.py
+++ b/pelican/settings.py
@@ -166,7 +166,7 @@ DEFAULT_CONFIG = {
     'WRITE_SELECTED': [],
     'FORMATTED_FIELDS': ['summary'],
     'PORT': 8000,
-    'BIND': '',
+    'BIND': '127.0.0.1',
 }
 
 PYGMENTS_RST_OPTIONS = None


justinmayer:
Sincere apologies for the delay in reviewing, @apoos-maximus. In the interim, I realized I merged  PR #2630 that already implements this. I'm very sorry for the oversight on my part! 😓

PR #2607, opened 2019-08-16
Fix PAGINATION_PATTERNS documentation example

jrarseneau:
The example for PAGINATION_PATTERNS is invalid and fails if someone would  just copy and paste into their own `pelicanconf.py`.

This should fix it.


justinmayer:
Thanks for the fix, @jrarseneau. Without realizing your PR was in the queue, I already fixed this via 7f4e614b.

Issue #2419, opened 2018-10-18
Markdown in title


fancsali:
At first sight it does not make sense, but that would allow us to use the same extensions (e.g. Smarty) in a consistent way; so one would be able to type ```---``` to have ```@emdash;``` in the final output.

justinmayer:
Much easier to just type the `—` m-dash directly in the title. Any other use cases that would warrant supporting Markdown in the title?

avaris:
`TYPOGRIFY = True`
(see [typogrify](https://github.com/mintchaos/typogrify))

justinmayer:
Ah, that's right. As @avaris mentioned, Typogrify should address this.

fancsali:
All fair points, however:

- Typing an em-dash, is not that easy after all. Also, in theory we could type everything, since we're using UTF-8, but for practicality reasons we chose Markdown. But to be honest, the inconsistency was the reason, I felt this might be a reasonable suggestion.
- Trust me, I did try enabling Typogrify, but unfortunately that is a rather old version, and I couldn't find any way to pass arguments/config settings to that. And thus, I was unable to type an en-dash for example. I felt, the 'Smarty' filter that comes with Markdown is much more flexible, and up-to-date.

I do accept, there are other ways, and from a non-Markdown-only perspective it does make sense to structure the system as it is at the moment. I just felt, it might be worthwhile thinking about this aspect as well.

justinmayer:
I agree that it's a reasonable enough suggestion to put forth. Like most software development, it's a question of costs and benefits. On macOS, for example, shift-option-hyphen produces an em-dash, so it's a quick way of addressing that need (on that platform, anyway). Absent other use cases, it's hard to conjure up the motivation to prioritize adding Markdown processing to titles.

I'm also one of the maintainers of Typogrify, so if you'd like to help improve it and make it do what you want it to do, I'd happy to try my best to review any contributions and help get them merged.

avaris:
@fancsali 
[`default`](https://pythonhosted.org/smartypants/reference.html#smartypants._Attr.default) for `smartypants` is to only convert `--` into `em-dash`. Add this to your settings file:
```python
import smartypants
smartypants.Attr.default = smartypants.Attr.set2
```
Then `--` should be `en-dash` and `---` should be `em-dash`. With that, following markdown content:
```markdown
Title: A---markdown--powered article
Date: 2018-01-01

Content
```
generates
```
A—markdown–powered article
```
with source
```html
[...]

PR #2615, opened 2019-09-04
Add Typogrify/SmartyPants dashes config parameter

fancsali:
As a result of  the discussion on #2419, I added a few lines of code, to make the behaviour of Typogrify configurable.

Naming of the different "modes" come from the source code of `smartypants.py`: https://github.com/leohemsted/smartypants.py/blob/master/smartypants.py


oulenz:
Hm, I'd expect the `oldschool` behaviour to be the default (coming from latex).

fancsali:
> Hm, I'd expect the `oldschool` behaviour to be the default (coming from latex).

Well I myself tend to agree; that's essentially the reason I created this wee change.

However, I felt we shall match the hard-coded default of SmartyPants/Typogrify, so to be backwards compatible: If someone upgrades, the behaviour should not change. Also, I felt, specifying `default` or omitting `TYPOGRIFY_DASHES` entirely should result in the same behaviour.

oulenz:
You're right, and this is a useful tweak!

I was wondering whether we should make this more general, and allow users to pass any smartypants options, but I don't really see anything else anyone might want to change, and that would only make the setting less intuitive.

In fact, could we simply turn this setting into a boolean? (I don't really see why anyone would want to use the inverted setting: -- for em-dash and --- for en-dash.)

Plus, could you test this as part of one of the existing reader tests?

justinmayer:
The `oldschool` setting is a bit of a misnomer. In fact, I think the existing default behavior is for even-more-old-school folks like me who remember using double-hyphens as replacements for em-dashes on typewriters, which had no built-in way to type em-dashes. So the default behavior is very familiar and natural for me. I rarely use en-dashes, and when I do, I type them directly (_option-hyphen_ on macOS).

While I don't understand why folks would want the inverted setting, I'm not sure that's a compelling argument to remove it as an option. Plus, perhaps it makes the most sense to mirror the underlying modes in `smartypants.py` as @fancsali has done here.

In the end, this is going to be a set-it-and-forget-it option for folks, so I'm not overly concerned about it being intuitive. Even a Boolean setting would require explanation as to what it actually does.

fancsali:
I've been contemplating the generic version, but then I too concluded, that it's less intuitive and user friendly to have to muck about with values like `set1`.

Regarding the boolean version, I did think about that too, however, as @justinmayer said, I guess, it's the smart thing to mirror the behaviour of the underlying code. Also, a very good point regarding the `oldschool` method: I still tend to find it intuitive to use double dashes as em-dashes.

All in all, I felt, both the `oldschool` and `oldschool_inverted` have their own legitimate arguments for: If you want to go with the logic "the longer the dash the more hyphens I type", you use one; but it is a complete valid thinking if you'd like to use double-hyphens for em-dashes, and whenever, once in a blue moon, you need en-dashes for things like number ranges, you just go with the triple solution.

@oulenz: Regarding the tests, which readers do you mean: the config-reader, or the ones that read the actual `.rst` and `.md` content?

avaris:
About tests, a single content file generated with different dash options and comparing the expected output should suffice (similar to [this](https://github.com/getpelican/pelican/blob/master/pelican/tests/test_readers.py#L356)).

justinmayer:
@fancsali: Many thanks for resuming work on this in order to add tests. It looks like the `test_article_order_by` and `test_generate_context` tests are currently failing, as well as one or more `flake8` linting checks.

fancsali:
@justinmayer: Cheers for pointing that out. I was finalising that stuff on-the-go, on my Chromebook, and that environment is rather limited, so some tests are expected to fail. Must have missed those in the noise.

justinmayer:
Kudos to @fancsali for this useful addition to the project, and thanks to @oulenz and @avaris for reviewing. It's a pleasure to merge contributions like this. 💫

fancsali:
> Kudos to @fancsali for this useful addition to the project, and thanks to @oulenz and @avaris for reviewing. It's a pleasure to merge contributions like this. 💫

The pleasure is all mine. 😊

PR #2617, opened 2019-09-10
Plugins in PLUGIN_PATHS should be searched after the ones installed from Pypi

Lucas-C:
_cf._ https://github.com/getpelican/pelican-plugins/issues/39 https://github.com/getpelican/pelican-plugins/issues/425 and https://github.com/getpelican/pelican-plugins/issues/1189 for more context.

As we are migrating some plugins out of `pelican-plugins`, this would make it easier for our users to switch to their counterpart packages on Pypi.

### DIFF
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -75,7 +75,7 @@ class Pelican(object):
         logger.debug('Temporarily adding PLUGIN_PATHS to system path')
         _sys_path = sys.path[:]
         for pluginpath in self.settings['PLUGIN_PATHS']:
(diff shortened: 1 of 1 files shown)


avaris:
Unfortunately, this won't be a good idea. Plugins have been quite liberal with their names and it's quite likely that a number of them would clash with other installable python packages. In those situations, it is broken with the only possible solution of uninstalling clashing package.

Lucas-C:
Ok, I understand that the risks are too high. Closing this
Thank you @avaris

PR #2621, opened 2019-09-23
Fix pelican.settings.load_source to avoid caching issues

Lucas-C:
I met this caching issue while working on `pelican-plugins` CI : https://github.com/getpelican/pelican-plugins/issues/1170

In this context, `pelican.settings.load_source` is repeatedly called in the same `nostests` process, and when `name='pelicanconf'` the same settings are returned on subsequent calls to `read_settings`.

Here is a minimal code example reproducing the issue : https://chezsoi.org/shaarli/?hBpGew

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+Release type: patch
+
+Fix pelican.settings.load_source to avoid caching issues - PR #2621
--- a/pelican/settings.py
+++ b/pelican/settings.py
@@ -14,12 +14,16 @@ import six
 
 from pelican.log import LimitFilter
 
+
 try:
-    # SourceFileLoader is the recommended way in Python 3.3+
-    from importlib.machinery import SourceFileLoader
+    # spec_from_file_location is the recommended way in Python 3.5+
+    import importlib.util
 
     def load_source(name, path):
-        return SourceFileLoader(name, path).load_module()
+        spec = importlib.util.spec_from_file_location(name, path)
+        mod = importlib.util.module_from_spec(spec)
+        spec.loader.exec_module(mod)
+        return mod
 except ImportError:
     # but it does not exist in Python 2.7, so fall back to imp
     import imp


avaris:
Code is good, but small nitpick on the comment: Python docs mention that this is 3.5+. And that is fine since `pelican` is also 3.5+.

About the issue, I'm not on my PC right now to confirm it, but I'm guessing `SourceFileLoader` uses `sys.modules` to store loaded modules with name and later retrieve loaded module if name was already in `sys.modules` (like normal `import` would do). `del sys.modules[name]` afterwards would probably also do the trick here, but this is better :).


PS: We need to update [Quickstart](https://github.com/getpelican/pelican/blob/master/docs/quickstart.rst). It still says 3.3.

Lucas-C:
Thanks for the review !
I updated the PR to mention Python3.5+

Lucas-C:
For the Quickstart doc update: https://github.com/getpelican/pelican/pull/2622

justinmayer:
Speaking of nitpicks, here is a _real_ nitpick… From the [Contributing docs](http://docs.getpelican.com/en/stable/contribute.html#using-git-and-github):
> First line of your commit message should start with **present-tense** verb […]

So in this case, instead of "Fixing pelican.settings.load_source to avoid caching issues", the canonical verb conjugation for this project would start with "Fix" instead.

Just a note for future reference. 😊

justinmayer:
@Lucas-C: Since this change includes something more significant than a typo/doc fix, would you please also add a [RELEASE](http://docs.getpelican.com/en/stable/contribute.html#using-git-and-github) file (see third bullet point) with `Release type: patch` and amend this commit to include it?

Lucas-C:
Ok for the note on the commit message convention, I fixed it :)

I also added a `RELEASE.md` file. I was curious to realize non existed previously.
Tell me if I created it as you wished.

justinmayer:
Looks good. Many thanks for the fix.

The `RELEASE.md` file is automatically deleted via [AutoPub](https://github.com/autopub/autopub) and the CI system once the information within the release file has been extracted, which is why it does not persist in the repository. Here's a [previous commit](https://github.com/getpelican/pelican/commit/b2da535fecafa7a81feee2dd5b19eb1fffcc9a94), after which AutoPub subsequently and dutifully [deleted the RELEASE.md file](https://github.com/getpelican/pelican/commit/e44aa294b0641655165fefdcbce72484e3ce7c57).

justinmayer:
Congrats! This has been released as Pelican 4.1.2. ✨

Lucas-C:
Thanks for the explanation :)

PR #2628, opened 2019-10-02
 New feature: support for {include} syntax. Fixes #1902.

### LINKED ISSUE #1902: RFE: Make it possible to include content files from other content files

Lucas-C:
I took over @atodorov work started in https://github.com/getpelican/pelican/pull/1909 in order to complete this feature. I changed the code accordingly to the recommandations in the original PR.

The new {include} syntax makes it possible to include
frequently used text snippets into your content.

### DIFF
--- a/pelican/contents.py
+++ b/pelican/contents.py
@@ -26,6 +26,9 @@ from pelican.urlwrappers import (Author, Category, Tag, URLWrapper)  # NOQA
 
 logger = logging.getLogger(__name__)
 
+INCLUDE_RE = re.compile(r'(?P<indent>\n[ \t]+)?'
+                        r'[{|]include[|}](?P<path>[\w./]+)')
+
 
 @python_2_unicode_compatible
 class Content(object):
@@ -359,14 +362,8 @@ class Content(object):
             path = value.path
             if what not in {'static', 'attach'}:
                 continue
-            if path.startswith('/'):
-                path = path[1:]
-            else:
-                # relative to the source path of this content
-                path = self.get_relative_source_path(
-                    os.path.join(self.relative_dir, path)
-                )
-            path = path.replace('%20', ' ')
+            path = relativize_path(self.settings['PATH'],
+                                   self.relative_dir, path)
             static_links.add(path)
         return static_links
 
@@ -449,24 +446,11 @@ class Content(object):
         """
         if not source_path:
             source_path = self.source_path
-        if source_path is None:
-            return None
-
-        return posixize_path(
-            os.path.relpath(
-                os.path.abspath(os.path.join(
-                    self.settings['PATH'],
-                    source_path)),
-                os.path.abspath(self.settings['PATH'])
-            ))
+        return get_relative_source_path(self.settings['PATH'], source_path)
 
     @property
     def relative_dir(self):
-        return posixize_path(
-            os.path.dirname(
-                os.path.relpath(
-                    os.path.abspath(self.source_path),
-                    os.path.abspath(self.settings['PATH']))))
+        return relative_dir(self.settings['PATH'], self.source_path)
 
     def refresh_metadata_intersite_links(self):
         for key in self.settings['FORMATTED_FIELDS']:
@@ -613,3 +597,111 @@ class Static(Content):
 
         self.override_save_as = new_save_as
         self.override_url = new_url
+
+
+def get_relative_source_path(content_path, source_path):
+    if source_path is None:
+        return None
+
+    return posixize_path(
+        os.path.relpath(
+            os.path.abspath(os.path.join(
+                content_path,
+                source_path)),
+            os.path.abspath(content_path)
+        ))
+
+
+def relativize_path(content_path, relative_dir, path):
+    """
+    Update path depending on whether this is an absolute
+    or relative value.
+    """
+    if path.startswith('/'):
+        path = path[1:]
+    else:
+        path = get_relative_source_path(content_path,
+                                        os.path.join(relative_dir, path))
+
+    path = path.replace('%20', ' ')
+
+    return path
+
+
+def relative_dir(content_path, path):
+    return posixize_path(
+        os.path.dirname(
+            os.path.relpath(
+                os.path.abspath(path),
+                os.path.abspath(content_path))))
+
+
+def insert_included_content(content,
+                            source_path,
+                            content_path,
+                            exclude_exts=()):
+    '''
+    Replace {include}some.file with the
+    contents of this file.
+    '''
+    processed_paths = set()
+    # In Python 3.x we can use the `nonlocal` declaration, in `replacer()`,
+    # to tell Python we mean to assign to the `source_path` variable from
+    # `insert_included_content()`.
+    # In Python 2.x we simply can't assign to `source_path` in `replacer()`.
+    # However, we work around this by not assigning to the variable itself,
+    # but using a mutable container to keep track about the current working
+    # directory while doing the recursion.
+    source_dir = [relative_dir(content_path, source_path)]
+
+    def replacer(m):
+        path, indent = m.group('path'), m.group('indent')
+        path = relativize_path(content_path, source_dir[0], path)
+        path = posixize_path(
+                os.path.abspath(
+                    os.path.join(content_path, path)
+                )
+            )
+
+        if not os.path.isfile(path):
+            logger.warning("Unable to find `%s`, skipping include.", path)
+            return m.group()
+
+        _, ext = os.path.splitext(path)
+        # remove leading dot
+        ext = ext[1:]
+
+        if ext in exclude_exts:
+            return m.group()
+
+        with open(path) as content_file:
+            text = content_file.read()
+
+        if indent:
+            prefix = ''
+            if indent[0] == '\n':
+                prefix = '\n'
+                indent = indent[1:]
+            text = prefix + '\n'.join(indent + line
+                                      for line in text.split('\n'))
+
+        # recursion stop
+        if path in processed_paths:
+            logger.warning("Circular inclusion detected for '%s'" % path)
+            return text
+        processed_paths.add(path)
+
+        # if we recurse into another file to perform more includes
+        # _path_replacer needs to know in which directory
+        # it operates otherwise it produces wrong paths
+        source_dir[0] = posixize_path(os.path.dirname(path))
+        current_source_dir = source_dir[0]
+
+        # recursively replace other includes
+        text = INCLUDE_RE.sub(replacer, text)
+
+        # restore source dir
+        source_dir[0] = current_source_dir
+        return text
+
+    return INCLUDE_RE.sub(replacer, content)
--- a/pelican/generators.py
+++ b/pelican/generators.py
@@ -155,16 +155,15 @@ class Generator(object):
 
             if os.path.isdir(root):
                 for dirpath, dirs, temp_files in os.walk(
-                        root, followlinks=True):
-                    drop = []
+                        root, topdown=True, followlinks=True):
                     excl = exclusions_by_dirpath.get(dirpath, ())
-                    for d in dirs:
+                    # We copy the `dirs` list as we will modify it in the loop:
+                    for d in list(dirs):
                         if (d in excl or
                             any(fnmatch.fnmatch(d, ignore)
                                 for ignore in ignores)):
-                            drop.append(d)
-                    for d in drop:
-                        dirs.remove(d)
+                            if d in dirs:
+                                dirs.remove(d)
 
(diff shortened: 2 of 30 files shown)


Lucas-C:
Given Travis results, I need to fix `pyflakes` warnings,
but else this code is ready for review.

Lucas-C:
There are 4 remaining failing tests, that only fail in Travis CI but not on my computer.

I'll try to fix them soon, but I'd be happy to receive any feedback on the code in the meantime :)

Lucas-C:
I finished working on this PR and the Travis CI pipeline is passing !

justinmayer:
Any feedback from @getpelican/reviewers on this new `{include}` feature?

pelican/contents.py
```diff
@@ -613,3 +594,114 @@ def _log_reason(reason):
...
+        path = relativize_path(content_path, source_dir[0], path)
+        path = posixize_path(
+                os.path.abspath(
+                    os.path.join(content_path, path)
+                )
+            )
+
+        if not os.path.isfile(path):
+            logger.warning("Unable to find `%s`, skipping include.", path)
+            return ''.join(('{include}', m.group('path')))
```
avaris: This strips the `indent` part. `m.group()` instead to put the full string back. Similarly down.

pelican/contents.py
```diff
@@ -613,3 +594,114 @@ def _log_reason(reason):
...
+                            content_path,
+                            exclude_exts=()):
+    """
+        Replace {include}some.file with the
+        contents of this file.
+
+        Perform conversion to HTML
+    """
+    regex = r"""(?P<indent>\n[ \t]+)?[{|]include[|}](?P<path>[\w./]+)"""
+    hrefs = re.compile(regex, re.X)
```
avaris: Couple issues:
* Triple quote (`"""`) unnecessary
* You won't match newline (`\n`) unless you do multiline regex. 
* And not much point passing `re.X` (verbose) flag, if you don't intend to write your regex that way.

pelican/contents.py
```diff
@@ -613,3 +594,114 @@ def _log_reason(reason):
...
+
+def insert_included_content(content,
+                            source_path,
+                            content_path,
+                            exclude_exts=()):
+    """
+        Replace {include}some.file with the
+        contents of this file.
+
+        Perform conversion to HTML
```
avaris: Where?

pelican/contents.py
```diff
@@ -613,3 +594,114 @@ def _log_reason(reason):
...
+            prefix = ''
+            if indent[0] == '\n':
+                prefix = '\n'
+                indent = indent[1:]
+            text = prefix + '\n'.join(indent + line
+                                      for line in text.split('\n'))
+
+        # recursion stop
+        if path in processed_paths:
+            raise RuntimeError("Circular inclusion detected for '%s'" % path)
```
avaris: `RuntimeError` is a bit much, no? Warning and skipping would be fine.

pelican/contents.py
```diff
@@ -613,3 +594,114 @@ def _log_reason(reason):
...
+                            exclude_exts=()):
+    """
+        Replace {include}some.file with the
+        contents of this file.
+
+        Perform conversion to HTML
+    """
+    regex = r"""(?P<indent>\n[ \t]+)?[{|]include[|}](?P<path>[\w./]+)"""
+    hrefs = re.compile(regex, re.X)
+    processed_paths = []
```
avaris: Could be `set()` for a slight performance boost.

pelican/generators.py
```diff
@@ -155,16 +155,21 @@ def get_files(self, paths, exclude=[], extensions=None):
...
-                    for d in drop:
-                        dirs.remove(d)
+                            if d in dirs:
+                                dirs.remove(d)
+
+                    d = os.path.basename(dirpath)
+                    if (d in excl or
+                            any(fnmatch.fnmatch(d, ignore)
+                                for ignore in ignores)):
+                        continue
```
avaris: Is this necessary? You won't get to an excluded folder because `dirs` above will filter any subfolders. And this will be in effect only if `root` of `os.walk` matches, which is odd...

avaris:
Also, about the implementation, I'd go a bit different than this. What I had in mind was, essentially in `Readers.read_file`:

* Parse and HTML-ize content
* Check content for `{include}`
  * if found parse and HTML-ize that file through `Readers.read_file`
  * Replace `{include}` bit with returned content

You won't need to much with individual readers, which is problematic, because any additional readers from plugins would need to also implement this.

Bonus: 
* share `context` with `Readers`
* add any included file to `context['included_files']`
* on the generator side, delay content checking (proper metadata etc) until all is processed and warn for only not `{include}`d invalid content.

Lucas-C:
@avaris : I addressed all your code-related comments. I agreed with all of them and made the changes you suggested.

Anwering your first global comment : as the `test_include_markdown_from_html` test shows, no, mixed types are not currently supported.

Lucas-C:
Now answering your 2nd comment:

> Check content for {include}

When would you perform this ? Before or after markup conversion to HTML ?
From my tests, the issue with that approach is that there no good unique answer.
For Markdown or RST, you want to process `{include}` blocks beforehand in order to make a single conversion pass and to avoid `<p>` being inserted all over.

I understand your concern about "mucking with individual readers" though. I guess `{inlude}`-parsing should be done in `read_file`, before the the call to `reader.read(path)`.

I also didn't quite get your "Bonus" suggestions. Could you please elaborate a bit ?

Just for information, I'm not sure I'll have the time to rework this soon.

avaris:
> When would you perform this ? Before or after markup conversion to HTML ?
From my tests, the issue with that approach is that there is no good unique answer.
For Markdown or RST, you want to process {include} blocks beforehand in order to make a single conversion pass and to avoid `<p>` being inserted all over.

After `HTML`, in order to make it reader agnostic so that you can include anything from anything. True, this might result in not-so-ideal `HTML` code, but I consider that as a necessary evil :).

I don't expect including `.md` in `.rst` (or vice versa) to be a common thing, but I could imagine `.html` included in either can be a useful. Granted, that would already work for `.md` with this because it allows raw `HTML` but not for `.rst`.

> I understand your concern about "mucking with individual readers" though. I guess `{inlude}`-parsing should be done in `read_file`, before the the call to `reader.read(path)`.

That would be a better implementation of this one. In addition to not changing individual readers, single point of impact is easier to maintain and debug.

> I also didn't quite get your "Bonus" suggestions. Could you please elaborate a bit ?

Yeah, I noticed it was a bit cryptic. That was a suggestion for auto-excluding `{include}`d files from being considered as regular content. IOW, you keep track of what's `{included}` and remove them from content file in `Generator` classes before doing `is_valid` checks. There are some edge cases for that to fail [*], but it would work most of the time and exceptions can be handled by the user.

[*] This requires checks to be made *after* everything is read, which is currently not the way things are done. Generators don't wait others for this.

As I understand it, currently you have to explicitly exclude them with `IGNORE_FILES` otherwise `pelican` will likely yell at you for missing metadata or such. 

> Just for information, I'm not sure I'll have the time to rework this soon.

I might be able to lend a hand, next weekend-ish :).

Lucas-C:
> True, this might result in not-so-ideal HTML code, but I consider that as a necessary evil :).

I fear that the aproach you suggest `test_include_code_from_markdown` will not work, and I think that's really an important use-case to have for the `{include}` syntax.
Actually, any implementation based on the "convert everything to HTML first" will fail to render this Markdown snippet properly I think:

    ```
    {include}/source.py
    ```

Also, thanks for the clarifications !

Feel free to contribute to this PR if you can, I'd be happy to see your code solution to implement this.

avaris:
Yes, what I was describing won't let you insert code snippets, at least not properly. It can fall back to dumping file content as plain text if there is no reader available for the file type, but then proper formatting, syntax highlighting etc. will be lost.

Given that there are valid use cases for both approaches, how about we extend the scope and add *two* commands: `{include}` and `{raw_include}`

* `{raw_include}` would be more or less this. Replaces with included content as plain text and *then* proceeds with parsing and conversion to HTML. Similar limitations apply: i.e. after placing the included content, final result should be in appropriate syntax for the format.
* `{include}` converts everything to HTML and *then* proceeds with replacement. You can combine `.md` file with `.rst` but anything included has to be parse-able by readers. Or it can be slightly relaxed and could fall back to plain text insertion when no valid readers are found.

Lucas-C:
Sounds like a convenient solution to me ! Great idea.

And I'd vote for the text insertion when no valid reader is found.

oulenz:
I think this is a great idea. Just FYI, I find `{raw_include}` and `{include}` confusing, because I interpreted them the other way around. `{raw_include}` to me suggests the text will be included in the final page/article as is, without being processed, whereas I'd expect `{include}` to work like in e.g. latex (include, then process).

Lucas-C:
Actually me too ! I now realize I understood it the other way round.

avaris:
Doesn't latex process `include`s separately and then combine? It looks similar to `{include}` in that sense. `{raw_include}` would correspond to latex `input`, I believe. i.e. pretty much a copy&paste on the source (my latex is a bit rusty).

But sure, we can decide on different names. These were basically off the top of my head. As long as they are properly documented I don't think it matters much.

Lucas-C:
I'd vote for `raw_include  -  parsed_include` to avoid any mistaken "naive" interpretation of `include`.
As long as each one behaviour is then documented, our users will be able to refer to the doc.

oulenz:
> Doesn't latex process `include`s separately and then combine? It looks similar to `{include}` in that sense. `{raw_include}` would correspond to latex `input`, I believe. i.e. pretty much a copy&paste on the source (my latex is a bit rusty).

You're right that `input` is 'rawer' and therefore closer to this proposal, but as far as I know `include` also uses all the declarations from the main file. (I don't use them regularly myself.)

I'm a bit in two minds over whether we really need two tags. As far as I can tell, there are three use cases:

1. Include md/rst content in an md/rst file: `raw_include` (Is there a use case where a user would want to preparse these?)
2. Include html content in a md/rst file: `parsed_include` (Which is a bit of a misnomer here, because the content won't be parsed.)
3. Include md/rst content in a rst/md file: `parsed_include` (No choice here.)

Doesn't that mean that we can decide which include type we need based on the filetype?

Lucas-C:
@oulenz There's one other use case: include raw content in a md/rst file, _cf_. https://github.com/getpelican/pelican/pull/2628#issuecomment-538877610

oulenz:
You're right, it took me an awful long time to get that.

I guess it would still be possible to decide this by saying: do a `raw_include` unless the included file is in `{html, md, rst}` and of a different type than the main document.

The only problem with that would be including html, md and rst code snippets. (For which an ugly hack would be to change their extension to match the main filetype.)

mosra:
Apologies if someone brought this up already -- the reST parser [already supports such syntax](http://docutils.sourceforge.net/docs/ref/rst/directives.html#including-an-external-document-fragment) (and there it's very flexible), so for `rst` content I'd say it's best to just keep it as-is (and for the rest you could maybe ~~steal~~ reuse what they have invented?). Some examples -- including a full file, with reST syntax parsed (and also all roles, directives and substitution definitions from there brought into current scope):

```rst
.. include:: file.rst
```

Include a fragment of a file delimited by the two identifiers, highlighted as C++ (slicing based on line numbers is also possible):
   
```rst
.. include:: main.cpp
    :code: c++
    :start-after: // begin
    :end-before: // end
```

Include a raw HTML file (or an inline SVG or whatnot) and put it directly into the output without any processing:

```rst
.. raw:: html
    :file: table.html
```

I'm using this feature quite heavily for content on https://blog.magnum.graphics, it's been very useful.

avaris:
Well, markdown also has an [extension for including files](https://github.com/neurobin/mdx_include). So...

Lucas-C:
I tried `mdx_include` and that totally fitted my need : https://github.com/Lucas-C/ludochaordic/commit/4d6b4a0
Thanks for pointing out extensions @mosra !

As far as I'm concerned, they totally answer my need. I'm a bit sad that I may have worked on this PR in vain, but I'm happy to close it knowing this feature isa available as a markup processor extension.

mosra:
> I'm happy to close it knowing this feature is available as a markup processor extension.

:+1: to that, in my opinion no code is the best code, so if this can be handled directly by the parsers Pelican uses, let's just keep the reponsibility there? :)

Lucas-C:
Maybe we should simply include this information in Pelican doc ?

Lucas-C:
Docs improvment suggested in https://github.com/getpelican/pelican/pull/2638

cup:
https://github.com/getpelican/pelican/issues/2669

PR #2632, opened 2019-10-09
Add extra/optional Markdown dependency to setup.py

### LINKED ISSUE #2585: Add Markdown as an (optional) dependency
Since its inception, this project has taken the collective position that since not everyone uses Markdown, the `markdown` package should not be a dependency of the project and should instead be manually installed by users who want to use Markdown.

On the other hand, the `docutils` package that parses reStructuredText markup has always been a dependency. Particularly given Markdown's widespread popularity, I cannot come up with a compelling argument to support the continuation of this discrepancy.

This project is littered with issues caused by environments with missing `markdown` packages. Many folks have encountered frustration when trying to build their Markdown-content sites, only to be met with an (often inscrutable) error message. And that's just the folks who have taken the time to file issues! Imagine how many threw up their hands and just moved on to another solution…

Adding Markdown as a dependency would have significant benefits, including simplifying installation instructions, reducing frustration, and minimizing support request load. So to take another approach at this topic… What are the reasons for *not* adding Markdown as a dependency, and do you think they are sufficiently important to out-weigh the concerns raised above?

davidag:
## PR contents
- Fixes #2585
- Adds `extras_require` for Markdown optional feature.
- Modifies documentation to use the extra dependency in order to install the recommended markdown package version.
- There is no `RELEASE.md` because I wasn't sure it was appropriate. Just let me know and I'll add it.

## Testing
1. Create new Python 3 venv and activate it.
2. Install wheel: `$ pip install wheel`.
3. Setup Pelican for development: `$ python setup.py develop`.
4. Install Pelican with optional markdown: `$ pip install pelican[markdown]`.
5. Verify that the `markdown` package is installed.

Tested again without optional markdown and verify that it's not installed.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -9,6 +9,10 @@ You can install Pelican via several different methods. The simplest is via
 
     pip install pelican
 
+Or, if you plan on using Markdown::
+
+    pip install pelican[Markdown]
+
 (Keep in mind that operating systems will often require you to prefix the above
 command with ``sudo`` in order to install Pelican system-wide.)
 
@@ -40,7 +44,11 @@ Optional packages
 -----------------
 
 If you plan on using `Markdown <http://pypi.python.org/pypi/Markdown>`_ as a
-markup format, you'll need to install the Markdown library::
+markup format, you can install Pelican with Markdown support::
+
+   pip install pelican[Markdown]
+
+Or you might need to install it a posteriori::
 
     pip install Markdown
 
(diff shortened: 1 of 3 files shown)


avaris:
[Installing Pelican](https://github.com/getpelican/pelican/blob/master/docs/install.rst) doc might need an update as well.

justinmayer:
@davidag: No worries about the `RELEASE.md` file for this particular PR, since I have a few recent commits that need to be rolled up and included as well. Many thanks for the enhancement! (And thanks, @avaris, for reviewing!)

Issue #2561, opened 2019-04-24
Inline SVG with a TITLE element causes corrupted HTML


bberberov:
**Steps to reproduce:**

1. Create an HTML content file
2. Add an inline SVG with a `title` element
3. Generate output

**Results:**

It appears the entire content of the body is injected into the `title` within the `head`, up to the closing `title` tag within the inline SVG.

avaris:
Confirmed and briefly checked the code. I *think* [this line](https://github.com/getpelican/pelican/blob/master/pelican/readers.py#L410) should be:
```python
elif tag == 'title' and self._in_head:
```

stuaxo:
Looks good to me, minor nit, but I'd reverse the order for this to make it consistent with the check for body tag, then content in the body, below.

It looks like this would need some new test data, e.g.   `article_with_line_svg.html` and a new test in `HTMLReaderTest`.

stuaxo:
Cheers - I was at work so I couldn't make a PR.

I've started the PR above, with a test for the inline SVG, need to work out how to make the remaining tests pass next.

avaris:
@stuaxo thanks for the PR. Reason for failing tests is that your new test article isn't included in [this](https://github.com/getpelican/pelican/blob/master/pelican/tests/test_generators.py#L240-L283) and [this](https://github.com/getpelican/pelican/blob/master/pelican/tests/test_generators.py#L240-L283) list.

stuaxo:
Cheers, updated.

justinmayer:
This was addressed in the just-released [Pelican 4.2](https://github.com/getpelican/pelican/releases/tag/4.2.0). ✨

PR #2634, opened 2019-10-10
Support inline SVG images

stuaxo:
PR for https://github.com/getpelican/pelican/issues/2561


justinmayer:
Many thanks for the contribution, Stuart!

@getpelican/reviewers / @bberberov: Any feedback on this implementation?

stuaxo:
I realised I haven't tried the HTML file in a browser separately yet, it's worth verifying that browsers are happy with it.

(I remember trying inline SVG years ago and having trouble if I included the namespace).

stuaxo:
OK, opens in browsers for me, so all good here.

justinmayer:
@stuaxo: Fantastic. Would you be so kind as to add a `RELEASE.md` file as noted in the [Contributing](http://docs.getpelican.com/en/stable/contribute.html#using-git-and-github) section of the docs, and then squash any commits that you think don't need to stand on their own? (if any)

justinmayer:
Many thanks to @stuaxo for the enhancement and to @avaris and @iKevinY for reviewing.

This improvement is included in the just-released [Pelican 4.2](https://github.com/getpelican/pelican/releases/tag/4.2.0). ✨

PR #2636, opened 2019-10-13
Include the tests package in the source distribution.

### LINKED ISSUE #2635: tests package is not included in the source tarball on PyPi
The setup.py file has:

```
packages=['pelican', 'pelican.tools'],
```

thus the `tests` package is not available if you install from anything but a github clone. It would be helpful if the tests package was included, as is common with many other python distributions. Maybe use `find_packages()` on this line instead.

moorepants:
Fixes #2635

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,6 @@ script:
   # testing installing with pip
   - pip install -e .
   - python -c "import pelican; import pelican.tools; import pelican.tests"
-  - pip uninstall pelican
 before_deploy:
   - 'if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then travis_terminate 0; fi'
   - pip install githubrelease
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,12 @@ before_install:
   - sudo locale-gen fr_FR.UTF-8 tr_TR.UTF-8
 install:
   - pip install tox==2.5.0
-script: tox -e $TOX_ENV
+script:
+  - tox -e $TOX_ENV
+  # testing installing with pip
+  - pip install -e .
+  - python -c "import pelican; import pelican.tools; import pelican.tests"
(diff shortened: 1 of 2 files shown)


moorepants:
This makes the tarball 414 kB.

.travis.yml
```diff
@@ -29,7 +29,12 @@ before_install:
   - sudo locale-gen fr_FR.UTF-8 tr_TR.UTF-8
 install:
   - pip install tox==2.5.0
-script: tox -e $TOX_ENV
+script:
+  - tox -e $TOX_ENV
+  # testing installing with pip
+  - pip install -e .
+  - python -c "import pelican; import pelican.tools; import pelican.tests"
```
avaris: This will always succeed when you do `-e` install.

moorepants: Can you install the current directory using pip without `-e`? If `-e` is `setup.py develop`, is there an equivalent to `setup.py install`?

avaris: Sure, `pip install .`. But, honestly I'd prefer a new environment without [`usedevelop`](https://github.com/getpelican/pelican/blob/master/tox.ini#L11) and actually running the unittests:
```
python -Wd unittest discover pelican.tests
```

moorepants: I'm not familiar with tox or what usedevelop is. Or what that python command does. Maybe you can recommend the complete changes you'd like to see in this PR? It will take me some time to learn what it all is/does.

justinmayer:
Except for the no-longer-relevant Travis changes, I merged this via: https://github.com/getpelican/pelican/commit/fafe0207ca69ab3bce5a6a8cf6d39200ce9cbc7d

In my testing, this does indeed appear to include the tests in the source distributions. Many thanks to @moorepants for the contribution and to @avaris for reviewing. 💫

PR #2638, opened 2019-10-16
Add docs on how to include other files - fix #1902

### LINKED ISSUE #1902: RFE: Make it possible to include content files from other content files

Lucas-C:
_cf._ discussion in https://github.com/getpelican/pelican/pull/2628

I also included minor code improvements identified in the aforementioned PR.

Preview of the result:

2019-10-16 11_30_07-Writing content — Pelican 4 1 2 documentation

### DIFF
--- a/docs/content.rst
+++ b/docs/content.rst
@@ -382,6 +382,40 @@ to allow linking to both generated articles and pages and their static sources.
 
 Support for the old syntax may eventually be removed.
 
+Including other files
+---------------------
+Both Markdown and reStructuredText syntaxes provide mechanisms for this.
+
+Following below are some examples for **reStructuredText** using `the include directive`_:
+
+    .. code-block:: rst
+
+        .. include:: file.rst
+
+Include a fragment of a file delimited by two identifiers, highlighted as C++ (slicing based on line numbers is also possible):
+
+    .. code-block:: rst
+
+        .. include:: main.cpp
+            :code: c++
+            :start-after: // begin
+            :end-before: // end
+
+Include a raw HTML file (or an inline SVG) and put it directly into the output without any processing:
+
+    .. code-block:: rst
+
+        .. raw:: html
+            :file: table.html
+
+For **Markdown**, one must rely on an extension. For example, using the `mdx_include plugin`_:
+
+    .. code-block:: none
+
+        ```html
+        {! template.html !}
+        ```
+
 
 Importing an existing site
 ==========================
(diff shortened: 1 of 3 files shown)


docs/content.rst
```diff
@@ -382,6 +382,40 @@ to allow linking to both generated articles and pages and their static sources.
...
+Including a fragment of a file delimited by two identifiers, highlighted as C++ (slicing based on line numbers is also possible):
+
+    .. code-block:: none
+
+        .. include:: main.cpp
+            :code: c++
+            :start-after: // begin
+            :end-before: // end
+
+Including a raw HTML file (or an inline SVG or whatnot) and put it directly into the output without any processing:
```
mosra: I didn't expect that my hastily-written comment becomes *the* official documentation :sweat_smile: -- the `whatnot` is maybe a bit too much.

docs/content.rst
```diff
@@ -382,6 +382,40 @@ to allow linking to both generated articles and pages and their static sources.
...
+
+There are some examples for **reStructuredText** using `reStructuredText include`_:
+
+    .. code-block:: none
+
+        .. include:: file.rst
+
+Including a fragment of a file delimited by two identifiers, highlighted as C++ (slicing based on line numbers is also possible):
+
+    .. code-block:: none
```
mosra: Would `.. code-block:: rst` work to highlight this? Not sure about Sphinx in particular, but Pygments at least understand this as reStructuredText. 

Same in other cases.

docs/content.rst
Lucas-C: Sure, I'll remove that ;)

docs/content.rst
Lucas-C: 👍 That works !
Strange because it hadn't found it on http://pygments.org/languages/
But that definitely performs some code highlighting :)

docs/content.rst
```diff
@@ -382,6 +382,40 @@ to allow linking to both generated articles and pages and their static sources.
...
+            :end-before: // end
+
+Including a raw HTML file (or an inline SVG) and put it directly into the output without any processing:
+
+    .. code-block:: rst
+
+        .. raw:: html
+            :file: table.html
+
+For **Markdown**, one must rely on a plugin. For example, using the `mdx_include plugin`_:
```
avaris: nitpick mode on: markdown calls those "extension"s

docs/content.rst
```diff
@@ -382,6 +382,40 @@ to allow linking to both generated articles and pages and their static sources.
...
+
+    .. code-block:: none
```
avaris: RST syntax?

docs/content.rst
mosra: ReST is listed there (and `rest` works too I think) :)

justinmayer:
I made a few minor English-related nitpick suggestions. 😅 

Are the code changes related to including files? If not, normally they'd go in a separate pull request. Even if they are related, I think it might be better if they were at least contained in a separate commit within this PR, if only so the reason and rationale for the changes can be included directly in the commit message for future viewers of this code. 😸

Lucas-C:
@justinmayer : I don't see your other suggestions, but I have split this PR in 2 commits as you suggested

justinmayer:
Many thanks to Lucas for the enhancements and to @mosra and @avaris for reviewing.

justinmayer:
This improvement is included in the just-released [Pelican 4.2](https://github.com/getpelican/pelican/releases/tag/4.2.0). ✨

Issue #2536, opened 2019-02-25
Pelican tables are not mobile-friendly


drboone:
When pelican generates HTML tables from rst inputs, it does not include the "data-label" attribute in the td tags.  On mobile screens with responsive css, this results in card-type table displays with no labels on the fields inside the card.  It would be nice if pelican could add this attribute.

eplata31:
Hi! I am wondering if I could help with this feature?

drboone:
It looks to me like actual generation of <td> tags is in docutils/writers/_html_base.py.  It fills in passed attributes, but it's not yet clear to me how to make data-label get passed.

justinmayer:
Hi Dennis / Enrique. Have either of you made any progress with this endeavor?

eredhan:
comment out of left field...
from a non-Pelican-code understanding.
Is this a bug or feature request?
E.g., did OP try to generate specific output, but Pelican code prevents it?
Reading this made my brain tingle--should this be in SSG code, or part of the site template?
Searching around, a possible short term work around until or instead of changing the python code of Pelican, is to add Javascript to templates to copy HTML table attrs...

See
https://stackoverflow.com/questions/55483961/automatically-append-data-labels-based-content-of-th-to-corresponding-td-for

drboone:
> Hi Dennis / Enrique. Have either of you made any progress with this endeavor?

I have not.

drboone:
> E.g., did OP try to generate specific output, but Pelican code prevents it?

The HTML table is entirely generated by pelican and/or docutils bits based on the rst (for example) input.  There's nothing in the template that can save you here.  You just have the outside body content block.

PR #2642, opened 2019-10-21
Extend stale bot issue closure window

Lucas-C:
I saw that this issue I just been closed by the bot : https://github.com/getpelican/pelican/issues/2536
and I think that's a good example of case where I am a bit sad it did so.
I think the discussion did not really found a satisfying conclusion there.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,5 +1,11 @@
 # Configuration for probot-stale - https://github.com/probot/stale
 
+# Number of days of inactivity before an Issue or Pull Request becomes stale
+daysUntilStale: 60
+
+# Number of days of inactivity before an Issue or Pull Request with the stale label is closed.
+daysUntilClose: 30
+
 # Set to true to ignore issues in a project (defaults to false)
 exemptProjects: true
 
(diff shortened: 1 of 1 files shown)


Lucas-C:
I just updated this PR to make configuration values more explicit, and extend the delay before issue auto-closing time.

I really don't understand why https://github.com/getpelican/pelican/issues/2536 was closed: given [probot/stale](https://github.com/probot/stale#usage) default configuration values, the issue should have been labeled "stale" before closing it0

justinmayer:
Thank you for the contribution, Lucas. The unfortunate reality is that many issues and PRs share an unsatisfying conclusion: they languish, for one reason or another. Marking as stale and auto-closing them simply serves as recognition and acknowledgement of that fact. With regards to the specific issue closed by the bot, I actually think that was an appropriate outcome. When 10 months pass without any notable movement on an issue, it is time to consider putting it to rest. If someone later wants to pick up the torch, the closed issue is waiting and ready to be re-opened.

That all said, I think these configuration changes are potentially useful, so let's try them for a while.

PR #2643, opened 2019-10-25
Add missing striptags filter to simple theme

mxjeff:
``page.title`` are missing ``striptags`` in title block.

Page titles might get some extra html (from typogrify) then html tags end up in title block, ie within ``html head title`` element.


justinmayer:
Thank you for the contribution!

PR #2644, opened 2019-10-28
Namespace plugin implementation

### LINKED ISSUE #2641: Support new namespace plugin format
As discussed in https://github.com/getpelican/pelican-plugins/issues/425#issuecomment-542474603 and other comments in that issue.

avaris:
Fixes #2641

* Creates `pelican.plugins` as a namespace
* Moves plugin related code under `pelican.plugins`
  * `pelican.plugins.signals` is now the location for signals, `pelican.signals` is kept
    for backwards compatibility (and could be deprecated in the future)
  * `pelican.plugins._utils` contains necessary bits for plugin discovery and loading.
    Logic from `Pelican` class is moved here. `Pelican` class now just asks for plugins
    and registers them
* Adds a new command to list installed namespace plugins: `pelican-plugins`
* Contains tests for old and new plugin loading

Behavior should be *mostly* backwards compatible.
- If you have `PLUGINS` setting, things should work as before.
- If you don't have `PLUGINS` setting *and* installed namespace plugins, they will be automatically registered.

TO-DO:
- [x] Documentation update for new plugin format and relevent changes. Something related to this maybe: getpelican/cookiecutter-pelican-plugin#4
- [x] Remove `python2` support. Since native namespace is not supported in `python2`, tests are failing and will fail.


justinmayer:
I tested this implementation with both old- and new-style plugins, and everything described above behaves precisely as expected.

* Pip-installed namespace plugins register without any plugin-related settings present
* Old-style plugins function when explicitly enabled via `PLUGINS` setting
* Both old & new plugins can be used concurrently if all are specified via `PLUGINS` setting
* `pelican-plugins` lists all installed namespace plugins

I'm working on the documentation, in conjunction with the aforementioned migration from individual plugin `CONTRIBUTING` files to a revamped [Contributing](http://docs.getpelican.com/en/stable/contribute.html) section of the Pelican docs.

Regarding Python 2, it is indeed time to remove support for it in Pelican, and I have taken the liberty of creating an issue to track [removing Python 2.7 support](https://github.com/getpelican/pelican/issues/2645). I don't have a strong preference whether that occurs in this PR or a separate one.

Any comments on this pull request from @getpelican/reviewers?

Excellent work on this endeavor, Deniz. Bravo! 👏

oulenz:
Impressive work!

Could you explain what still relies on `pelican.signals`, i.e. why we can't delete it right away? (I admit I still haven't grasped the mechanism behind signals, so the answer may be obvious...)

avaris:
> Could you explain what still relies on `pelican.signals`, i.e. why we can't delete it right away?

Pretty much [every](https://github.com/getpelican/pelican-plugins/blob/master/always_modified/always_modified.py#L5) [single](https://github.com/getpelican/pelican-plugins/blob/master/asciidoc_reader/asciidoc_reader.py#L11) [plugin](https://github.com/getpelican/pelican-plugins/blob/master/assets/assets.py#L22) :)... I moved the signals under `pelican.plugins` because that seemed like the logical place for them. But that also means we also need to keep `pelican.signals` until plugins adjust their imports. I don't have a strong opinion about moving `signals` (it can stay as is), but since we are going for a restructure in plugins this sounded like a good opportunity to slight logical reorganization (to me at least) in the core too.

*PS*: tl;dr, on signals: signals is how `pelican` communicates with `plugins`. `plugins` register callback functions to certain "signals", then `pelican` emits/sends signals at certain stages of the run and callbacks from plugins will do as they please.

avaris:
Oh... I totally missed the import in `pelican/__init__.py`. Thanks for pointing that out. I'll fix that. I think with that changed, plugins can stay as is.

This should definitely work:
```python
from pelican import signals
```

I am not 100% sure but this would also work I believe:
```python
from pelican.signals import some_signal
```

oulenz:
I've also been able to test this now, and new-style and old-style plugins, as well as a mixture of both all work for me!

justinmayer:
@avaris: Now that Python 2 support has been removed, would you mind resolving the merge conflicts in this PR and rebasing on current master? Looking forward to getting this merged. 😁

avaris:
Yes, I was expecting conflicts and waiting for the merge :).

avaris:
rebase, cleanup and documentation update: done

docs/plugins.rst
```diff
@@ -9,16 +9,31 @@ features to Pelican without having to directly modify the Pelican core.
...
-    PLUGINS = ['package.myplugin',]
+If you leave the ``PLUGINS`` setting as default (``None``), Pelican will then
+collect the namespace plugins and register them. If on the other hand you
+specify a ``PLUGINS`` settings as a list of plugins, this autodiscovery will
+be disabled and only listed plugins will be registered and you will have to
+explicitly list the namespace plugins as well.
 
-Alternatively, another method is to import them and add them to the list::
+If you are using ``PLUGINS`` setting, you can specify plugins in two ways.
+First method is using strings to the plugins. Namespace plugins can be
```
justinmayer: First method is using strings to the plugins. → The first method specifies plugins as a list of strings.

docs/plugins.rst
```diff
@@ -36,11 +51,12 @@ the ``PLUGIN_PATHS`` list can be absolute or relative to the settings file::
 
 Where to find plugins
 =====================
+Namespace plugins can be found in the `pelican-plugins organization`_ as
+individual repositories. Legacy plugins are collected in the `pelican-plugins
+repository`_ and they will be slowly phased out to the namespace versions.
```
justinmayer: […] phased out to the namespace versions → phased out in favor of the namespace versions

docs/plugins.rst
```diff
@@ -70,6 +86,33 @@ which you map the signals to your plugin logic. Let's take a simple example::
     your ``register`` callable or they will be garbage-collected before the
     signal is emitted.
 
+Namespace Plugin structure
```
justinmayer: `Namespace Plugin structure` → `Namespace plugin structure` (to maintain consistent capitalization)

docs/plugins.rst
```diff
@@ -70,6 +86,33 @@ which you map the signals to your plugin logic. Let's take a simple example::
...
+--------------------------
+
+Namespace plugins must adhere to a certain structure in order to function
+properly. They need to installable (i.e. contain ``setup.py`` or equivalent)
```
justinmayer: `They need to installable` → `They need to be installable`

docs/plugins.rst
```diff
@@ -70,6 +86,33 @@ which you map the signals to your plugin logic. Let's take a simple example::
...
+    myplugin
+    ├── pelican
+    │   └── plugins
+    │       └── myplugin
+    │           ├── __init__.py
+    │           └── ...
+    ├── ...
+    └── setup.py
+
+It is crucial that ``pelican`` or ``pelican/plugins`` folder **should not**
```
justinmayer: `**should not**` → `**not**` (already implied via "crucial")

docs/plugins.rst
```diff
@@ -70,6 +86,33 @@ which you map the signals to your plugin logic. Let's take a simple example::
...
+contain an ``__init__.py`` file. In fact, it is best to have those folders
+empty besides the listed folders in the above structure and contain your
+plugin related files solely in the ``pelican/plugins/myplugin`` folder to
+avoid any issues.
+
```
justinmayer: `contain your plugin related files` → `keep your plugin-related files contained`

docs/plugins.rst
justinmayer: I would probably remove the word "should" altogether, but that's admittedly a very minor nitpick. 😅

avaris: Done :)

Issue #2645, opened 2019-10-29
Remove Python 2.7 support


justinmayer:
[Official support for Python 2.7 ends in two months](https://pythonclock.org/), after which no bug fixes or security updates will be released. In preparation for that, it is time to remove code from Pelican related to Python 2.7, as well as removing Python 2.7 from the CI test matrix.

The following checklist tracks the progress of this endeavor:

- [x] sign the [pledge to drop Python 2](https://python3statement.org/)
- [x] remove Python 2 support in `.travis.yml`
- [x] remove Python 2 classifiers in `setup.py`, `pyproject.toml`, and `poetry.lock`
- [x] remove `six` from the code
- [x] remove `utils.python_2_unicode_compatible`
- [x] update `CONTRIBUTING.rst` and`docs/` accordingly
- [x] remove comments related to Python 2 (e.g., in `settings.py` & `tools/pelican_import.py`)

Lucas-C:
Great plan !
Maybe a checklist of things to do in order to achieve this would be helpful ?
A few items from the top of my head, completing the one you already mentioned:
- [ ] remove support in `.travis.yml`
- [ ] remove Python 2 classifiers in `setup.py` & `pyproject.toml` & `poetry.lock`
- [ ] remove uses of `six` in the code
- [ ] remove `utils.python_2_unicode_compatible`
- [ ] update `CONTRIBUTING.rst` & `docs/` accordingly
- [ ] remove Python 2 specific comments (I found 2, in `settings.py` & `tools/pelican_import.py`)

justinmayer:
Agreed, @Lucas-C. I've moved that checklist to the issue description. Many thanks for writing that up! 👏 

Unless there are any objections, I will submit a pull request to the repo for https://python3statement.org/ in order to pledge our solidarity with the other projects that have committed to dropping support for Python 2 by the end of the year. (cc: @getpelican/reviewers)

I'm working on making the appropriate changes to the `CONTRIBUTING` and `docs/contributing.rst` files. Would anyone like to submit pull requests for the other tasks on the checklist above?

pauloxnet:
I tried to contribute to the Python 2.7 removal with a small PR #2647

avaris:
I was planning to work on this this weekend (at least CI and code cleaning bits) as part of #2644, mainly because I don't like submitting PRs without a green check mark :).

iKevinY:
[...]

PR #2647, opened 2019-11-05
Remove Python 2.7 support from settings

pauloxnet:
As requested by @justinmayer  in #2645 I've just opened a PR to remove Python 2.7 support from some settings.

I'm not familiar with poetry, so I left `poetry.lock` as is.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,6 @@ env:
   matrix:
     - TOX_ENV=docs
     - TOX_ENV=flake8
-    - TOX_ENV=py27
     - TOX_ENV=py35
     - TOX_ENV=py36
 matrix:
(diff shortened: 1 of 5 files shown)


Lucas-C:
LGTM 👍

pauloxnet:
Do you think we can merge this ?

justinmayer:
The only reason I haven't merged this yet is that @iKevinY included your commit in his PR (linked above), and I wasn't sure if merging this now would create merge conflicts for his PR. But perhaps because the commit would exist in both master and Kevin's PR, there would be no such conflict. Anyone know for sure?

avaris:
It should be fine, but if it's OK with @pauloxnet we can continue from #2649 to keep things in one place. It already contains the commit, not much would change.

pauloxnet:
For me it's ok.

iKevinY:
> I wasn't sure if merging this now would create merge conflicts for his PR.

Sorry about that! I thought this PR would be merged in sooner, and I was going to just do a quick rebase for my PR, but I'm okay with just collaboratively working on the larger PR now as well. 😛

PR #2649, opened 2019-11-06
Remove Python 2 support

iKevinY:
This PR is a first-pass at removing Python 2 support for Pelican, as described in the #2645 metabug. Specifically, I went through and removed all instances of `six`, as well as `utils.python_2_unicode_compatible`. While I did try to remove Python 2-specific logic whenever I saw it, I wasn't searching carefully for it (this will likely require a finer comb through the codebase), nor has documentation been updated either.

(I pulled in #2647 so that Travis wouldn't complain about failing Python 2 builds. 😅)

### DIFF
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -1,10 +1,6 @@
 # -*- coding: utf-8 -*-
 
 import argparse
-try:
-    import collections.abc as collections
-except ImportError:
-    import collections
 import logging
 import multiprocessing
 import os
@@ -12,6 +8,7 @@ import pprint
 import sys
 import time
 import traceback
+from collections.abc import Iterable
 
 # pelican.log has to be the first pelican module to be loaded
 # because logging.setLoggerClass has to be called before logging.getLogger
@@ -184,7 +181,7 @@ class Pelican(object):
         for pair in signals.get_generators.send(self):
             (funct, value) = pair
 
-            if not isinstance(value, collections.Iterable):
+            if not isinstance(value, Iterable):
                 value = (value, )
 
             for v in value:
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -1,12 +1,10 @@
 # -*- coding: utf-8 -*-
-from __future__ import print_function, unicode_literals
 
 import argparse
 try:
     import collections.abc as collections
 except ImportError:
     import collections
-import locale
 import logging
 import multiprocessing
 import os
@@ -14,8 +12,6 @@ import pprint
 import sys
 import time
 import traceback
-
-import six
 
 # pelican.log has to be the first pelican module to be loaded
 # because logging.setLoggerClass has to be called before logging.getLogger
@@ -76,11 +72,10 @@ class Pelican(object):
             sys.path.insert(0, pluginpath)
         for plugin in self.settings['PLUGINS']:
             # if it's a string, then import it
-            if isinstance(plugin, six.string_types):
+            if isinstance(plugin, str):
                 logger.debug("Loading plugin `%s`", plugin)
                 try:
-                    plugin = __import__(plugin, globals(), locals(),
-                                        str('module'))
+                    plugin = __import__(plugin, globals(), locals(), 'module')
                 except ImportError as e:
                     logger.error(
                         "Cannot load plugin `%s`\n%s", plugin, e)
@@ -375,15 +370,6 @@ def get_config(args):
         config['BIND'] = args.bind
     config['DEBUG'] = args.verbosity == logging.DEBUG
 
-    # argparse returns bytes in Py2. There is no definite answer as to which
-    # encoding argparse (or sys.argv) uses.
-    # "Best" option seems to be locale.getpreferredencoding()
-    # http://mail.python.org/pipermail/python-list/2006-October/405766.html
-    if not six.PY3:
-        enc = locale.getpreferredencoding()
-        for key in config:
-            if key in ('PATH', 'OUTPUT_PATH', 'THEME'):
-                config[key] = config[key].decode(enc)
     return config
 
 
@@ -397,7 +383,7 @@ def get_instance(args):
     settings = read_settings(config_file, override=get_config(args))
 
     cls = settings['PELICAN_CLASS']
-    if isinstance(cls, six.string_types):
+    if isinstance(cls, str):
         module, cls_name = cls.rsplit('.', 1)
         module = __import__(module)
         cls = getattr(module, cls_name)
--- a/pelican/log.py
+++ b/pelican/log.py
@@ -1,17 +1,9 @@
 # -*- coding: utf-8 -*-
-from __future__ import print_function, unicode_literals
 
-import locale
 import logging
 import os
 import sys
 from collections import defaultdict
-try:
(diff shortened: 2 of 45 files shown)


pelican/utils.py
```diff
@@ -51,79 +49,6 @@ def sanitised_join(base_directory, *parts):
     return joined
 
 
-def strftime(date, date_format):
```
avaris: I have two reservations for simply removing this:
* It had extra functionality (see: #1446)
* It might be breaking if plugins/themes/etc were using these (similarly `SafeDatetime`)

poetry.lock
```diff
@@ -362,7 +271,7 @@ description = "Python 2 and 3 compatibility utilities"
 name = "six"
 optional = false
 python-versions = ">=2.6, !=3.0.*, !=3.1.*"
-version = "1.12.0"
+version = "1.13.0"
```
avaris: Why didn't this go away? Some other package is depending on it?

setup.py
```diff
@@ -26,8 +25,6 @@
 CHANGELOG = open('docs/changelog.rst', encoding='utf-8').read()
 
 description = u'\n'.join([README, CHANGELOG])
```
avaris: `u'...'` :thinking:

poetry.lock
iKevinY: I'm guessing...this is the lock file generated after I ran `poetry remove six`. ¯\\_(ツ)_/¯

setup.py
iKevinY: 😅

pelican/__init__.py
```diff
@@ -76,7 +72,7 @@ def init_plugins(self):
             sys.path.insert(0, pluginpath)
         for plugin in self.settings['PLUGINS']:
             # if it's a string, then import it
-            if isinstance(plugin, six.string_types):
+            if isinstance(plugin, str):
                 logger.debug("Loading plugin `%s`", plugin)
                 try:
                     plugin = __import__(plugin, globals(), locals(),
```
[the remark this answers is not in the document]
iKevinY: Nice catch, thanks!

pelican/log.py
```diff
@@ -159,16 +153,13 @@ def _log(self, level, msg, args, exc_info=None, extra=None):
...
         localized systems have errors in native language if locale is set
         so convert the message to unicode with the correct encoding
         '''
         if isinstance(arg, Exception):
             text = str('%s: %s') % (arg.__class__.__name__, arg)
-            if six.PY2:
-                text = text.decode(self._exc_encoding)
             return text
         else:
             return arg
```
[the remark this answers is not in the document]
iKevinY: Done. 👍

pelican/utils.py
iKevinY: Ah, I got overzealous and figured it was unused after glancing at the docstring. Reverted the commit that removed these; I'll take another pass through it to just remove the Python 2 logic.

iKevinY:
Thanks for the initial review @avaris! I'll take a closer pass through the codebase again tonight to remove some more of the subtle Python 2 stuff.

poetry.lock
avaris: `poetry` probably knows its stuff :+1: :)

justinmayer:
@avaris: Any chance you might be able to go through the code in detail this weekend? It'd be great to get this PR, and your namespace plugin PR, merged soon. 😅

avaris:
@justinmayer that's the plan :).

avaris:
Below are `Python2` related code. These might include parts touched by this PR but not completely cleaned. I'm putting them all here for completeness sake, instead of adding comments to lines.

- https://github.com/getpelican/pelican/blob/bae6de5d26646872fedbd00d37a849bba1837a3b/pelican/settings.py#L18-L30
- Unnecessary: https://github.com/getpelican/pelican/blob/bae6de5d26646872fedbd00d37a849bba1837a3b/pelican/generators.py#L9
- Other `codecs` imports: 
    - `tests/test_importer.py`
    - `tests/test_generators.py`
    - `tools/pelican_import.py`
    - `tools/pelican_quickstart.py`
- This is kind of the opposite (py3 only code :) ) but it is fixed in [`docutils 0.15`](http://docutils.sourceforge.net/HISTORY.html#release-0-15-2019-07-20), should be fine to remove if we fix dependecy: https://github.com/getpelican/pelican/blob/bae6de5d26646872fedbd00d37a849bba1837a3b/pelican/readers.py#L206-L217
- Currently, this is changed to `str(...)` but `str`ifying is not needed at all: https://github.com/getpelican/pelican/blob/bae6de5d26646872fedbd00d37a849bba1837a3b/pelican/urlwrappers.py#L69
- `from cgi ...` is Python2-only: https://github.com/getpelican/pelican/blob/bae6de5d26646872fedbd00d37a849bba1837a3b/pelican/utils.py#L34-L37
- `super.__init__(convert_charrefs=False)` should be fine for all this: https://github.com/getpelican/pelican/blob/bae6de5d26646872fedbd00d37a849bba1837a3b/pelican/utils.py#L461-L467
- Somewhat redundant now: https://github.com/getpelican/pelican/blob/bae6de5d26646872fedbd00d37a849bba1837a3b/pelican/utils.py#L637-L642
- This looks a lot like [`os.makedirs`](https://docs.python.org/3/library/os.html#os.makedirs) with the new argument `exist_ok=True`: https://github.com/getpelican/pelican/blob/bae6de5d26646872fedbd00d37a849bba1837a3b/pelican/utils.py#L859-L864
- Don't need the `except` part: https://github.com/getpelican/pelican/blob/bae6de5d26646872fedbd00d37a849bba1837a3b/pelican/tools/pelican_import.py#L24-L28

**PS:** I didn't go through the tests yet, but I need a break :).

avaris:
Bonus: I didn't verify but, I *think* all the magic in `utils.pelican_open` is handled by built-in `open`: https://github.com/getpelican/pelican/blob/bae6de5d26646872fedbd00d37a849bba1837a3b/pelican/utils.py#L256-L266

pauloxnet:
I think all your comments are correct. Maybe you can push a commit with all the edits and we can see how tests  say and then we can eventually comment the code.

avaris:
I pushed the commits for the changes (+ a few more I noticed while doing them) I mentioned.

avaris:
Added one more commit to clean up `super()` calls.

iKevinY:
Thank you for adding to the PR @avaris and to @pauloxnet for the added review!

justinmayer:
@avaris & @iKevinY: Do you think this PR is ready to merge? Just wanted to get your sign-off before merging. 😊

iKevinY:
I'm good to merge; not sure if you'd like me to rebase first (to at least squash my commits together).

justinmayer:
@iKevinY: Probably a good idea. Would you mind rebasing and squashing into logical commits?

iKevinY:
@justinmayer Done, let me know if you'd like anything else changed!

pauloxnet:
Look good to me

justinmayer:
Many thanks to @pauloxnet , @iKevinY , and @avaris for your work on this endeavor. Exciting to see the streamlined code! 💫

PR #2651, opened 2019-11-09
Remove Python 2.x references from docs

justinmayer:
In addition to removing Python 2.x references from the Pelican documentation, this PR also:

* adds new development tool configuration: Invoke tasks, Pre-commit, and EditorConfig
* updates the documentation to mention Poetry and take above the tools into account
* adds a Contributing section for plugin-related development

Looking forward to comments from @getpelican/reviewers. Refs #2645


avaris:
Do we want to convert the `tox/CI` tests to `pytest` as well?

justinmayer:
Sure. Would anyone like to lend a hand with that endeavor?

avaris:
Adjusting [this part from `tox.ini`](https://github.com/getpelican/pelican/blob/master/tox.ini#L12-L22) for `pytest` instead of `nose` should do it I believe:

```ini
deps =
    -rrequirements/test.pip
    pytest
    pytest-cov
    coveralls
    pygments==2.1.3

commands =
    {envpython} --version
    pytest -sv --cov=pelican pelican
    - coveralls
```

**Q**: Do we use `coveralls` at all? As I remember, that part fails and has been like that for a loong time. If we don't use it, those (`deps` and `commands` parts for that) could go away.
```
Submitting coverage to coveralls.io...
Could not submit coverage: 422 Client Error: Unprocessable Entity for url: https://coveralls.io/api/v1/jobs
```

justinmayer:
Thanks for the suggested changes, Deniz. Coveralls has been busted for a long time, and I've been meaning to remove it.

I added one commit that replaces Nose with Pytest in `tox.ini` and another commit that removes Coveralls.

docs/contribute.rst
```diff
@@ -7,72 +7,78 @@ can also help out by reviewing and commenting on
...
-    $ pip install -e .
+Development
+===========
 
-To conveniently test on multiple Python versions, we also provide a ``.tox``
-file.
+Once Pelican has been set up for local development, create a topic branch for
+your bug fix or feature:
 
+    git checkout -b name-of-your-bugfix-or-feature
```
avaris: Needs one extra `:` after `feature`, Otherwise looks like quote(?).

docs/contribute.rst
```diff
@@ -82,11 +88,11 @@ the nature of your changes, then you should update the output used by the
...
 
-    $ LC_ALL=en_US.utf8 pelican -o pelican/tests/output/custom/ \
+    LC_ALL=en_US.utf8 pelican -o pelican/tests/output/custom/ \
         -s samples/pelican.conf.py samples/content/
-    $ LC_ALL=fr_FR.utf8 pelican -o pelican/tests/output/custom_locale/ \
+    LC_ALL=fr_FR.utf8 pelican -o pelican/tests/output/custom_locale/ \
         -s samples/pelican.conf_FR.py samples/content/
-    $ LC_ALL=en_US.utf8 pelican -o pelican/tests/output/basic/ \
+    LC_ALL=en_US.utf8 pelican -o pelican/tests/output/basic/ \
         samples/content/
```
avaris: Maybe we can make an invoke task for this?

docs/contribute.rst
```diff
@@ -101,48 +107,68 @@ environments.
...
+Building the docs
+-----------------
+
+If you make changes to the documentation, you should preview your changes
+before committing them::
+
+    cd docs
+    make html
+
+Open ``_build/html/index.html`` in your browser to preview the documentation.
```
avaris: Again, invoke? Since we are using it, might as well use it a lot :). Potentially useful: [built-in `webrowser` module](https://docs.python.org/3.8/library/webbrowser.html).

docs/contribute.rst
justinmayer: Nice catch! Fixed via 703c2810.

docs/contribute.rst
justinmayer: Good idea. I added an Invoke task for updating functional test output in b7368f91.

docs/contribute.rst
justinmayer: I had planned on adding an Invoke task for building & serving documentation at a later point, but I agree that I might as well add it now, so I did. 😅  I took a different approach in 535df9cd, allowing for real-time browser refresh upon file modification via `livereload`.

justinmayer:
@avaris: Great suggestions! I believe I addressed all of them. Any further comments on this PR?

justinmayer:
Many thanks for reviewing, Deniz. Very helpful! 😁

PR #2658, opened 2019-11-30
fix: tag like C++ and C# are treated as a same tag

talha131:
## Problem

Create an article. Use tags `C++` and `C#`.

A single article has used the tags `C++` and `C#`. Therefore count of articles in these tags should be one. Instead you get 2 articles for each of these tags. The same article is added twice.

## Why it happens?

The problem starts from [this line](https://github.com/getpelican/pelican/blob/master/pelican/generators.py#L654).

```python
                    self.tags[tag].append(article)
```

Keys tag `C++` and tag `C#` overwrite each other.

Why do they overwrite? Url_wrapper class uses [this regex](https://github.com/getpelican/pelican/blob/master/pelican/settings.py#L147) to create hash of the object.

```python
        (r'[^\w\s-]', ''),  # remove non-alphabetical/whitespace/'-' chars
```

This regex removes `+` and `#` from a tag. Hence `C++` and `C#` both become `C`.

## Solution

This patch changes the regex to allow `+` and `#` characters.
slug method is also used to get the filename and url. Both of these characters are allowed in filenames and in URLs.

---
<sup>Related to https://github.com/Pelican-Elegant/elegant/issues/516</sup>

### DIFF
--- a/pelican/settings.py
+++ b/pelican/settings.py
@@ -144,7 +144,9 @@ DEFAULT_CONFIG = {
     'TEMPLATE_EXTENSIONS': ['.html'],
     'IGNORE_FILES': ['.#*'],
     'SLUG_REGEX_SUBSTITUTIONS': [
-        (r'[^\w\s-]', ''),  # remove non-alphabetical/whitespace/'-' chars
+        # remove all characters except alphanumeric, whitespace,
+        # '_', '-', '+', '#' characters
+        (r'[^\w\s\-#\+]', ''),
         (r'(?u)\A\s*', ''),  # strip leading whitespace
         (r'(?u)\s*\Z', ''),  # strip trailing whitespace
         (r'[-\s]+', '-'),  # reduce multiple whitespace or '-' to single '-'
@@ -387,7 +389,7 @@ def handle_deprecated_settings(settings):
 
                 if replace:
                     regex_subs += [
-                        (r'[^\w\s-]', ''),
+                        (r'[^\w\s\-#\+]', ''),
                         (r'(?u)\A\s*', ''),
                         (r'(?u)\s*\Z', ''),
                         (r'[-\s]+', '-'),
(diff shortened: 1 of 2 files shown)


avaris:
> Both of these characters are allowed in filenames and in URLs.

Any reference for that?

As far as I know. in URLs:
*  `+` is usually encoded version of `space`, i.e. `hello world` -> `hello+world`
* `#` denotes an anchor inside the page, i.e. `example.com/foo.html#bar` -> go to `bar` anchor at `foo.html`

talha131:
Good points.

`+` has special meaning in the query part of the URL. Hence following urls are valid



For `#`, themes have to use Jinja method [`urlencode`](https://jinja.palletsprojects.com/en/2.10.x/templates/#urlencode).



I can make this change in the default themes.

avaris:
> For #, themes have to use Jinja method urlencode.

Umm, that's not ideal and possibly wrong. For example, in `elegant` theme you seem to be forwarding `tag` urls to `tags.html#foo` style links. Granted, you don't do it in `tag.url` but you could have. If that was the case you would *not* want to encode that.

I still think this is a non-issue. It's not like the machinery is hidden in the code. [This is an exposed setting with a default that has a detailed explanation of what it does](http://docs.getpelican.com/en/stable/settings.html#url-settings). If user doesn't like this, they can simply override it in their settings.

talha131:
> Umm, that's not ideal and possibly wrong. 

I see your point. Also, this patch may break existing themes and websites. Pay off is very small compared to the benefit.

----

For now, I have added following snippet to my personal `pelicanconf.py` and it is working for my blog.

```python
SLUG_REGEX_SUBSTITUTIONS = [
    # remove all characters except alphanumeric, whitespace,
    # '_', '-', '+', '#' characters
    (r'[^\w\s\-#\+]', ''),
    (r'(?u)\A\s*', ''),  # strip leading whitespace
    (r'(?u)\s*\Z', ''),  # strip trailing whitespace
    (r'[-\s]+', '-'),  # reduce multiple whitespace or '-' to single '-'
]
```

I will keep an eye out to see if any URL breaks down.

----

@avaris Thank you for your time and reviewing my PR.

PR #2659, opened 2019-12-03
Support --exclude-from for rsync publishing using excludes.txt

dcantrell:
Add an excludes.txt example file where users can list rsync exclude
specifications.  If the site you are publishing has any paths on the
remote site that Pelican should ignore and not remove, list them in
excludes.txt so they are not removed each time you do:

    make rsync_upload

The example file has a comment pointing users to the rsync man page
for information on the format used in exclude files.  The only entry
in it by default is the .DS_Store entry which was from the rsync
command in tasks.py.

### DIFF
--- a/pelican/tools/pelican_quickstart.py
+++ b/pelican/tools/pelican_quickstart.py
@@ -347,6 +347,14 @@ needed by Pelican.
                 fd.close()
         except OSError as e:
             print('Error: {0}'.format(e))
+        try:
+            with open(os.path.join(CONF['basedir'], 'excludes.txt'),
+                      'w', 'utf-8') as fd:
+                _template = _jinja_env.get_template('excludes.txt.jinja2')
+                fd.write(_template.render(**CONF))
+                fd.close()
+        except OSError as e:
+            print('Error: {0}'.format(e))
 
     print('Done. Your new project is available at %s' % CONF['basedir'])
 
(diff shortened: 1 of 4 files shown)


avaris:
Two comments:

1) `excludes.txt` is a bit too generic for a name. Something like `rsync_excludes.txt` would be better for understanding what the file is without opening it.
2) Does it really need a jinja template of its own? It's all static text. You can easily write it from a string.

dcantrell:
> Two comments:
> 
>     1. `excludes.txt` is a bit too generic for a name. Something like `rsync_excludes.txt` would be better for understanding what the file is without opening it.

Makes sense, I can change that.

>     2. Does it really need a jinja template of its own? It's all static text. You can easily write it from a string.

Initially I didn't make it a jinja template, but I decided to follow what the other template files were doing.  I don't know what jinja is anyway.  I can just drop all that.

Changes coming up.

justinmayer:
Hey David. I keep looking at this PR from time to time, and I guess my question is what is the impetus for storing the excludes outside `tasks.py` / `Makefile`? While I understand the DRY concept, if that's the reason I'm not sure it makes a lot of sense in this case given that most folks are only going to ignore a handful of file types. In the end, I'm just not sure it makes sense to create a separate file for this purpose. It seems like it creates unnecessary clutter.

Am I missing something? Is it just me?

dcantrell:
The main objective is to allow rsync exclude pattern syntax in the external excludes list and avoid string manipulation problems that may come about while writing an excludes list as a string in Python or in a Makefile or something like that.  The other objective I had was to permit site-specific customization in a file not owned by the upstream project to allow for easier merging/rebasing of upstream code from Pelican to your own local project.

PR #2661, opened 2019-12-06
Update links to HTTPS and current 301 redirects in docs/templates/themes

kurtmckee:
This also updates "example.com" -> "domain.example", and updates the
Tumblr API to use HTTPS as documented in the current Tumblr API docs.


justinmayer:
Some good changes in here. Two that I’m not sure about:

1. example.com actually routes to an example page — domain.example does not. I prefer the former behavior.

2. getpelican.com links are preferred over blog.getpelican.com. Yes, the bare domain currently redirects to blog.getpelican.com, but that may not always be the case going forward.

kurtmckee:
> Some good changes in here.

Thanks! I see that the checks are failing and I've been working on that. Unfortunately I'm experiencing a train wreck of CRLF-related issues on Windows. I've forced git to use 100% LF-only and have confirmed that the whole repo is LF-only but then started receiving git error messages: "CRLF will be replaced by LF". Quite the opposite of what I was expecting. Also, somehow the feeds are still outputting as CRLF and are failing the git diff checks. `_rmtree_unsafe()` is also throwing PermissionErrors because the Python process is still using the output directories when it's trying to remove them...what a mess. I'm going to fix this as much as possible and then rely on Travis-CI to confirm that everything is working.

> 1. example.com actually routes to an example page — domain.example does not. I prefer the former behavior.

Okay, I'll fix that.

> 2. getpelican.com links are preferred over blog.getpelican.com. 

Okay, I'll fix that, too.

kurtmckee:
@justinmayer, thanks for reviewing this! I used grep and sed to fix notes 1 and 2, and the build is now passing according to Travis CI. Please let me know if this needs further adjustment.

kurtmckee:
Updated to resolve a recent merge conflict.

justinmayer:
@avaris: Would you mind taking a look at this PR, particularly the two test-related modifications?

justinmayer:
Many thanks, Kurt, for all your efforts here. This is an extremely helpful contribution, and I am very grateful for your assistance. Much appreciated! 🙇

PR #2664, opened 2019-12-16
Remove CSS references to the non-existent digg.png icon

kurtmckee:
`main.css` in the notmyidea theme contains a reference to a non-existant digg.png icon. I searched the git history and couldn't find a reference to it so I wasn't able to restore it (assuming it ever existed).

I've removed the line in the CSS that references the file and updated the unit test output files. Please let me know if there's anything else required. Thanks!

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+Release type: patch
+
+Remove the reference to the non-existent digg.png icon in main.css.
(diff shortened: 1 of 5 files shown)


kurtmckee:
Fixed the "non-existant" typo.

justinmayer:
Many thanks to @kurtmckee for the fix and to @iKevinY for reviewing. Merged via 4db9b944a237302b5558395038f6125a910c6a05.

Issue #2608, opened 2019-08-17
Pelican 4.1.0 - quickstart generates project in root of venv, rather than in project dir


lesatdgl:
## run pelican-quickstart
---
Welcome to pelican-quickstart v4.1.0.
This script will help you create a new Pelican-based website.

Please answer the following questions so this script can generate the files
needed by Pelican.

Using project associated with current virtual environment.Will save to:
/home/les/dgl-home/dglEnvironments/pelican-projects
  <skip dialog>
Done. Your new project is available at /home/les/dgl-home/dglEnvironments/pelican-projects
...

### Behavior of pelican-quickstart is different from previous releases - 
Using project associated with current virtual environment.Will save to: /home/les/dgl-home/dglEnvironments/pelican-projects

lesatdgl:
Hello? Is this new behavior intended?

avaris:
Looking at the code associated with this, no, not exactly new: 7-8 years :).

If it detects you're in a virtualenv, it uses the env root. You can pass `--path` to override it.

lesatdgl:
Is that anywhere in the docs? If so, I missed it

Les Dunaway
770-490-7546

lesatdgl:
Is that anywhere in the docs? If so, I missed it
This was not the behavior in the past, in venv

Les Dunaway
770-490-7546

lesatdgl:
Thanks, is that useful or best to focus on docs for current release?

Les Dunaway
770-490-7546



On Fri, Sep 13, 2019, at 02:13, Deniz Turgut wrote:
> You can install a specific version with `pip` by specifying the version after `==`. e.g. To install `pelican` version 3.5:

> `pip install pelican==3.5
`

justinmayer:
Hi Les. Pelican's quick-start script looks for a `$VIRTUALENV/.project` file, and if it finds one, the script will automatically use the directory path contained within that file as the destination directory for the to-be-generated new Pelican project.

As @avaris pointed out via the blame lines noted above, Pelican's behavior regarding this topic has not changed since at least 2013. So for the most part, "it has always been this way" is in fact an accurate statement. The most likely explanation is that Pipenv's behavior changed, which it has done to an [infamous degree](https://chriswarrick.com/blog/2018/07/17/pipenv-promises-a-lot-delivers-very-little/).

[...]

PR #2675, opened 2019-12-28
Docs: Add info about pelican-quickstart command path flag.

carlosperate:
Resolves https://github.com/getpelican/pelican/issues/2608.

Adds a short paragraph about the `path` flag for the `pelican-quickstart` command to save the new project into a different directory than the venv one.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -102,9 +102,12 @@ your site::
 
     pelican-quickstart
 
-If it is running inside a virtual environment it will try to detect the path
-associated to it and save the project there. A custom path can be configured
-using the ``--path`` flag.
+If run inside an activated virtual environment, ``pelican-quickstart`` will
+look for an associated project path inside ``$VIRTUAL_ENV/.project``. If that
+file exists and contains a valid directory path, the new Pelican project will
+be saved at that location. Otherwise, the default is the current working
+directory. To set the new project path on initial invocation, use:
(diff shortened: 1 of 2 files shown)


docs/install.rst
```diff
@@ -102,6 +102,10 @@ your site::
 
     pelican-quickstart
 
+If it is running inside a virtual environment it will try to detect the path
+associated to it and save the project there. A custom path can be configured
+using the ``--path`` flag.
+
```
justinmayer: In the interest of extra clarity, what do you think about this proposed alternate text?
> If run inside an activated virtual environment, `pelican-quickstart` will look for an associated project path inside `$VIRTUAL_ENV/.project`. If that file exists and contains a valid directory path, the new Pelican project will be saved at that location. Otherwise, the default is the current working directory. To set the new project path on initial invocation, use: `pelican-quickstart --path /your/desired/directory`

justinmayer:
Thank you for the contribution, @carlosperate, as well as your patience. I suggested some more detailed language to hopefully clear up any potential confusion. Please let me know what you think. 😄

docs/install.rst
carlosperate: Yeah, that sounds good!
I've updated the text in 66818be4418c813ab939efc02f63d7026ef80998, do you want me to squash the commits?

justinmayer: Thanks! No need — I'll take it from here.

PR #2682, opened 2020-01-13
Fix the LOG_FILTER setting

### LINKED ISSUE #2552: LOG_FILTER not working as it should?
Good evening folks,

I have a problem with LOG_FILTER. According to the docs, one can remove `TAG_SAVE_AS is set to False` by configuring `LOG_FILTER` this way:

```python
import logging
LOG_FILTER = [(logging.WARN, 'TAG_SAVE_AS is set to False')]
```

So to ignore `AUTHOR_SAVE_AS is set to False` and `CATEGORY_SAVE_AS is set to False`, it must be :

```python
import logging
LOG_FILTER = [
    (logging.WARN, 'AUTHOR_SAVE_AS is set to False'),
    (logging.WARN, 'CATEGORY_SAVE_AS is set to False')
]
```
Right?

So this is what I did:
```bash
$ head pelicanconf.py -n 20
#!/usr/bin/env python
# -*- coding: utf-8 -*- #
import datetime
import logging

import pelican

PELICAN_VERSION = pelican.__version__

[...]

LOG_FILTER = [
    (logging.WARN, 'AUTHOR_SAVE_AS is set to False'),
    (logging.WARN, 'CATEGORY_SAVE_AS is set to False')
]

AUTHOR_SAVE_AS = False
AUTHORS_SAVE_AS = False
TAG_SAVE_AS = False
TAGS_SAVE_AS = False
CATEGORY_SAVE_AS = False
CATEGORIES_SAVE_AS = False
ARCHIVES_SAVE_AS = False

$ make publish
pelican /mnt/c/Users/max/Code/mlcdf/content -o /mnt/c/Users/max/Code/mlcdf/output -s /mnt/c/Users/max/Code/mlcdf/publishconf.py
WARNING: CATEGORY_SAVE_AS is set to False
WARNING: AUTHOR_SAVE_AS is set to False
Done: Processed 2 articles, 0 drafts, 4 pages, 1 hidden page and 0 draft pages in 0.52 seconds.
```
2 things here:
- it didn't work: `AUTHOR_SAVE_AS is set to False` and `CATEGORY_SAVE_AS is set to False` are still logged.
- and, you may have noticed that, despite setting `TAG_SAVE_AS = False`, and not filtering it, I did not see `TAG_SAVE_AS is set to False` in the logs.

Using the template to filter worked (but it will also filter out other logs matching the template which is not what you would necessary want):
```
import logging
LOG_FILTER = [
    (logging.WARN, '%s is set to %s'),
]
```

* Am I missing something?
* What's the intended behaviour here? Is there an actual bug in the code? Or is the code example in the docs just wrong?

I'm a bit lost. 🙃 

Maxime

2pxsolidblack:
Fixes #2552

### DIFF
--- a/pelican/tests/test_pelican.py
+++ b/pelican/tests/test_pelican.py
@@ -265,3 +265,21 @@ class TestPelican(LoggedTestCase):
             sys.executable, '-m', 'pelican', '--help'
         ]).decode('ascii', 'replace')
         assert 'usage:' in output
+
+    def test_log_filter(self):
+        settings = read_settings(path=SAMPLE_CONFIG, override={
+            'PATH': INPUT_PATH,
+            'OUTPUT_PATH': self.temp_path,
+            'LOCALE': locale.normalize('en_US'),
+            'LOG_FILTER': [
+                (logging.WARN, 'CATEGORY_SAVE_AS is set to False'),
+                (logging.WARN, 'AUTHOR_SAVE_AS is set to None'),
+            ],
+            'CATEGORY_SAVE_AS': False,
+            'AUTHOR_SAVE_AS': None,
+        })
+        pelican = Pelican(settings=settings)
+        mute(True)(pelican.run)()
+        self.assertLogCountEqual(
+            count=0,  # LOG_FILTER should filter out the log messages
+            level=logging.WARNING)
--- a/pelican/urlwrappers.py
+++ b/pelican/urlwrappers.py
@@ -102,7 +102,7 @@ class URLWrapper(object):
         setting = "%s_%s" % (self.__class__.__name__.upper(), key)
         value = self.settings[setting]
         if not isinstance(value, str):
-            logger.warning('%s is set to %s', setting, value)
+            logger.warning('%s is set to %s' % (setting, value))
             return value
         else:
             if get_page_name:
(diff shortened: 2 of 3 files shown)


justinmayer:
Any comments on this PR, perhaps from @avaris and/or other @getpelican/reviewers?

avaris:
Well, this is the wrong way to do it and it only "fixes" some of the templated filters. Also there seems to be an unrelated change in `writers.py`, for some reason.

I think, the proper way to handle it would be here: https://github.com/getpelican/pelican/blob/cf96c113e1f521a85372f81c0d4d5a6dada941ca/pelican/log.py#L113-L118

with a change something like this:
```python
 # ignore LOG_FILTER records by templates when "debug" isn't enabled 
 logger_level = logging.getLogger().getEffectiveLevel() 
 if logger_level > logging.DEBUG: 
     template_key = (record.levelno, record.msg)
     message_key = (record.levelno, record.getMessage()) 
     if template_key in self._ignore or message_key in self._ignore: 
         return False 
```

justinmayer:
Many thanks for your contribution, Carlo. Would you take a look at the comment from @avaris and make some appropriate adjustments based on that feedback?

justinmayer:
Since the original submitter seems to have disappeared, would anyone be willing to help out by taking the suggestion proposed by @avaris and submitting a new pull request? 🥇

avaris:
Will do.

justinmayer:
Thanks to @avaris, a fix for this is now in master via the above-linked separate pull request.

PR #2684, opened 2020-01-17
Improve UI when `pelican --listen` web server is quit

### LINKED ISSUE #2676: KeyboardInterrupt
If you run:

~~~
pelican --listen
~~~

Pelican does not tell you how to stop. Compare with other projects:

~~~
Press Ctrl+C to stop
~~~

I assume that Pelican accepts the same key, but this is produced:

~~~
  File "/usr/bin/pelican", line 10, in <module>
    sys.exit(main())
  File "/usr/lib/python3.6/site-packages/pelican/__init__.py", line 559, in main
    settings.get("OUTPUT_PATH"))
  File "/usr/lib/python3.6/site-packages/pelican/__init__.py", line 495, in listen
    httpd.serve_forever()
  File "/usr/lib/python3.6/socketserver.py", line 236, in serve_forever
    ready = selector.select(poll_interval)
  File "/usr/lib/python3.6/selectors.py", line 376, in select
    fd_event_list = self._poll.poll(timeout)
KeyboardInterrupt
~~~

justinmayer:
Fixes #2676. Any comments from @getpelican/reviewers?

### DIFF
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -475,6 +475,10 @@ def listen(server, port, output, excqueue=None):
             excqueue.put(traceback.format_exception_only(type(e), e)[-1])
         return
 
+    except KeyboardInterrupt:
+        print("\nKeyboard interrupt received. Shutting down server.")
+        httpd.socket.close()
+
 
 def main(argv=None):
     args = parse_arguments(argv)
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -466,8 +466,9 @@ def listen(server, port, output, excqueue=None):
             excqueue.put(traceback.format_exception_only(type(e), e)[-1])
         return
 
-    logging.info("Serving at port %s, server %s.", port, server)
     try:
+        print("\nServing site at: {}:{} - Tap CTRL-C to stop".format(
+            server, port))
         httpd.serve_forever()
     except Exception as e:
         if excqueue is not None:


avaris:
I don't think `logger.info` is visible without `-v` or `-D`. Maybe `logger.warning` instead?

pelican/__init__.py
```diff
@@ -466,14 +466,22 @@ def listen(server, port, output, excqueue=None):
...
         httpd.serve_forever()
     except Exception as e:
         if excqueue is not None:
             excqueue.put(traceback.format_exception_only(type(e), e)[-1])
         return
 
+    except KeyboardInterrupt:
+        stop_msg = "Keyboard interrupt received. Shutting down server."
+        logger.info(stop_msg)
+        print("\n{}".format(stop_msg))
```
iKevinY: Shouldn't we only have one of either `logger.foo` or `print`, otherwise it'll double-output the message (assuming the logging visibility is high enough)?

justinmayer:
@avaris / @iKevinY: Thanks for the input. I tried replacing the `logger.info()` + `print()` combination with `logger.warning()`, but that yields a rather ugly and inaccurate `WARNING` prefix:
```
WARNING: Serving site at: 127.0.0.1:8000 - Tap CTRL-C to stop
^CWARNING:
  | Keyboard interrupt received. Shutting down server.
```
Whereas before it looked much more user-friendly:
```
Serving site at: 127.0.0.1:8000 - Tap CTRL-C to stop
^C
Keyboard interrupt received. Shutting down server.
```
Any suggestions as to how to reconcile this situation?

iKevinY:
Hmm, agreed that since this is user-facing it would be nicer to have a `print` somewhere at least. If the logger output isn't strictly necessary, could we get by with only a `print` here?

justinmayer:
I modified the commits here to only print to the console, which seems like a workable approach to me. Any further comments on this PR?

avaris:
Yes, `warning` will do that. I considered that as a necessary evil. Alternative would be dropping level to `logging.INFO` when `--listen` is provided but that would result in a really verbose output when combined with `--autoreload`.

I suppose `print` is the simpler solution here.

justinmayer:
Thanks for the input, @avaris and @iKevinY. 💫

PR #2688, opened 2020-01-25
Pythonmeister add xing support

pythonmeister:
This adds support for Xing as a social network.


justinmayer:
Hi Stefan. I appreciate the submission, but I think we are trying to get away from including these kinds of additions. If anything, we are moving toward removing them instead of adding more. So for the moment I think it's best to close this PR. Nonetheless, once again, thank you for the submission.

PR #2689, opened 2020-01-28
Fix bug of open function parameter in pelican_quickstart

xiaojieluo:
Fix a small bug about the encoding parameter of the open function

Signed-off-by: xiaojieluo <xiaojieluoff@gmail.com>


justinmayer:
Hi @xiaojieluo. Many thanks for the fix you submitted. I noticed your contribution just now, after having already merged #2705, which also addresses the problem you encountered. Since that has been merged and the associated problem ostensibly fixed, I think it's best to close this. Thanks again for your contribution! 😅

PR #2705, opened 2020-03-13
Make encoding a named parameter

nikolausschueler:
Without the name, the parameters are used in sequence and the "encoding"
param is used in place of the "buffering" param, which leads to
problems.


justinmayer:
Many thanks to Nikolaus for the fix and to @iKevinY and @avaris for reviewing! ✨

PR #2706, opened 2020-03-14
Handle case of missing port in target "serve-global"

### LINKED ISSUE #2696: Task "serve-global" in Makefile generated by pelican-quickstart requires port, but it is not defined
The task "serve-global" requires that the PORT variable is defined, but it is defined nowhere in the Makefile. 

The problem is in the Makefile template (pelican/tools/templates/Makefile.jinja2):

```
serve-global:
ifdef SERVER
  $(PELICAN) -l $(INPUTDIR) -o $(OUTPUTDIR) -s $(CONFFILE) $(PELICANOPTS) -p $(PORT) -b $(SERVER)
else
  $(PELICAN) -l $(INPUTDIR) -o $(OUTPUTDIR) -s $(CONFFILE) $(PELICANOPTS) -p $(PORT) -b 0.0.0.0
endif
```

The "serve" task checks for port, but "serve-global" only checks for the server. Maybe the solution is a nested check for port (as done in "serve").

nikolausschueler:
Also factor out the check for the SERVER variable to avoid nested
ifdefs.

Fixes #2696

### DIFF
--- a/pelican/tools/templates/Makefile.jinja2
+++ b/pelican/tools/templates/Makefile.jinja2
@@ -50,6 +50,14 @@ ifeq ($(RELATIVE), 1)
 	PELICANOPTS += --relative-urls
 endif
 
+SERVER ?= "0.0.0.0"
+
+PORT ?= 0
+ifneq ($(PORT), 0)
+	PELICANOPTS += -p $(PORT)
+endif
+
+
 help:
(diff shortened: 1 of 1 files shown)


justinmayer:
Thank you for the enhancement, @nikolausschueler!

PR #2707, opened 2020-03-16
Add image tag for cover image of articles.

firvida:
So many blogs may need cover image in their articles. In this commit we add a
way to insert image in the article metadata. Also an example of how to use it
in a theme is added to the simple theme.


jorgesumle:
You should probably add it to the [Writing content section](https://docs.getpelican.com/en/stable/content.html#writing-content) of the documentation.

Anyway, it's trivial to add cover images and you can do that easily, so I don't think it's necessary to add new code for that.

firvida:
I will close this PR as I found a better solution in this repo

https://github.com/peterdesmet/pelican-cover-image

Issue #2712, opened 2020-04-11
Tests are dependent on environment's locale


galaxy4public:
On a clean clone of the repository one could affect the outcome of integrated tests by changing their system locale environment, e.g.:
```
 [galaxy@archlinux pelican]$ LANG=C invoke tests | tail -5
=========================== short test summary info ============================
FAILED pelican/tests/test_pelican.py::TestPelican::test_basic_generation_works
FAILED pelican/tests/test_pelican.py::TestPelican::test_custom_generation_works
FAILED pelican/tests/test_pelican.py::TestPelican::test_custom_locale_generation_works
================== 3 failed, 232 passed, 20 skipped in 5.83s ===================
```
Unfortunately, on my system, there is some other environmental issue that slightly fails these three tests (the content is generated correctly, but class names differ in a minor way, e.g. `class="testingbp"` becomes `class="testingkc"`, I am still trying to figure this out)

Now, if we provide an invalid `LANG` value, suddenly, French tests will start to fail:
```
[galaxy@archlinux pelican]$ LANG=invalid invoke tests | tail -5
FAILED pelican/tests/test_pelican.py::TestPelican::test_custom_locale_generation_works
FAILED pelican/tests/test_settings.py::TestSettingsConfiguration::test_default_encoding
FAILED pelican/tests/test_utils.py::TestDateFormatter::test_french_locale - l...
FAILED pelican/tests/test_utils.py::TestDateFormatter::test_french_strftime
================== 6 failed, 229 passed, 20 skipped in 6.02s ===================
```

They also fail with a valid locale such as `ru_RU` (Russian):
```
[galaxy@archlinux pelican]$ LANG=ru_RU invoke tests | tail -5
FAILED pelican/tests/test_pelican.py::TestPelican::test_custom_locale_generation_works
FAILED pelican/tests/test_settings.py::TestSettingsConfiguration::test_default_encoding
FAILED pelican/tests/test_utils.py::TestDateFormatter::test_french_locale - l...
FAILED pelican/tests/test_utils.py::TestDateFormatter::test_french_strftime
================== 6 failed, 229 passed, 20 skipped in 5.97s ===================
[galaxy@archlinux pelican]$
```
Although, I think it should not really matter for a self-contained test suite, but I have ensured that all locales are installed on my system (despite that I have no use for the French locale):
```
[galaxy@archlinux pelican]$ locale -a
C
en_AU.utf8
en_US.utf8
fr_FR.utf8
POSIX
[...]

PR #2713, opened 2020-04-12
Pin Pygments & Jinja to specific versions

justinmayer:
Tests were failing on developer workstations as functional test output became out-of-sync as new versions of Pygments were released. Tests were passing in automated CI, however, because the version was pinned in CI-specific configuration. This moves that pinned dependency specification to a requirements file in order to (hopefully) ensure that test output and Pygments are kept in sync across environments.

Refs #1916 #1961 #2712


pelican/tests/output/basic/a-markdown-powered-article.html
```diff
@@ -9,7 +9,7 @@
 
 <body id="index" class="home">
         <header id="banner" class="body">
-                <h1><a href="/">A Pelican Blog </a></h1>
```
avaris, line 12: This is a mysterious change.

justinmayer: Agreed. It makes one wonder why the space was there in the first place, no? Perhaps a bug that was fixed?

justinmayer: It seems that Jinja v2.11+ is what changed the handling of this particular space.

justinmayer:
Many thanks for reviewing, @avaris.

PR #1816, opened 2015-09-08
Add JINJA_TESTS to use custom Jinja2 tests. Refs #1498

oseledets:
Sorry if I did anything wrong (first commit), but I tried to do a really simple thing that I use for my website, JINJA_TESTS settings added.

Issue #1498, opened 2014-10-11
Add JINJA_TESTS in the settings


sylvinus:
Should be straightforward to do, just like JINJA_FILTERS

justinmayer:
@sylvinus: Is this something you would like to work on?

leotrs:
Did [800c3c1](https://github.com/oseledets/pelican/commit/800c3c1013fbbc487e2a4adf5e6f360fb2be21a4) ever get merged into pelican? It looks like a good idea.

leotrs:
@oseledets Are you interested in PRing [300c3c1](https://github.com/oseledets/pelican/commit/800c3c1013fbbc487e2a4adf5e6f360fb2be21a4)?

justinmayer:
Pull request #1816 was submitted to address this, but we are still awaiting follow-up changes from @oseledets.

justinmayer:
I implemented this via the above-linked PR, which has been merged to master. 💫

PR #2715, opened 2020-04-12
Add settings for custom Jinja globals and tests

justinmayer:
Pelican has a setting for custom Jinja filters but not custom tests or custom environment namespace globals. This adds both of the latter, including documentation and tests for the new settings.

Closes #1196 #1498 #1816


justinmayer:
Thank you for reviewing, @avaris! 🏅

CNG:
THANK YOU! It works!

PR #2716, opened 2020-04-12
Fix LOG_FILTER to act on exact messages

### LINKED ISSUE #2552: LOG_FILTER not working as it should?

avaris:
* Adds the ability to filter templated messages by exact message,
as well as templates.
* Adds a test for LimitFilter.

Fixes #2552,  #2682

### DIFF
--- a/pelican/log.py
+++ b/pelican/log.py
@@ -110,11 +110,13 @@ class LimitFilter(logging.Filter):
         else:
             self._raised_messages.add(message_key)
 
-        # ignore LOG_FILTER records by templates when "debug" isn't enabled
+        # ignore LOG_FILTER records by templates or messages
+        # when "debug" isn't enabled
         logger_level = logging.getLogger().getEffectiveLevel()
         if logger_level > logging.DEBUG:
-            ignore_key = (record.levelno, record.msg)
-            if ignore_key in self._ignore:
+            template_key = (record.levelno, record.msg)
+            message_key = (record.levelno, record.getMessage())
+            if (template_key in self._ignore or message_key in self._ignore):
                 return False
 
         # check if we went over threshold
--- a/pelican/tests/test_log.py
+++ b/pelican/tests/test_log.py
@@ -1,0 +1,78 @@
+import logging
+import unittest
+from collections import defaultdict
+
+from pelican import log
+from pelican.tests.support import LogCountHandler
+
+
+class TestLog(unittest.TestCase):
+    def setUp(self):
+        super().setUp()
+        self.logger = logging.getLogger(__name__)
+        self.handler = LogCountHandler()
+        self.logger.addHandler(self.handler)
+
+    def tearDown(self):
+        self._reset_limit_filter()
+        self.logger.removeHandler(self.handler)
+        super().tearDown()
+
+    def _reset_limit_filter(self):
+        log.LimitFilter._ignore = set()
+        log.LimitFilter._raised_messages = set()
+        log.LimitFilter._threshold = 5
+        log.LimitFilter._group_count = defaultdict(int)
+
+    def test_log_filter(self):
+        def do_logging():
+            for i in range(5):
+                self.logger.warning('Log %s', i)
+                self.logger.warning('Another log %s', i)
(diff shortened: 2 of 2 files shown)


justinmayer:
Many thanks for this fix, Deniz. Much appreciated! 🏅

jorgesumle:
How could you filter logs when [you can't guess a template or the exact messages are many and too long](https://github.com/getpelican/pelican-plugins/issues/1300)?

avaris:
Why would you guess when you can [look for the template](https://github.com/getpelican/pelican/search?q="original+(not+translated)+items")?

jorgesumle:
> we can change that to be a proper template.

The files affected can be 2, 3, 4, 5..., it's not possible. We could add a regex filter, but I don't think it's worth changing Pelican code (it's just a warning related to a plugin). The plugin cannot be easily reprogrammed, because it works by changing the `SITEURL` for every subsite. I finally gave up and simply commented the lines of code of the warning in my local Pelican version.

avaris:
Why would it be "not possible"? Instead of adding N times `'\n%s'` to the template and passing file names individually, it can pre-format file names (`'\n'.join(file_names)`) and have a single `%s` in the template for that ([like this](https://github.com/getpelican/pelican/blob/7a6686f4679adabdd690668a4ba7cedaf494e243/pelican/server.py#L86-L88)).

And, it does worth changing. You provided the reason :). It makes a feature (`LOG_FILTER`) not feasible.

Now, whether or not this should be automatically filtered in the plugin is another matter. Filtering every warning like this is a bit like "nuclear option". Should be left as a last resort and used only if plugin cannot be refactored to produce warning free output.

Issue #1564, opened 2014-12-18
Add support for installing themes from Python packages


rfleschenberg:
I think it would be cleaner if pelican themes were just plain Python packages that could be installed using setuptools / pip. This would enable users to rely on `pip freeze` / `pip install -r` for repeatable virtualenvs.

I believe it should be possible to do this change while retaining backwards compatibility with existing themes.

FedericoCeratto:
Many users cannot or don't want to install stuff downloaded from GitHub as root.
Maybe the documentation and pelican-themes's manpage should recommend creating local copies as a preferred option (especially given that users might want to tweak the themes).

MinchinWeb:
I think there's some value in being able to install a theme from PyPI; I think it decreases the learning curve when starting with Pelican for the first time, and is also helpful if you want to get a site up quickly.

If anyone else wants to do this: To post my theme to PyPI, I first created a `setup.py` (like any other Python package, as @ingwinlu did), but I also created a small function that returns the theme location on the filesystem. Then in my `pelicanconf.py`, I specify my theme ([Seafoam](https://github.com/MinchinWeb/seafoam)) as follows:

```python
import seafoam

THEME = seafoam.get_path()
```

The `get_path()` function is very simple: it just returns the root directory the file is stored in:

```python
from pathlib import Path

def get_path():
    return(str(Path(__file__).parent))
```

justinmayer:
While something like Peru was floated as an idea in #1655, I think standard Python packaging is the best way forward. Hopefully the community can come together to implement that for the next major Pelican release. :sparkles:

kdeldycke:
I just did as @MinchinWeb proposed and also ended up packaging my theme in a standard Python package: https://pypi.python.org/pypi/plumage . Works great so far!

kdeldycke:
I guess this issue can now be closed: Pelican perfectly support plugins and themes as Python packages.

justinmayer:
Pelican 4.5+ supports installing plugins via Pip, in which those plugins are automatically detected and registered, and they become available immediately without any further action needed by the end user.

Ergo, the scope of this issue has been narrowed to track the addition of similar support for Pip-installable themes, perhaps as [namespace packages](https://packaging.python.org/en/latest/guides/creating-and-discovering-plugins/#using-namespace-packages).

[...]

PR #2719, opened 2020-04-14
Add support for themes to define their own defaults

galaxy4public:
Until #1564 is implemented (and it is a lot of work both on the technical side as well as public relations), I am proposing quite a simple enhancement to allow theme authors to define defaults for things they are using in their themes, e.g. I have configurable layout in the theme and the user could choose to display the navigation menu on the left or the right side, then I could provide the default using `pelicanconf.py` inside my theme folder.

The way this PR is implemented, the theme cannot override any of the system provided setting.  This means that the primary configuration file + command line arguments take precedence in all cases.  This choice was made consciously to avoid any unexpected outfalls from the change.

If the community decides that we actually want to give theme authors the power to mangle system settings, well, then we should discuss how to do it in a correct way (it gets a bit complicated, since , in my opinion, the theme config in this case should be able to override the `DEFAULT_CONFIG` settings, this way user's config and command-line arguments will override it if needed).

### DIFF
--- a/pelican/settings.py
+++ b/pelican/settings.py
@@ -529,16 +529,17 @@ def configure_settings(settings):
         preserved = {}
         logger.debug('Theme provides a config `%s`', theme_config)
         # preserve variables that are dynamically calculated in read_settings()
-        for p in ['PATH', 'OUTPUT_PATH', 'THEME', 'CACHE_PATH', 'PLUGIN_PATHS']:
+        for p in ['PATH', 'OUTPUT_PATH', 'THEME', 'CACHE_PATH',
+                  'PLUGIN_PATHS']:
             if settings.get(p) is not None:
                 preserved[p] = settings.get(p)
-        settings.pop('THEME') # avoid recursion
+        settings.pop('THEME')  # avoid recursion
         settings = read_settings(theme_config, settings)
         # restore them back to look like we did not calculate them again
-        for p in ['PATH', 'OUTPUT_PATH', 'THEME', 'CACHE_PATH', 'PLUGIN_PATHS']:
+        for p in ['PATH', 'OUTPUT_PATH', 'THEME', 'CACHE_PATH',
+                  'PLUGIN_PATHS']:
             if preserved.get(p) is not None:
                 settings[p] = preserved.pop(p)
-
 
     # make paths selected for writing absolute if necessary
     settings['WRITE_SELECTED'] = [
--- a/pelican/settings.py
+++ b/pelican/settings.py
@@ -524,6 +524,22 @@ def configure_settings(settings):
             raise Exception("Could not find the theme %s"
                             % settings['THEME'])
 
+    theme_config = os.path.join(settings['THEME'], 'pelicanconf.py')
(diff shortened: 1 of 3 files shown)


galaxy4public:
The filed test (test_typogrify_ignore_tags) has nothing to do with my changes, so I have no idea on how to fix this :(

avaris:
Yes, it is not your fault. Root of the issue comes from the global state that is required by RST Pygments directive:
https://github.com/getpelican/pelican/blob/e10dcfc5e7a9beff6cce398bb56e78263b021d1e/pelican/settings.py#L204-L209

And parallel test execution surfaced the problem and demonstrated why "global states are bad".

justinmayer:
Yup, my fault. Sorry about my "Let's try running tests in parallel" idea. @avaris said we might be able to work around the global state in our tests, but it could be a long shot. If that's not feasible, we'll remove that option from the `tox.ini` configuration file. Even if that ends up being the only realistic solution, individual developers can still opt into multi-core test runs by setting an environment variable. For two cores: `PYTEST_ADDOPTS=-n 2`

avaris:
I did some more digging and unfortunately the global state is unavoidable because it is actually deep in `docutils`. Looking at [the way `pytest-xdist` works](https://github.com/pytest-dev/pytest-xdist/blob/master/OVERVIEW.md), I think the issue is more about the change in test order. Because it does change it. More precisely, some tests are altering the global state I mentioned above and leaving it like that. This was an issue waiting to blow up :). Any test dealing with `PYGMENT_RST_OPTIONS` needs to clean up afterwards. Calling `pelican.settings.read_settings()` afterwards should be enough, in fact.

Problematic tests seems to be in [`test_pelican`](https://github.com/getpelican/pelican/blob/master/pelican/tests/test_pelican.py#L18-L25) which uses [this](https://github.com/getpelican/pelican/blob/cefe984c7e230c32b2c1b6921c9d55039277ebb0/samples/pelican.conf.py#L53-L54) or [this](https://github.com/getpelican/pelican/blob/cefe984c7e230c32b2c1b6921c9d55039277ebb0/samples/pelican.conf_FR.py#L54-L55) config file. I'm (somewhat) confident that if we put `read_settings()` inside [`tearDown`](https://github.com/getpelican/pelican/blob/cefe984c7e230c32b2c1b6921c9d55039277ebb0/pelican/tests/test_pelican.py#L52-L56) this won't pop up :). I'd say leaving parallel execution on and trying this for a while.

justinmayer:
@avaris addressed the global state in the tests via #2727, which has resolved the above-mentioned test failures.

galaxy4public:
Please do not merge this PR.  I found a better way via the plugin system :).  I think it would be much better to be implemented through a plugin, so people could decide whether they want this functionality or not.  It would be nice to make such a plugin official, though. However, it is a topic for the time I polish the plugin I am working on.

galaxy4public:
Just if anybody stumbles upon this thread, I created the plugin and am quite happy with its functionality: https://github.com/getpelican/pelican-plugins/pull/1256

PR #2725, opened 2020-04-15
Preserve category case in importer

justinmayer:
Adds a `preserve_case` parameter to the `slugify()` function and uses it to preserve capital letters in category names when using the Pelican importer.

Closes #2249


justinmayer:
Thank you for reviewing, Deniz.

PR #2727, opened 2020-04-16
Reset global state set by tests in test_pelican

avaris:
Calling `read_settings` to clear the options put into
`pelican.settings.PYGMENTS_RST_OPTIONS` in `tearDown` so
that tests that are run after are not affected


justinmayer:
Many thanks for this fix, Deniz!

PR #2730, opened 2020-04-17
Fix legacy plugin loading

avaris:
Legacy plugin loading was semi-broken. It would fail if the module imported itself at some point (common for packages). This updates the tests to account for that and fixes the problem.


justinmayer:
Thank for the fix, Deniz! 💯

PR #2731, opened 2020-04-19
Add support for Unicode slugs

avaris:
This has two commits for adding the option to use unicode in auto-generated slugs
* Reworks `slugify` and adds the `use_unicode` argument. See commit message for details about what changed and why. Also adds direct tests for `use_unicode`
* Exposes the option in `settings` as `SLUGIFY_USE_UNICODE` (default: `False`), uses it internally, updates docs and adds an indirect test for it. While updating the docs, I also moved the `SLUGIFY_SOURCE` documentation to where the other related settings are.

Fixes #1207

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -628,6 +628,12 @@ corresponding ``*_URL`` setting as string, while others hard-code them:
    in auto-generated slugs. Otherwise, unicode characters will be replaced
    with ASCII equivalents.
 
+
+.. data:: SLUGIFY_PRESERVE_CASE = False
+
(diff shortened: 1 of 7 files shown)


avaris:
On a slightly related note, `slugify` also has an option for `preserve_case` but this wasn't exposed as a setting. If exposing it sounds beneficial, I can add to this PR.

justinmayer:
@avaris: I think exposing the `preserve_case` setting makes sense. Would you add that to this PR as you suggested?

avaris:
@justinmayer Done :heavy_check_mark:

justinmayer:
Looks great. Many thanks for your work on this! 💯🥇

PR #2735, opened 2020-04-21
Remove duplicate tests in test_contents

avaris:
`TestArticle` was a subclass of `TestPage` and therefore included all
the tests `TestPage` already had. Effectively tests from `TestPage`
were run twice. This splits the common parts of TestPage to a
base class called `TestBase` and `TestPage/TestArticle` will both
inherit from it.


justinmayer:
Many thanks for your work on this! 🎖️

PR #2737, opened 2020-04-23
Docs should lead to existing pelican themes

fridde:
It took me a google search to find out that others already have created a large amount of themes and I was confused that there even is a collection maintained by the Pelican community. Shouldn't that be mentioned in the docs for Pelican?

Also: The difference between "Themes" and "pelican-themes" in the sidebar nav are only obvious *after* reading through their content, but this inconsistency is *not adressed* in this PR. I just put them closer together so that a reader will miss them less likely.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -64,8 +64,8 @@ Documentation
    publish
    settings
    themes
-   plugins
    pelican-themes
+   plugins
    importer
    faq
    tips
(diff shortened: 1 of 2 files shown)


justinmayer:
Thanks for the submission, @fridde. I'm included to take a simpler approach and thus submitted an alternative set of changes in #2933.

justinmayer:
@fridde. Thanks again for pointing out how we could improve things here. Much appreciated!

PR #2738, opened 2020-04-23
Settings via pathlib

MinchinWeb:
The discussion in #2431 suggested that it would be better to deal with the various path settings using Pathlib. This accomplishes that. The earlier discussion suggested that this would effectively drop support for Python 3.5; this does not address that.

It also updates the documentation to show the new defaults.

Any paths with `{slug}` or `{lang}` need to be cast as strings to allow the appropriate replacements.

This would supersedes #2431, and thus closes #2431.

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -454,80 +454,80 @@ This would save your articles into something like
    The place where we will save an article which doesn't use the default
    language.
 
-.. data:: DRAFT_URL = 'drafts/{slug}.html'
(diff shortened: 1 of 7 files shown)


MinchinWeb:
Per #2431, that approach is preferred. This can be closed when that Pull Request is committed.

MinchinWeb:
Closed because the all strings approach in 2431 was selected instead.

PR #2739, opened 2020-04-23
Windows invoke support

MinchinWeb:
This add basic support for Windows to the project's `tasks.py` to support Windows development. In particular:

- it adds *invoke* and *black* to the dev dependencies, which are expected by `tasks.py`
- increases the Python version to 3.6, as *black* doesn't support Python 3.5 (and *poetry* won't install it otherwise...)
- notes that virtualenv binaries are stored in a different path on Windows
- notes that *pip* can't directly update itself on Windows

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,7 @@ classifiers = [
 ]
 
 [tool.poetry.dependencies]
-python = "^3.5"
+python = "^3.6"
 feedgenerator = "^1.9"
 jinja2 = "~2.11"
 pygments = "~2.6.1"
@@ -56,6 +56,8 @@ tox = "^3.13"
 flake8 = "^3.7"
 flake8-import-order = "^0.18.1"
 invoke = "^1.3"
+isort = "^4.3.21"
+black = {version = "^19.10b0", allow-prereleases = true}
 
 [tool.poetry.extras]
 markdown = ["markdown"]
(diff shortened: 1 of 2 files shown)


justinmayer:
Many thanks for the improvements, @MinchinWeb. I took a slightly different approach for the Invoke tasks modifications via 1ac4abcb, but they should have the same effect. Please let me know if there's anything I did there that doesn't work on Windows.

MinchinWeb:
Thanks for tackling this, and for all your support of the project.

In your version, I couldn't find where *black* or *isort* are provided. I think they should be provided either by the *poetry* dependencies (my preference, I think) or in the "tools" list at the top of ‘tasks.py‘. In the same vein, why isn't *pre-commit* included on the *poetry* dependencies?

MinchinWeb:
(P.S. the Windows side looks like it should work, but I'll test it shortly to be sure)

justinmayer:
Sorry for not being more clear... At the same time I made the above-mentioned Invoke-related changes, I merged your commit that adds the Black and `isort` dev dependencies: 13e138a

> Why isn't `pre-commit` included in the Poetry dependencies?

I have been trying to be a bit more conservative when adding these kinds of changes to Pelican core than I have been in, say, a plugin. Plus, the advantage to how it is currently handled in the existing Invoke task is that if Pre-commit is already installed globally, it won't be redundantly installed in the virtual environment when running `poetry install`.

MinchinWeb:
Thanks for the explanation Justin, both make sense. I'd missed that you'd directly cherry-picked that commit.

PR #2742, opened 2020-04-27
Modernize code base to Python 3+ syntax

justinmayer:
This replaces syntax that was relevant in earlier Python versions but that now has modernized equivalents. The changes were generated by running [`pyupgrade --py3-plus`](https://github.com/asottile/pyupgrade) on all Python files in the repository.

In addition, this also relaxes the hard/enforced line-length limit from 79 to 88 characters. Keeping to 79 characters is still preferred, but since sometimes that requires unreasonable contortions and code readability issues, continuous integration checks will no longer fail if, say, they encounter an 80-character line.

### DIFF
--- a/pelican/tests/test_contents.py
+++ b/pelican/tests/test_contents.py
@@ -1,5 +1,3 @@
-# -*- coding: utf-8 -*-
-
 import datetime
 import locale
 import logging
@@ -27,7 +25,7 @@ class TestBase(LoggedTestCase):
     def setUp(self):
         super().setUp()
         self.old_locale = locale.setlocale(locale.LC_ALL)
-        locale.setlocale(locale.LC_ALL, str('C'))
+        locale.setlocale(locale.LC_ALL, 'C')
         self.page_kwargs = {
             'content': TEST_CONTENT,
             'context': {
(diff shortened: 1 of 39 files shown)


justinmayer:
Thanks for having a look, Deniz.

pelican/tests/test_contents.py
```diff
@@ -56,13 +54,13 @@ def _enable_limit_filter(self):
 
     def _copy_page_kwargs(self):
         # make a deep copy of page_kwargs
```
oulenz, line 56: I wonder why we aren't using `copy.deepcopy` here to begin with?

pelican/tests/test_contents.py
```diff
@@ -56,13 +54,13 @@ def _enable_limit_filter(self):
...
-        page_kwargs = dict([(key, self.page_kwargs[key]) for key in
-                            self.page_kwargs])
+        page_kwargs = {key: self.page_kwargs[key] for key in self.page_kwargs}
```
oulenz, line 57: This could just be `page_kwargs = self.page_kwargs.copy()`

pelican/tests/test_contents.py
```diff
@@ -56,13 +54,13 @@ def _enable_limit_filter(self):
...
         for key in page_kwargs:
             if not isinstance(page_kwargs[key], dict):
                 break
```
oulenz, line 60: Should this be `continue`?

avaris:
Yeah, honestly `_copy_page_kwargs` should just be `_get_page_kwargs` and construct the `dict`. `setUp` is run for each test anyways, there should be minimal performance difference, if any.

PR #2746, opened 2020-04-29
Fix error strings whitespace

kernc:


justinmayer:
Thanks to @kernc for the fixes and to @avaris for reviewing. 👍

Issue #2497, opened 2018-12-21
Automatically set Date and Modified fields


Kristinita:
### 1. Question

I don't find, what can I do, that `Date` and `Modified` would be determined automatically.

If it not possible at the time, it would be nice to have this feature.

### 2. Argumentation

I print in my articles and pages:

+ `Date` — date of file creation;
+ `Modified` — date of last file modification.

![Kira__Pelican--ExampleDateAndModified](https://camo.githubusercontent.com/fd78ad55bdbad8a5a3421bee23920e749b2eeabe/68747470733a2f2f692e696d6775722e636f6d2f5945714731674d2e706e67)

[**This is not hard to determine automatically in Python**](https://stackoverflow.com/q/237079/5951529). But in Pelican I need to wrote 2 Metadata extra lines `Date` and `Modified` in every my article/page. I don't need manually write 2 extra lines when I can do it automatically.

### 3. Not helped

1. Pelican documentation, issues, plugins.
1. [**filetime_from_git**](https://github.com/getpelican/pelican-plugins/tree/master/filetime_from_git) can solve this issue, if I deploy a site from local machine. But for remote site deploying [**I have problems, see #813 for details**](https://github.com/travis-ci/dpl/issues/813).

    ![Kira__Pelican_&_filetime_from_git--SameDateAndModified](https://camo.githubusercontent.com/bea25f9ff5aa56beec477ad121c8d4e2428c49ed/68747470733a2f2f692e696d6775722e636f6d2f357261574c61492e706e67)

1. On Stack Overflow I find [**this question**](https://stackoverflow.com/q/20766692/5951529) about similar problem, but I don't understand, how I can accept it for each article/page, not for specific; see [**my comment**](https://stackoverflow.com/questions/20766692/#comment94594599_20766804).

Thanks.

rask004:
From [this](http://docs.getpelican.com/en/stable/settings.html), you can auto-set the creation date with DEFAULT_DATE='fs', which collects a date from the current system timestamp. 

Note that articles without a metadata creation date, if modified, will update this date to the new system timestamp.

I am unsure about automating the modification date at this time.

justinmayer:
[...]

PR #2748, opened 2020-05-03
Setting "modified:" from "date:" when the latter was set from the filesystem

galaxy4public:
Sets the `modified:` metadata from `date:` if the user asked us
to use the filesystem for determining timestamps.

Please see discussion in #2497 .


galaxy4public:
The introduced change is still in line with the [official documentation](https://docs.getpelican.com/en/stable/content.html#file-metadata) that states:
```
modified should be last time you updated the article, and defaults to date if not specified.
```

avaris:
Tests please :).

galaxy4public:
I will try to come up with some in the next 15 minutes or so :)

galaxy4public:
It turned out to be quite difficult since I could not find any tests in the current test suite that do check on the filesystem metadata.  I am thinking about where to put the tests in (right now, I am thinking about working with the default reader, since the change I introduced is not related to any particular reader).

galaxy4public:
OK, now that I introduced 4 additional files, it seems that I need to fix all these unrelated to my change tests which are using the test content directory for working with list of articles :(

avaris:
You could create files in a temp folder to test, and it would've been probably easier that way.

avaris:
e.g.:

```python
def create_content(path, date=None, modified=None):
    with open(path, 'w') as f:
        f.write('<html><head>\n')
        if date is not None:
            f.write('<meta name="date", content="{:%Y-%m-%d %H:%M}" />'.format(date))
        if modified is not None:
            f.write('<meta name="modified", content="{:%Y-%m-%d %H:%M}" />'.format(date))
        f.write('</head></html>\n')

folder = tempfile.mkdtemp()
path = os.path.join(folder, 'content.html')
# test with various combinations of date and modified
create_content(path, date=..., modified=...)
# ...
# cleanup
shutil.rmtree(folder)
```

galaxy4public:
Should I rework the tests, or was it for my future PRs? ;)

avaris:
Well... This is OK as is, even though the number of "super articles" is getting somewhat out of hand. But I wouldn't say "no" if you wanted to refactor these tests :).

galaxy4public:
@avaris , let's agree on me refactoring it a bit later (I promise I will!).  It is just 2am on Monday here and work starts at 9am :).  I only can reliably allocate time during weekends :(.

I think the current state allows this to be merged as is.  I am learning more and more about the codebase, so the quality should step up over the time.

justinmayer:
Many thanks to those who made this possible:
* @Kristinita for reporting
* @galaxy4public for implementing the feature and adding tests
* @avaris for reviewing

Bravo! 👏

PR #2751, opened 2020-05-09
Fix docs/quickstart.rst to use reStructured Text.

wsnyder:
I was following the quickstart but the described page did not show up.

Previously the docs described making a .md, but markdown isn't in the default installation, so using a format which is.  An alternative fix would be to show how to install markdown in the quickstart.

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -38,14 +38,18 @@ Create an article
 You cannot run Pelican until you have created some content. Use your preferred
 text editor to create your first article with the following content::
 
-    Title: My First Review
-    Date: 2010-12-03 10:20
-    Category: Review
-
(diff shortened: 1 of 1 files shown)


justinmayer:
> On May 9, 2020, at 4:37 PM, Wilson Snyder <notifications@github.com> wrote:
> 
> An alternative fix would be to show how to install markdown in the quickstart.

That’s already in the quick-start docs:
   pip install pelican[Markdown]

wsnyder:
Which I didn't install because I didn't intend to use it, not realizing the examples would ;)

Anyhow please feel free to close without merging.

wsnyder:
Actually I'll close without merging, take it if you want & thanks.

PR #2752, opened 2020-05-09
Override settings from the commandline

sabaini:
Add a --setting-overrides KEY=VAL commandline option to override
arbitrary items from pelicanconf.py settings files. This adds
flexibility in running pelican and helps reduce sprawl of settings
files

### DIFF
--- a/pelican/settings.py
+++ b/pelican/settings.py
@@ -1,6 +1,7 @@
 import copy
 import importlib.util
 import inspect
+import json
 import locale
 import logging
 import os
@@ -660,3 +661,23 @@ def configure_settings(settings):
             continue            # setting not specified, nothing to do
 
     return settings
+
+
+def coerce_overrides(overrides):
+    coerced = {}
+    types_to_cast = {int, str}
+    for k, v in overrides.items():
+        if k not in overrides:
+            logger.warning('Override for unknown setting %s, ignoring', k)
+            continue
+        setting_type = type(DEFAULT_CONFIG[k])
+        if setting_type not in types_to_cast:
+            coerced[k] = json.loads(v)
+        else:
+            try:
+                coerced[k] = setting_type(v)
+            except ValueError:
+                logger.debug('ValueError for %s override with %s, try to '
+                             'load as json', k, v)
+                coerced[k] = json.loads(v)
+    return coerced
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 .*.swo
 *.pyc
 .DS_Store
+.idea
 docs/_build
 docs/fr/_build
 build
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -23,7 +23,7 @@ from pelican.plugins import signals
 from pelican.plugins._utils import load_plugins
 from pelican.readers import Readers
 from pelican.server import ComplexHTTPRequestHandler, RootedHTTPServer
-from pelican.settings import read_settings
+from pelican.settings import coerce_overrides, read_settings
 from pelican.utils import (FileSystemWatcher, clean_output_dir, maybe_pluralize)
 from pelican.writers import Writer
 
@@ -230,6 +230,18 @@ class PrintSettings(argparse.Action):
         parser.exit()
 
 
+class ParseDict(argparse.Action):
+    def __call__(self, parser, namespace, values, option_string=None):
+        d = {}
+        if values:
+            for item in values:
+                split_items = item.split("=", 1)
+                key = split_items[0].strip()
+                value = split_items[1].strip()
+                d[key] = value
+        setattr(namespace, self.dest, d)
+
+
 def parse_arguments(argv=None):
     parser = argparse.ArgumentParser(
         description='A tool to generate a static blog, '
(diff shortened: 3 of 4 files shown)


avaris:
I like the idea but this is limited to `string` settings. And I'm pretty sure things will break if you override a setting that expects a non-string, and there are quite a lot of those (i.e. `SUMMARY_MAX_LENGTH`).

Before I throw `ast.literal_eval` out here, I'd like to understand what this is trying to solve. What kind of settings sprawl are we talking about? Usually *two* are enough. One for local development and other for publishing. And you [can reuse one in the other](https://github.com/getpelican/pelican/blob/e6df35330205aa6bdbe56cd40e641fab4958af8c/pelican/tools/templates/publishconf.py.jinja2#L7-L10).

sabaini:
Hey, good point about non-string settings. 

My use case is for maintaining sub-blogs with different configuration, plus the need for a dev/stage/prod area. 

I agree, I don't want to overengineer this. Maybe just put a warning into the help string not to try this with non-strings? Alternatively I think for my use case having a --siteurl cli option would go a long way too.

justinmayer:
Another random thought to throw into the mix… What about supporting environment variables for settings that take string and integer values? For example:

    SITEURL=https://example.com pelican -s settings.py content
    SUMMARY_MAX_LENGTH=75 pelican -s settings.py content

sabaini:
@justinmayer , the env var would be a good addition too (personally I slightly favor cmdline args for being more explicit, but both have their uses).

I've tried to do make this --setting-overrides a bit safer by 

a) cover simple cases by casting values where we have a known int or str type 

b) treating other types as json strings which would then also cover more complex settings

What do you guys think?

pelican/settings.py
```diff
@@ -660,3 +661,16 @@ def configure_settings(settings):
             continue            # setting not specified, nothing to do
 
     return settings
+
+
+def coerce_overrides(overrides):
+    coerced = {}
+    setting_types = {k: type(v) for k, v in DEFAULT_CONFIG.items()}
```
avaris, line 668: You don't need to construct this. You can look up as you go through overrides.

pelican/settings.py
```diff
@@ -660,3 +661,16 @@ def configure_settings(settings):
...
+    types_to_cast = {int, str}
+    for k, v in overrides.items():
+        setting_type = setting_types[k]
```
avaris, line 671: You'll get an unclear error here if an unknown setting is passed. Better to check and provide a more helpful error message. Either log a warning and ignore or raise error and stop. I'm partial to both approaches.

pelican/settings.py
```diff
@@ -660,3 +661,16 @@ def configure_settings(settings):
             continue            # setting not specified, nothing to do
 
     return settings
+
+
+def coerce_overrides(overrides):
+    coerced = {}
+    setting_types = {k: type(v) for k, v in DEFAULT_CONFIG.items()}
+    types_to_cast = {int, str}
```
avaris, line 669: This is slightly tricky. Some settings allow `None` as well as a value (i.e. `SUMMARY_MAX_LENGTH` again... or `FEED_*` options). I don't have a clear solution for those.

pelican/settings.py
sabaini: Nixed that set and queried directly

pelican/settings.py
sabaini: I've now added a logging.warning()

pelican/settings.py
```diff
@@ -660,3 +661,18 @@ def configure_settings(settings):
...
 
     return settings
+
+
+def coerce_overrides(overrides):
+    coerced = {}
+    types_to_cast = {int, str}
+    for k, v in overrides.items():
+        if k not in overrides:
+            logger.warning("Override for unknown setting {}, ignoring".format(k))
```
avaris, line 671: [`%s` style](https://docs.getpelican.com/en/stable/contribute.html#logging-tips) please :).

pelican/settings.py
sabaini: Instead of auto-coercing str/int I could make all overrides use JSON notation. I wanted to avoid this because you'd need to extra quote strings then
` overridesetting='"some-val"' `

Which seems a bit verbosey

pelican/settings.py
sabaini: oops, updated

pelican/settings.py
```diff
@@ -660,3 +661,18 @@ def configure_settings(settings):
...
 
     return settings
+
+
+def coerce_overrides(overrides):
+    coerced = {}
+    types_to_cast = {int, str}
+    for k, v in overrides.items():
+        if k not in overrides:
+            logger.warning('Override for unknown setting %s, ignoring' % k)
```
avaris, line 671: Close but not exactly :). That's still formatting it beforehand. Note **`,`** instead of **`%`**
 
```python
logger.warning('Override for unknown setting %s, ignoring', k)
```

pelican/settings.py
avaris: The problem is you can't use override to set these variables as `None` (for JSON it means `SETTING=null` I guess). We can consider that as a limitation I suppose and I won't be too bothered with it. 

For the `int` defaults, you could `try` to `int()` it and fallback to JSON if it errors, but string ones are hopeless since `'null'` is valid.

sabaini: Ok, I've added a second pass when getting a ValueError on int() casts

pelican/settings.py
sabaini: d'oh

sabaini:
Thanks, I've just squashed changes and pushed, the Linter gave his nod too now ;-)

sabaini:
Hey @avaris any chance this PR could be merged? Cheers

justinmayer:
Sorry for the delay, Peter. I will do my best to review this as soon as possible.

justinmayer:
Sorry for the delay, Peter. Two requested changes...

1. Remove the `.idea` addition to `.gitignore` — these kinds of vendor-specific files are better ignored on your individual workstation via the global `~/.gitignore` file.
2. Change the short option flag from `-c` from `-e` — more consistent with other tools' options for overriding variables (e.g., Ansible, etc.)

justinmayer:
Merged via 1c50a18d0a45d76f6b50647536faf25954740d20. I changed the option flag syntax as per my previous comment via d728cd2696ecaef0f1070faf3c12ce1fd2819c6f and then added documentation via 04c0ea53e6bc79a1447e1ead246037103f39aca9.

@sabaini: Many thanks, Peter, for implementing this feature, which will be a worthy addition to the next Pelican release. If you ever make it down south to Trentino, beers are on me! 🍻

PR #2753, opened 2020-05-09
Fix dictionary logging in formatter

avaris:
Python special cases single Mapping arguments to logging. This
adjusts BaseFormatter to skip "fancy" formatting if argument
is of type Mapping. Also adds various tests for formatted log outputs.

Fixes #2596


justinmayer:
Another well-conceived enhancement. Kudos and much gratitude! 🙏🏻

PR #2758, opened 2020-05-22
Settings support for pathlib

MinchinWeb:
Further to the work I did in PR #2738 (moving all settings to `pathlib.Path`).

It was decided that it was better to keep the default settings as strings, but this is the code that allows Path's to provided for any of these settings.

Perhaps it makes sense to add this support (since the work is done already), so I've extracted the support code.

- [ ? ] Added **tests** for changed code
- [ ? ] Updated **documentation** for changed code

### DIFF
--- a/pelican/contents.py
+++ b/pelican/contents.py
@@ -215,7 +215,7 @@ class Content:
         if not klass:
             klass = self.__class__.__name__
         fq_key = ('{}_{}'.format(klass, key)).upper()
-        return self.settings[fq_key].format(**self.url_format)
+        return str(self.settings[fq_key]).format(**self.url_format)
 
     def get_url_setting(self, key):
         if hasattr(self, 'override_' + key):
--- a/pelican/generators.py
+++ b/pelican/generators.py
@@ -18,7 +18,6 @@ from pelican.readers import Readers
 from pelican.utils import (DateFormatter, copy, mkdir_p, order_content,
                            posixize_path, process_translations)
 
-
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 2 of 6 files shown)


avaris (approved):

justinmayer (approved):
Thanks to @MinchinWeb for the enhancement (and the patience) and also to @avaris for reviewing. 👏

MinchinWeb:
Delighted to see it merged (and useful) after all this time!

adnia:
I believe this change breaks the [Dynamic Pelican Settings](https://www.stefaanlippens.net/dynamic-pelican-settings.html) workaround that I also relied on. I also had the use case of preserving old link structures from before switching to Pelican to the best extent possible.

Especially [this change](https://github.com/getpelican/pelican/pull/2758/commits/cfba3d72beef60de10311a75f83ebb259269fed8#diff-c2d6a33a7c119350a999ffb11f5fa6892979f3a0291a54d4c73eaca17de1b9c2L218) seems to be the culprit as it now returns `str(<class>).format(...)`, i.e, dynamically overwriting the `format()` function is no longer possible.

Indeed, changing the logic to `str(<class>.format(...))` resolves the issue for me. Is there a better, intended solution instead of using this workaround?

avaris:
@adnia from a brief glance, for that to work you also need to define what `str(...)` returns, I believe this would work.

```python
class DynamicSetting(object):
    def __init__(self, f):
        self.f = f
    def format(self, **metadata):
        return self.f(metadata).format(**metadata)
    def __str__(self):
        return self
```

But, I have to emphasize that this method of monkey-patching, although neat, still relies on certain structure for things that are "internal" to `pelican`. So, I cannot guarantee that things won't change in the future :).

MinchinWeb:
> Indeed, changing the logic to str(<class>.format(...))

This doesn't work, as the logic needs to work on both `Pathlib.path` and `str`. In particular, `.format()` is a function of the `str` class, so you need to have a string before you can format it.

The plugin idea is probably a good way to go; if you get stuck, ask here, I've written a number of plugins over the years. :)

adnia:
Just as a quick, yet belated update: I finally came around adjusting my Pelican setup. I noticed that I had unsuccessfully tried to incorporate the handling of this corner case in my larger plugin. However, with the the pointers provided by @avaris, I was able to adjust my code and succeed, ditching the workaround. :)
So, thank you everyone for the support. 👍

PR #2761, opened 2020-05-28
Introduce paragraph count summary

gagath:
Let the users use only the n-first paragraphs of the article as article
summary.

The advantage of this approach is that we avoid the random word count
ellipsis that will cut content in pieces, while not having to copy the
first paragraph of the article into the article's summary metadata.

# Pull Request Checklist


pelican/tests/test_contents.py
```diff
@@ -113,6 +113,18 @@ def test_summary_max_length(self):
         page = Page(**page_kwargs)
         self.assertEqual(page.summary, '')
 
+    def test_summary_paragraph(self):
+        # If a :SUMMARY_MAX_LENGTH: is set, there is no other summary and the
+        # value is negative, generated summary should not exeed the given
```
boxydog, line 118: typo: exeed => exceed

boxydog:
The functionality (paragraphs) seems reasonable. Using negative and positive SUMMARY_MAX_LENGTH feels confusing. I'd introduce a different config var. But, my opinion is not binding.

boxydog:
Less confusing would be SUMMARY_MAX_LENGTH (for words) and SUMMARY_MAX_PARAGRAPHS (for paragraphs).

If both are set, could 1) apply both, or 2) pick a winner (e.g., SUMMARY_MAX_PARAGRAPHS), or 3) raise an error, or 4) hidden option 4.

paulocoutinhox:
+1

pelican/utils.py
```diff
@@ -603,6 +603,25 @@ def truncate_html_words(s, num, end_text='…'):
...
+
+    Newlines in the HTML are preserved.
+    """
+    paragraphs = []
+    tag_stop = 0
+    substr = s[:]
+    for i in range(count):
+        substr = substr[tag_stop:]
+        tag_start = substr.find('<p>')
+        tag_stop = substr.find('</p>') + len('</p>')
```
avaris, line 619: My time on the interwebs taught me that the possibility of finding a closing tag for `<p>` is like a coin toss. Even the HTML spec is permissive about it and "formally" allows omitting `</p>` in most cases. Long story short, I am trying to say that this is quite unreliable.

gagath: The HTML is mostly generated from a generator (reST, Markdown…) that hopefully emits closing tags, so in practice this should not happen very much.

avaris: > in practice this should not happen very much.

That doesn't cut it though :). `rst` and `md` generators may play nice (for now) but content can come from any number for sources not limited to those, including direct `html` files, plugins for other formats etc. In software, edge cases happen more common than one would think initially :).

I'm not suggesting employing full DOM parsers like `lxml` or similar, although that would be more complete. But that can be postponed until the need arises. It should at least produce something reasonable in those cases: i.e. if `</p>` is not present (`substr.find('</p>') == -1`), take up to the next `<p>`?

docs/settings.rst
```diff
@@ -308,6 +308,14 @@ Basic settings
    does not otherwise specify a summary. Setting to ``None`` will cause the
    summary to be a copy of the original content.
 
+.. data:: SUMMARY_MAX_PARAGRAPHS = 1
```
avaris, line 311: Why `= 1` part? You don't define a default for it (in `settings.py`).

pelican/contents.py
```diff
@@ -431,8 +432,17 @@ def get_summary(self, siteurl):
         if "summary" in self.metadata:
             return self.metadata["summary"]
 
+        content = self.content
+        if (
+            "SUMMARY_MAX_PARAGRAPHS" in self.settings
+            and self.settings["SUMMARY_MAX_PARAGRAPHS"] is not None
+        ):
```
avaris, line 439: Slightly simplified version of this:
```python
if self.settings.get("SUMMARY_MAX_PARAGRAPHS") is not None:
```

gagath:
@avaris Thanks for your comments, fixed them.

justinmayer (approved):
Many thanks to @gagath for the enhancement as well as for your patience.

Thanks also to @boxydog and @avaris for reviewing!

PR #2766, opened 2020-06-07
Fix port 0 & remove duplicate port arguments in Makefile

davidjb:
The previous Makefile was forcing the `-p` port argument as `0`, given the checks for PORT via `ifdef`; this fixes that by using the PORT variable check earlier in the Makefile sets up the `-p` argument as part of PELICANOPTS correctly. 

This also fixes the situation where PORT was specified and `-p` was being duplicated as it was part of PELICANOPTS and then additionally added within each of the serve targets.

# Pull Request Checklist


justinmayer:
Nice catch, David. Many thanks! 👏

PR #2785, opened 2020-07-30
Lowercase metadata field name when comparing with FORMATTED_FIELDS in rST reader

jwodder:
# Pull Request Checklist

- Updated **documentation** for changed code — Not necessary

Currently, when Pelican processes metadata fields in a reStructuredText document, it lowercases the field name *after* processing each field's value.  This means that, if you write your summary as `:Summary: Text goes *here*` and leave `FORMATTED_FIELDS` at its default value, the summary will not be formatted.  This is in contrast to the Markdown reader, which lowercases field names before processing values, and so `Summary: Text goes *here*` does get formatted in Markdown.

This PR changes the rST reader to lowercase the field name when comparing it against the elements of `FORMATTED_FIELDS`.


justinmayer:
Thank you for the submission, John.

@avaris: What do you think? Do you foresee any problems with this change?

justinmayer:
Many thanks to @jwodder for the enhancement and to @avaris for reviewing. ✨

PR #2788, opened 2020-08-11
Makefile: Include tags directory in rsync command

elboulangero:
By default, the rsync option '--cvs-exclude' excludes the 'tags'
directory. For a blog, it's a bit unfortunate, as it's quite common to
have a `tags` directory in a blog, either for the tag pages or the tag
feeds.

With this commit, we force rsync to include this directory, and save a
little headache to users who wonder why their tags are present in the
output directory, but are not present on the server.

# Pull Request Checklist

Resolves: #issue-number-here 


justinmayer:
Thank you for the enhancement, Arnaud! 💫

Issue #2805, opened 2020-10-09
Document breaking changes in Pelican 4.5


shniubobo:
- [x] I have read the [Filing Issues](https://docs.getpelican.com/en/latest/contribute.html#filing-issues) and subsequent “How to Get Help” sections of the documentation.
- [x] I have searched the [issues](https://github.com/getpelican/pelican/issues?q=is%3Aissue) (including closed ones) and believe that this is not a duplicate.

- **OS version and name**: Ubuntu 18.04 (WSL)
- **Python version**: 3.8.6

# Issues

## The first one

**As of https://github.com/getpelican/pelican/commit/58edad6897931f21c7c2f4314cfab33c2b167680, `import pelican.signals` is no longer valid.**

```
$ git reset 58edad^ --hard
$ python -c "import pelican.signals"  # exits 0

$ git reset 58edad --hard
$ python -c "import pelican.signals"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'pelican.signals'
```

## The second one

**As of https://github.com/getpelican/pelican/commit/ed1eca160e80bebbfd5870afe1f2a111f388e0ed, legacy plugins with `.` in their names stop working.**

### To reproduce

Create these files:
```
├── content
│   └── 1.rst
├── pelicanconf.py
├── plugins
│   └── m
│       └── test.py
```

```rst
.. content/1.rst
###
123
###
```

```python
# pelicanconf.py
PLUGINS = [
    'm.test',
]
PLUGIN_PATHS = [
    'plugins'
]
```

```python
# plugins/m/test.py
from pelican import signals


def takes_nothing():
    pass


def register():
    signals.initialized.connect(takes_nothing)
```

And then:
```
$ git reset ed1eca^ --hard
$ python -m pelican content
WARNING: Feeds generated without SITEURL set properly may not be valid
WARNING: No timezone information specified in the settings. Assuming your timezone is UTC for feed generation. Check http://docs.getpelican.com/en/latest/settings.html#timezone for more information
CRITICAL: takes_nothing() takes 0 positional arguments but 1 was given
# plugin correctly loaded

$ git reset ed1eca --hard
$ python -m pelican content
WARNING: Feeds generated without SITEURL set properly may not be valid
WARNING: No timezone information specified in the settings. Assuming your timezone is UTC for feed generation. Check http://docs.getpelican.com/en/latest/settings.html#timezone for more information
ERROR: Cannot load plugin `m.test`
[...]

PR #2813, opened 2020-10-19
Fix legacy plugin loading for packages

avaris:
Resolves: #2805 

* Resolves packages for legacy plugins (i.e. `"foo.bar"`) and adds tests for it
* Restores `pelican.signals` but it will raise an `ImportError`. So, plugins using `import pelican.signals` directly will see this:
    > Importing from `pelican.signals` is deprecated. Use `from pelican import signals` or `import pelican.plugins.signals` instead.

<hr>

~Updated **documentation** for changed code~


justinmayer:
Really appreciate you taking the time to work on solutions for these issues. You rock! 🥇

PR #2821, opened 2020-11-21
Improve logging of generators and writer loaders

lgiordani:
This improves the logging of `get_generator_classes()` to provide the source of the generator (either internal or from a plugin). It also reviews the code to make it clearer (in my opinion, but I'm biased :wink:), renames `get_writer()` to `_get_writer_class()` for consistency with the renamed `_get_generator_classes()`. Both are purely internal methods so I think it makes sense to signal it with the initial underscore.

Aside from cosmetic changes this doesn't affect the Pelican loading mechanism at all.

This is my first PR for this project, so please help me double checking that everything was done according to the guidelines.

# Pull Request Checklist

### DIFF
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -114,7 +114,7 @@ class Pelican:
 
         signals.all_generators_finalized.send(generators)
 
-        writer = self._get_writer_class()
+        writer = self._get_writer()
 
         for p in generators:
             if hasattr(p, 'generate_output'):
@@ -164,47 +164,42 @@ class Pelican:
 
     def _get_generator_classes(self):
         discovered_generators = [
-            (ArticlesGenerator, None),
-            (PagesGenerator, None)
+            (ArticlesGenerator, "internal"),
+            (PagesGenerator, "internal")
         ]
 
         if self.settings["TEMPLATE_PAGES"]:
-            discovered_generators.append((TemplatePagesGenerator, None))
+            discovered_generators.append((TemplatePagesGenerator, "internal"))
 
         if self.settings["OUTPUT_SOURCES"]:
-            discovered_generators.append((SourceFileGenerator, None))
+            discovered_generators.append((SourceFileGenerator, "internal"))
 
         for receiver, values in signals.get_generators.send(self):
             if not isinstance(values, Iterable):
                 values = (values,)
 
                 discovered_generators.extend(
-                    [(generator, receiver) for generator in values]
+                    [(generator, receiver.__module__) for generator in values]
                 )
 
         # StaticGenerator must run last, so it can identify files that
         # were skipped by the other generators, and so static files can
         # have their output paths overridden by the {attach} link syntax.
-        discovered_generators.append((StaticGenerator, None))
+        discovered_generators.append((StaticGenerator, "internal"))
 
         generators = []
 
-        for generator, receiver in discovered_generators:
-            if receiver is None:
-                origin = "internal"
-            else:
-                origin = receiver.__module__
-
+        for generator, origin in discovered_generators:
             if not isinstance(generator, type):
-                logger.error("Generator %s (%s) cannot be loaded" % (generator, origin))
+                logger.error("Generator %s (%s) cannot be loaded", generator, origin)
                 continue
 
-            logger.debug("Found generator: %s (%s)" % (generator.__name__, origin))
+            logger.debug("Found generator: %s (%s)", generator.__name__, origin)
             generators.append(generator)
 
         return generators
 
-    def _get_writer_class(self):
+    def _get_writer(self):
         writers = [w for _, w in signals.get_writer.send(self) if isinstance(w, type)]
         num_writers = len(writers)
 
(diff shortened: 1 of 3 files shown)


lgiordani:
I didn't realise tests were running on my forked branch as well :man_facepalming: Good, first mistake done. I'll fix that and come back, sorry!

avaris:
I'll look in more detail later but wanted to mention the [logging scheme we use](https://docs.getpelican.com/en/latest/contribute.html#logging-tips): Use `%s` placeholders instead and pass values as arguments. That allows processing of arguments when necessary as well as ability to [filter](https://docs.getpelican.com/en/latest/settings.html#logging) them on the user side.

lgiordani:
Sorry, I should have noticed that part on the contribution document you mentioned. I will change the PR accordingly

pelican/__init__.py
```diff
@@ -162,46 +162,62 @@ def run(self):
...
+            if receiver is None:
+                origin = "internal"
+            else:
+                origin = receiver.__module__
+
+            if not isinstance(generator, type):
+                logger.error("Generator %s (%s) cannot be loaded" % (generator, origin))
+                continue
+
+            logger.debug("Found generator: %s (%s)" % (generator.__name__, origin))
```
avaris, line 202: Still formatting beforehand :). This will construct a string and *then* pass that to logging. What I meant was, pass the message template (with placeholders) and values to be replaced as arguments instead:

```python
logger.debug("Found generator: %s (%s)", generator.__name__, origin)
```

avaris (changes requested):

pelican/__init__.py
```diff
@@ -162,46 +162,62 @@ def run(self):
...
-        generators.append(StaticGenerator)
+        discovered_generators.append((StaticGenerator, None))
+
+        generators = []
+
+        for generator, receiver in discovered_generators:
+            if receiver is None:
+                origin = "internal"
+            else:
+                origin = receiver.__module__
```
avaris, line 196: Why don't you do this while you're accumulating the generators? Just add `"internal"` instead of `None` and use `receiver.__module__` instead of `receiver above.

pelican/__init__.py
```diff
@@ -162,46 +162,62 @@ def run(self):
...
+            generators.append(generator)
+
         return generators
 
-    def get_writer(self):
-        writers = [w for (_, w) in signals.get_writer.send(self)
-                   if isinstance(w, type)]
-        writers_found = len(writers)
-        if writers_found == 0:
+    def _get_writer_class(self):
```
avaris, line 207: Well, technically this is returning a writer *instance*, not class.

pelican/__init__.py
lgiordani: Now, sometime I truly feel like an idiot :facepalm:

pelican/__init__.py
lgiordani: :facepalm: :facepalm: :facepalm:

pelican/__init__.py
lgiordani: OK I hope they are all fixed now

pelican/__init__.py
lgiordani: Done

pelican/__init__.py
lgiordani: I reverted the name change, it was clearly a mistake :joy_cat:

avaris (approved):
Looks good, thanks :+1:

justinmayer:
Many thanks to @lgiordani for the enhancements and to @avaris for reviewing. ✨

justinmayer:
Thanks to [AutoPub](https://justinmayer.com/projects/autopub/), these changes have been released as [Pelican v4.5.2](https://github.com/getpelican/pelican/releases/tag/4.5.2). 🎉

lgiordani:
Thanks both! I learned more than I expected :tada:

pelican/__init__.py
```diff
@@ -162,46 +162,57 @@ def run(self):
...
+            if not isinstance(values, Iterable):
+                values = (values,)
 
-            for v in value:
-                if isinstance(v, type):
-                    logger.debug('Found generator: %s', v)
-                    generators.append(v)
+                discovered_generators.extend(
+                    [(generator, receiver.__module__) for generator in values]
+                )
```
clokep, line 183: Is this meant to be inside of the `if not isinstance(...)` check? If `values` is already an iterable from `get_generators.send()` call then this code will never run (and they won't be added to `discovered_generators`).

avaris: Oh, you are right. I missed the indent level in the diff 🤦 .

clokep: That was my guess! 😢 I think the old behavior just needs to de-dent these three lines!

lgiordani: Oh, snap, sorry guys, such a stupid mistake. I'll send a fix asap

lgiordani: See #2823 Thanks @clokep for pointing it out

PR #2823, opened 2020-11-23
Fix code indented incorrectly in generators logging

lgiordani:
# Pull Request Checklist

This fixes a mistake merged with #2821. The code was incorrectly intended and thus is not executed in all the required cases.


justinmayer:
Many thanks to @lgiordani for the fix and to @clokep for reviewing. ✨

PR #2826, opened 2020-12-09
Add "http://" prefix to `pelican --listen` output

someone-somenet-org:
Normally my terminal highlights URLs and i can not only easily spot them, but also make the open in a browser with one click.
This obviously cannot happen, if there is no url to be found.

This change adds ``http://`` before the ip and port in the log output so terminals that support URL parsing can pick this up.

### DIFF
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -469,7 +469,7 @@ def listen(server, port, output, excqueue=None):
         return
 
     try:
-        print("\nServing site at: {}:{} - Tap CTRL-C to stop".format(
+        print("\nServing site at: http://{}:{} - Tap CTRL-C to stop".format(
             server, port))
         httpd.serve_forever()
     except Exception as e:


justinmayer:
Normally I'd prefer for this to be scheme-agnostic, in case some folks are serving locally via HTTPS, but I can't think of an easy way to determine that automatically. In which case, I imagine this would be fine since there are probably very few folks who are previewing via https://localhost.

Any thoughts on this, @getpelican/reviewers ?

kylef:
Does this particular code allow https?

Looking at this particular code path, I don't see anywhere that ssl would be enabled. The only listening configuration would be the host and port https://github.com/getpelican/pelican/blob/049bb2e1b3d06721023a12be8e0885b96fed9a0e/pelican/__init__.py#L339-L347

I did notice that `server.py` https://github.com/getpelican/pelican/blob/049bb2e1b3d06721023a12be8e0885b96fed9a0e/pelican/server.py#L121-L124 to use it, but that doesn't appear to apply to this code path. If it does, we could check for `--ssl` and then use `https` scheme.

someone-somenet-org:
As I understand it, the code that was able to serve https was deprecated and the "new way" doesn't support ssl.
Therefore there is no reason to be "scheme agnostic" as the only protocol served is http.

If that should ever change, it would not be too difficult to also change the scheme in the log message too.
I am actually not really sure, if its a good idea and an expected use-case to have a static site generator serve dynamically generated pages in a non-development environment, in which case https support is not needed anyway.

iKevinY:
> Normally I'd prefer for this to be scheme-agnostic [...] I imagine this would be fine since there are probably very few folks who are previewing via https://localhost.

I agree with this sentiment; I suspect the benefit of just adding in an `http://` to help terminals auto-link the URL outweighs the downside of displaying `http` to folks who are _actually_ serving via `https`.

someone-somenet-org:
> > Normally I'd prefer for this to be scheme-agnostic [...] I imagine this would be fine since there are probably very few folks who are previewing via https://localhost.
> 
> I agree with this sentiment; I suspect the benefit of just adding in an `http://` to help terminals auto-link the URL outweighs the downside of displaying `http` to folks who are _actually_ serving via `https`.

There doesnt even seem to be https support in that code path.

iKevinY (approved):
Seems like a nice QoL change, thanks!

justinmayer:
Many thanks to @someone-somenet-org for the contribution and to @kylef and @iKevinY for reviewing. ✨ 

@kylef: Long time no see. Many thanks for chiming in! 😁

PR #2831, opened 2020-12-29
Add all-numeric period variable for the period_archives template

encukou:
Hello!
I would like to use the language-agnostic date format `2020-12-29` in my Period Archive, but the template only gets [a localized name for the month](https://docs.getpelican.com/en/stable/themes.html?highlight=period_archives#period-archives-html).
Would it be possible to provide an all-numeric `period` as well?

Here is an idea of how it could work.

# Pull Request Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+Release type: minor
+
+Add a `period_num` variable for the `period_archives` template.
--- a/docs/themes.rst
+++ b/docs/themes.rst
@@ -327,6 +327,8 @@ period                  A tuple of the form (`year`, `month`, `day`) that
                         given year. It contains both `year` and `month`
                         if the time period is over years and months and
                         so on.
+period_num              A tuple of the form (`year`, `month`, `day`), as in
+                        `period`, except all values are numbers.
 
 ===================     ===================================================
 
--- a/pelican/generators.py
+++ b/pelican/generators.py
@@ -511,6 +511,7 @@ class ArticlesGenerator(CachingGenerator):
 
                 if key == period_date_key['year']:
                     context["period"] = (_period,)
+                    context["period_num"] = (_period,)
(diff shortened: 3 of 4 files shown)


oulenz (approved):
Makes sense! I've tried to think of other ways to tackle this but can't come up with anything that's better.

MinchinWeb:
It was to solve this problem I wrote the *datetime_from_period* filter (included in the [Jinja Filters](https://github.com/pelican-plugins/jinja-filters) plugin). It allows you to do things like this:

```html+jinja
{% block title -%}
    {%- if period | length == 1 -%}
        {{ period | datetime_from_period | strftime("%Y") }}
    {%- elif period | length == 2 -%}
        {{ period | datetime_from_period | strftime("%B %Y") }}
    {%- else -%}
        {{ period | datetime_from_period | strftime("%B %-d, %Y") }}
    {%- endif %} Archives &mdash; {{ super.super() }}
{%- endblock %}
```

And get out (for the daily archive page): `May 3, 2021 Archives &mdash; SITE_TITLE`

You could adjust the `strftime` format string to be whatever you want.

You might also get the format you want like this:

```html+jinja
{{ (period | datetime_from_period).year }}-{{ (period | datetime_from_period).month }}-{{ (period | datetime_from_period).day }}
```

(The code samples are from my seafoam theme --> https://github.com/MinchinWeb/seafoam/blob/archive-periods/seafoam/templates/period_archives.html )

Hopefully this helps!

avaris (approved):

justinmayer:
Manually merged via 487da3550bffb96ebfed9dbf206d8b3574886a21. Will issue a release soon containing this new variable. Many thanks to @encukou for the enhancement and to @oulenz, @avaris, and @MinchinWeb for reviewing. 🥇

Issue #2828, opened 2020-12-17
"can't pickle module objects" - Referencing a plugin as a module object breaks caching


m90:
- [x] I have read the [Filing Issues](https://docs.getpelican.com/en/latest/contribute.html#filing-issues) and subsequent “How to Get Help” sections of the documentation.
- [x] I have searched the [issues](https://github.com/getpelican/pelican/issues?q=is%3Aissue) (including closed ones) and believe that this is not a duplicate.

- **OS version and name**: `python:3.7` Docker image
- **Python version**: 3.7
- **Pelican version**: 4.5.3
- **Links to plugins**: https://github.com/offen/pelican-decorate-content
- **Link to your source**: https://github.com/offen/website

## Issue

When importing a Pelican plugin as a Python module and passing its reference to the `PLUGINS` setting like this

```py
from pelican_decorate_content import decorate_content

PLUGIN_PATHS = ['./plugins']
PLUGINS = [decorate_content, 'assets']
```

and using `CACHE_CONTENT`

```py
CACHE_CONTENT = True
```

Pelican fails building my site with the following error:

```
Traceback (most recent call last):
  File "/root/.local/bin/pelican", line 8, in <module>
    sys.exit(main())
  File "/root/.local/lib/python3.7/site-packages/pelican/__init__.py", line 516, in main
    autoreload(args)
  File "/root/.local/lib/python3.7/site-packages/pelican/__init__.py", line 440, in autoreload
    pelican.run()
  File "/root/.local/lib/python3.7/site-packages/pelican/__init__.py", line 109, in run
    p.generate_context()
  File "/root/.local/lib/python3.7/site-packages/pelican/generators.py", line 681, in generate_context
    self.readers.save_cache()
  File "/root/.local/lib/python3.7/site-packages/pelican/cache.py", line 67, in save_cache
    pickle.dump(self._cache, fhandle)
TypeError: can't pickle module objects
make: *** [Makefile:47: regenerate] Error 1
```

which I would assume is due to the fact that caching tries to pickle the settings file, but fails to pickle the imported plugin. If I either __remove the plugin or disable caching__, building the site __works as expected__.

Is this:
- a problem with the plugin itself?
- wrong configuration on my end?
- a bug in pelican that got introduced when allowing to import plugins like this?

avaris:
Well, neither. Placing imported plugins directly into `PLUGINS` was supported way before cache was introduced. Looks more like an oversight in the caching.

m90:
[...]

Issue #2797, opened 2020-08-27
Enable themes to discover loaded plugins


saagit:
- [x] I have searched the [issues](https://github.com/getpelican/pelican/issues?q=is%3Aissue) (including closed ones) and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://docs.getpelican.com/) and believe that my question is not covered.
- [x] I am willing to lend a hand to help implement this feature. 

## Feature Request

Historically, before 4.5.0, a few themes have examined the PLUGINS variable to determine if a particular plugin is loaded so they can output code as appropriate.  One example is [here](https://github.com/getpelican/pelican-themes/blob/8f2a03c1a3afd22ad418a70f1406cd43382be953/pelican-bootstrap3/templates/base.html#L70).  This mechanism was not intended as a stable API by the developers. When PLUGINS is not set in pelicanconf.py, 4.5.0 defaults PLUGINS to None now instead of [].  This results in the exception `TypeError: argument of type 'NoneType' is not iterable`.

It would be great if there was a stable API to allow themes to discover what plugins are loaded.  In IRC, Avaris  expressed a preference for something like `{% if plugin_enabled('assets') %}`.  If backwards compatibility is deemed important enough, PLUGINS could be also be set to a list of loaded plugins.

saagit:
As I was looking for something else, in the [docs](https://docs.getpelican.com/en/stable/themes.html#templates-and-variables) I came across this statement:

> All templates will receive the variables defined in your settings file, as long as they are in all-caps. You can access them directly.

One _could_ infer that because PLUGINS is all-caps, that its use by templates to determine if a plugin was requested to be loaded was implied.

PR #2835, opened 2021-01-03
Stringify plugin definitions so they can be pickled during caching

m90:
# Pull Request Checklist

Resolves: #2828

This picks up on the idea from this particular comment:  https://github.com/getpelican/pelican/issues/2828#issuecomment-748368756 and fixes the issue (I can use both caching and module references just fine using the patch). I do have very limited knowledge about Pelican internals though, so I am not sure if this might be breaking something else somewhere.

### DIFF
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -20,7 +20,7 @@ from pelican.generators import (ArticlesGenerator,  # noqa: I100
                                 PagesGenerator, SourceFileGenerator,
                                 StaticGenerator, TemplatePagesGenerator)
 from pelican.plugins import signals
-from pelican.plugins._utils import load_plugins, stringify_plugins
+from pelican.plugins._utils import get_plugin_name, load_plugins
 from pelican.readers import Readers
 from pelican.server import ComplexHTTPRequestHandler, RootedHTTPServer
 from pelican.settings import coerce_overrides, read_settings
@@ -65,17 +65,18 @@ class Pelican:
             sys.path.insert(0, '')
 
     def init_plugins(self):
-        self.plugins = load_plugins(self.settings)
-        for plugin in self.plugins:
-            logger.debug('Registering plugin `%s`', plugin.__name__)
+        self.plugins = []
+        for plugin in load_plugins(self.settings):
+            name = get_plugin_name(plugin)
+            logger.debug('Registering plugin `%s`', name)
             try:
                 plugin.register()
+                self.plugins.append(plugin)
             except Exception as e:
                 logger.error('Cannot register plugin `%s`\n%s',
-                             plugin.__name__, e)
+                             name, e)
 
-        if 'PLUGINS' in self.settings and self.settings['PLUGINS'] is not None:
-            self.settings['PLUGINS'] = stringify_plugins(self.settings['PLUGINS'])
+        self.settings['PLUGINS'] = [get_plugin_name(p) for p in self.plugins]
 
     def run(self):
         """Run the generators and return"""
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -20,10 +20,10 @@ from pelican.generators import (ArticlesGenerator,  # noqa: I100
                                 PagesGenerator, SourceFileGenerator,
                                 StaticGenerator, TemplatePagesGenerator)
 from pelican.plugins import signals
-from pelican.plugins._utils import load_plugins
+from pelican.plugins._utils import load_plugins, stringify_plugins
 from pelican.readers import Readers
 from pelican.server import ComplexHTTPRequestHandler, RootedHTTPServer
-from pelican.settings import coerce_overrides, read_settings, stringify_plugins
+from pelican.settings import coerce_overrides, read_settings
 from pelican.utils import (FileSystemWatcher, clean_output_dir, maybe_pluralize)
 from pelican.writers import Writer
 
@@ -73,7 +73,9 @@ class Pelican:
             except Exception as e:
                 logger.error('Cannot register plugin `%s`\n%s',
                              plugin.__name__, e)
-        stringify_plugins(self.settings)
+
+        if 'PLUGINS' in self.settings and self.settings['PLUGINS'] is not None:
+            self.settings['PLUGINS'] = stringify_plugins(self.settings['PLUGINS'])
 
     def run(self):
         """Run the generators and return"""
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -23,7 +23,7 @@ from pelican.plugins import signals
 from pelican.plugins._utils import load_plugins
 from pelican.readers import Readers
 from pelican.server import ComplexHTTPRequestHandler, RootedHTTPServer
-from pelican.settings import coerce_overrides, read_settings
+from pelican.settings import coerce_overrides, read_settings, stringify_plugins
 from pelican.utils import (FileSystemWatcher, clean_output_dir, maybe_pluralize)
 from pelican.writers import Writer
 
@@ -73,6 +73,7 @@ class Pelican:
             except Exception as e:
                 logger.error('Cannot register plugin `%s`\n%s',
                              plugin.__name__, e)
+        stringify_plugins(self.settings)
 
     def run(self):
         """Run the generators and return"""
--- a/pelican/plugins/_utils.py
+++ b/pelican/plugins/_utils.py
@@ -1,6 +1,7 @@
 import importlib
 import importlib.machinery
 import importlib.util
+import inspect
 import logging
 import pkgutil
 import sys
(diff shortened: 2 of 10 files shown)


avaris:
I was writing a comment saying "I'm surprised this worked" and you updated :). I am assuming this is "finished" and commenting regarding that.

m90:
>  was writing a comment saying "I'm surprised this worked" and you updated :)

Sorry for that... The current state is good to review though. Thanks!

pelican/__init__.py
```diff
@@ -73,6 +73,7 @@ def init_plugins(self):
             except Exception as e:
                 logger.error('Cannot register plugin `%s`\n%s',
                              plugin.__name__, e)
+        stringify_plugins(self.settings)
```
avaris, line 76: It is probably better to stringify loaded plugins (`self.plugins` or subset if some failed) instead of the original `PLUGINS` setting.

pelican/settings.py
```diff
@@ -681,3 +681,15 @@ def coerce_overrides(overrides):
...
+
+
+def stringify_plugins(settings):
+    """
+    Plugins can be passed as module objects, however this breaks caching as
+    module objects cannot be pickled. To work around this, we stringify all
+    plugin definitions post-initialization.
+    """
+    if 'PLUGINS' in settings and settings['PLUGINS'] is not None:
+        settings['PLUGINS'] = [str(p) for p in settings['PLUGINS']]
```
avaris, line 693: I assume you reworked because you realized `str(p)` is a bit unhelpful. For modules `p.__name__` gives a better version. For class-like plugins, `p.__class__.__qualname__`.

pelican/settings.py
```diff
@@ -681,3 +681,15 @@ def coerce_overrides(overrides):
                              'load as json', k, v)
                 coerced[k] = json.loads(v)
     return coerced
+
+
+def stringify_plugins(settings):
```
avaris, line 686: I think it is better to put this in `pelican.plugins._utils`. And, I'd prefer this returned a list of strings instead of replacing the setting in-place.

pelican/tests/test_settings.py
```diff
@@ -319,3 +320,11 @@ def test_coerce_overrides(self):
             'THEME_STATIC_DIR': 'theme',
         }
         self.assertDictEqual(overrides, expected)
+
+    def test_stringify_plugins(self):
+        settings = {
+            'PLUGINS': ['string_plugin', NoopPlugin]
```
avaris, line 326: Well, technically, due to the way you structured `NoopPlugin` this won't "work". You need an instance (`NoopPlugin()`).

pelican/tests/test_settings.py
```diff
@@ -319,3 +320,11 @@ def test_coerce_overrides(self):
...
+        }
+        stringify_plugins(settings)
+        # Plugins are converted to string references
+        self.assertTrue(all(str(p) == p for p in settings['PLUGINS']))
```
avaris, line 330: It would also be nice to have an expected result for the value and test for that. Something like:
```python
self.assertCountEqual(
    ['string_plugin', 'NoopPlugin'],
    settings['PLUGINS']
)
```

PS: Of course the value `'NoopPlugin'` would depend on how you stringify them.

pelican/tests/test_settings.py
```diff
@@ -319,3 +320,11 @@ def test_coerce_overrides(self):
             'THEME_STATIC_DIR': 'theme',
         }
         self.assertDictEqual(overrides, expected)
+
+    def test_stringify_plugins(self):
```
avaris, line 324: It would probably makes more sense to move this test into `test_plugins.py`. And it would be nice to extend with some namespace and normal plugin tests (there are some already there).

avaris (changes requested):

pelican/__init__.py
m90: Would updating `self.plugins` also update the reference in the `settings` module, thus allowing us to pickle it?

pelican/tests/test_settings.py
m90: Yeah this is a remnant of not really having a good way of turning the module into a string. Using your suggestion from above this should work just fine.

pelican/__init__.py
```diff
@@ -74,6 +74,9 @@ def init_plugins(self):
                 logger.error('Cannot register plugin `%s`\n%s',
                              plugin.__name__, e)
 
+        if 'PLUGINS' in self.settings and self.settings['PLUGINS'] is not None:
+            self.settings['PLUGINS'] = stringify_plugins(self.settings['PLUGINS'])
+
```
avaris, line 79: I was thinking of a slightly different approach actually. We can simplify things. I'd say not bothering with `PLUGINS` setting at all here. `load_plugins` handles that and returns a list of objects that are plugin-like. All we need is a way to turn them into strings. I'd say scrapping `stringify_plugins` and changing `_stringify_plugin` to `get_plugin_name` with some modification. Code here would look like:

```python
    def init_plugins(self):
        self.plugins = []
        for plugin in load_plugins(self.settings):
            name = get_plugin_name(plugin)
            logger.debug('Registering plugin `%s`', name)
            try:
                plugin.register()
                self.plugins.append(plugin)
            except Exception as e:
                logger.error('Cannot register plugin `%s`\n%s',
                             name, e)
        self.settings['PLUGINS'] = [get_plugin_name(plugin) for plugin in self.plugins]
```

and `get_plugin_name` would look like:
```python
def get_plugin_name(plugin):
    try:
        return plugin.__class__.__qualname__
    except AttributeError:
        try:
            return plugin.__name__
         except AttributeError:
            pass
    return str(plugin)
```

PS: I changed `self.plugins` to store only registered plugins.

m90: Ah ok, now I get it. That makes a lot of sense.

pelican/plugins/_utils.py
```diff
@@ -107,3 +108,23 @@ def load_plugins(settings):
         plugins = list(namespace_plugins.values())
 
     return plugins
+
+
+def get_plugin_name(plugin):
```
m90, line 113: Over here: https://github.com/getpelican/pelican/blob/8033162ba4393db60791b201fb100d1be0f04431/pelican/tests/test_plugins.py#L85 things seem to check for a `NAME` attribute on a plugin. Do we need to do this too?

pelican/plugins/_utils.py
```diff
@@ -107,3 +108,23 @@ def load_plugins(settings):
...
+    """
+    Plugins can be passed as module objects, however this breaks caching as
+    module objects cannot be pickled. To work around this, all plugins are
+    stringified post-initialization.
+    """
+    if isinstance(plugin, str):
```
m90, line 119: These special cases are required to keep the function as sketched in your comment from returning `'str'` or `'type'`.

pelican/plugins/_utils.py
avaris: No, that was something I added to the tests. It's not a required interface for plugins. I suppose we can switch those to the `get_plugin_name` as well.

pelican/plugins/_utils.py
avaris: Well, for one, this should not receive a `str` at all. It will be passed loaded plugins which will be modules or similar plugin-like objects. And current conditions are a bit wrong. imported modules would not return correct values for example. Class as an input is a bit unusual but I guess you can force a class to behave like plugin (if you throw around `@staticmethod`s). Normal classes (like the one you wrote for test) would require an instance of it. Still we can account for a class.

So the values would be:
* `plugin.__name__` if it's a module
* `plugin.__qualname__` if it's a class (`if inspect.isclass(plugin):` would correspond to this, so it's not exactly correct right now)
* `plugin.__class__.__qualname__` (or `type(plugin).__qualname__`) if it's an instance. I think this will always succeed, so this should also cover everything else

Which could look like:
```python
def get_plugin_name(plugin):
    if inspect.ismodule(plugin):
        return plugin.__name__

    if inspect.isclass(plugin):
        return plugin.__qualname__

    return type(plugin).__qualname__
```

avaris (approved):
Thank you! Looks great.

avaris:
I'd like to also mention that this should partially resolve #2797 as well. `PLUGINS` won't be `None` anymore. `plugin_enabled` part is missing but that should be relatively simple after this.

m90:
@avaris Thanks for the thorough review, I definitely learned a lot on this one.

justinmayer:
@m90: Would you mind adding a `RELEASE.md` file to the root of the project, containing `Release type: patch` and a brief description of the changes, following the example mentioned in the [relevant docs](https://docs.getpelican.com/en/latest/contribute.html#using-git-and-github)? That way a new release — dedicated solely to the work you've done here — will be published when this PR is merged. 🎉

m90:
@justinmayer Done so.

justinmayer:
Many thanks to @m90 for the enhancement and to @avaris for reviewing. 💫

PR #2838, opened 2021-01-05
Use a top (instead of bottom) border in hentry

namn:
`hentry` uses `clear: both` with the intention of placing a separator line (1px border) between each entry. However, it is wrongly using `border-bottom` instead of `border-top` to make that separator. CSS `clear` makes space to clear the **preceding** floats. Hence a `border-bottom` won't enjoy that `clear` effect.


avaris:
I honestly don't understand the difference. Existing code places a border at the bottom of entries (except last one), yours places at the top (except first one). Functionally they seem to be identical (?).

namn:
That's correct. Functionally, they are almost identical. The difference is when the article text is too short, then the border will be right after the text, and cut through the post-info float on the left. This is because `clear` applies to preceding floats (i.e. the floats of previous hentry block), not the floats in this block.

avaris:
Ah, now I see what you mean, thank you for explanation. However, this will still have the potential issue of info bleeding into the next article if it is long, right? My CSS knowledge is a bit outdated but adding `overflow: auto` to `.hentry` would force a block element that contains including floats. Would that be better?

By the way, you need to [update test outputs](https://docs.getpelican.com/en/latest/contribute.html#running-the-test-suite) to match your changes so that tests will pass.

namn:
Auto:

auto

Border top:

border-top

Looks like there is a small difference in extra white space.

namn:
Functional tests updated. Tests passed. Would you prefer `overflow: auto` though?

namn:
Ping, @avaris

avaris (approved):

justinmayer:
Many thanks to @namn for the enhancement and to @avaris for reviewing. 🏅

Issue #2820, opened 2020-11-08
pygments version riquired <2.7.0, current version is 2.7.2


verga:
- [x ] I have read the [Filing Issues](https://docs.getpelican.com/en/latest/contribute.html#filing-issues) and subsequent “How to Get Help” sections of the documentation.
- [ x] I have searched the [issues](https://github.com/getpelican/pelican/issues?q=is%3Aissue) (including closed ones) and believe that this is not a duplicate.

- **OS version and name**: linux opensuse tumbleweed version N/A kernel 5.9
- **Python version**: 3.8.5
- **Pelican version**: 4.5.1
- **Link to theme**: custom
- **Links to plugins**: math
- **Link to your site**: verga.cpt.univ-mrs.fr
- **Link to your source**: 
- **Link to a [Gist](https://gist.github.com/) with the contents of your settings file**: 

## Issue

``pelican`` command send Error:

    pkg_resources.DistributionNotFound: The **'pygments<2.7.0,>=2.6.1'** distribution was not found and is required by pelican

My installed version is effectively **pygments==2.7.2**; I think that a dependency on a specific (old) version of a package is a bug.

Thanks!

avaris:
No, it's not a bug to have limitations on the versions of a dependency. [Using virtual environments](https://docs.getpelican.com/en/latest/install.html#installing-pelican) is recommended to avoid clashes between dependencies for different applications, precisely for this reason.

verga:
OK, thank you!

justinmayer:
@avaris is right. Pygments is pinned to a specific version because otherwise it causes things to break unexpectedly. Case in point: tests currently fail on Pygments v2.7.2. Pull requests to address that would be welcome! 😁

PR #2840, opened 2021-01-06
Loosen Pygments requirement in poetry

namn:
Currently, PyPI package is built with `~2.6.1` when Pygments' latest version is `2.7.3`. This creates a conflict with, e.g., Plumage theme, which asks for `^2.7.0`.


justinmayer:
Effective dup of #2820. Always a good idea to check existing issues first! 😅 

Have you installed Pygments 2.7.3 into your Pelican virtual environment and confirmed whether the Pelican test suite passes? (cc: @kdeldycke)

kdeldycke:
I don't know if the test suite of Pelican pass with Pygments 2.7.3, but I use the lastes Pygments without any issue on my theme for a while: https://github.com/kdeldycke/plumage/blob/5f1798ca9a80c7b938c7dbf1a6bdd1a21472bfde/pyproject.toml#L39

justinmayer:
Last I checked, Pygments' output changed in 2.7.x, which means Pelican's functional test output would need to be correspondingly updated in order to avoid test failures.

namn:
I don't observe any visual change to my blog either.

avaris:
Can you also [match the requirement](https://github.com/getpelican/pelican/blob/659c4cbf0f0cd823d5eb199220e5a087f167733d/requirements/test.pip#L2) in tests? That should almost certainly break the tests.

namn:
`test.pip` updated, as well as functional tests. I looked at the changes due to Pygments. They all seem to be simple CSS class change from `lineno` to `linenos`. Built-in CSS files don't define that class.

justinmayer (changes requested):
This topic is more complicated than one would might comprehend on first glance. I imagine there are at least two issues at play:

1. We don’t want people to encounter Pip installation errors because one or more projects have defined overly-restrictive dependency version constraints.
2. We don’t want functional tests to fail suddenly and unexpectedly just because Pygments released a new version with slightly-modified rendering.

We get into trouble when we confuse and/or conflate these two things, of which I have been altogether too guilty in the past, I might add. I think the takeaways are:

1. All Python projects, including Pelican and plugins, should choose the **lowest** minimum dependency versions that don’t cause things to break — not the most recent version.
2. Test environments should pin dependencies to specific versions when they otherwise might cause tests to fail unexpectedly.

@kdeldycke: To that end, unless lesser versions cause the theme to break in some horrible way, I think Plumage should also loosen its minimum dependency version for Pygments. Otherwise, I believe we are just asking for these kinds of dependency conflicts to proliferate.

justinmayer:
@namn: Could you upgrade your local dev environment to the just-released Pygments 2.7.4 and ensure the functional test output is not affected, and if it is, update it accordingly? Also, I'm seeing "no newline at end of file" changes for some of the functional test output files, which sometimes can be caused by text editors or IDEs that strip said newlines. Would you address those as well?

namn:
I simply ran `invoke update-functiona-tests` to regenerate the templates. No manual edits involved. My environment is macOS Catalina with Homebrew Python 3.9. Not sure what those new line changes are about. Perhaps docutils?

justinmayer:
Odd. I see the same behavior on my workstation. Seems it's only the "custom" index files that are affected. No idea why.

justinmayer (approved):

PR #2847, opened 2021-02-08
livereload task improvements

mirekdlugosz:
- use custom build command, with caching turned on - this reduces site build time by around 40% on my testing machines
- collect all glob patterns in a list and then call `server.watch` on each item - this allows to have single place where callback function must be specified
- use '**/*.html' as glob in template, to track changes in subdirectories (some themes include "partials" directory, where they store common widgets used in few different templates)

### DIFF
--- a/pelican/tools/templates/tasks.py.jinja2
+++ b/pelican/tools/templates/tasks.py.jinja2
@@ -99,23 +99,31 @@ def preview(c):
 def livereload(c):
     """Automatically reload browser tab upon file modification."""
     from livereload import Server
-    build(c)
+
+    def cached_build():
+        cmd = '-s {settings_base} -e CACHE_CONTENT=True LOAD_CONTENT_CACHE=True'
+        pelican_run(cmd.format(**CONFIG))
+
+    cached_build()
     server = Server()
-    # Watch the base settings file
-    server.watch(CONFIG['settings_base'], lambda: build(c))
-    # Watch content source files
+    theme_path = SETTINGS['THEME']
+    watched_globs = [
+        CONFIG['settings_base'],
+        '{}/templates/**/*.html'.format(theme_path),
+    ]
+
     content_file_extensions = ['.md', '.rst']
     for extension in content_file_extensions:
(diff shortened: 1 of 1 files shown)


pelican/tools/templates/tasks.py.jinja2
```diff
@@ -99,23 +99,31 @@ def preview(c):
 def livereload(c):
     """Automatically reload browser tab upon file modification."""
     from livereload import Server
-    build(c)
+
+    def cached_build():
+        cmd = '-s {settings_base} -e CACHE_CONTENT=True LOAD_CONTENT_CACHE=True'
+        pelican_run(cmd.format(**CONFIG))
+
+    cached_build(c)
```
justinmayer, line 107: Since `cached_build()` was defined as taking no arguments, won't invoking the function with `cached_build(c)` yield an error? Something like:
```
TypeError: cached_build() takes 0 positional arguments but 1 was given
```

justinmayer (commented):
Thanks for the submission, Mirek. There seem to be some interesting changes in here. Have you seen any instances in which caching interfered with normal development? (e.g., stale results due to improper cache invalidation, etc.)

pelican/tools/templates/tasks.py.jinja2
mirekdlugosz: Right!

mirekdlugosz:
> Have you seen any instances in which caching interfered with normal development? (e.g., stale results due to improper cache invalidation, etc.)

So far I have not encountered any problems with caching. I tried changing stylesheet and jinja template file, and I edited one article, all while this task was running. That's not much, but I have been using this only for few days.

I guess one's mileage may vary depending on installed plugins and what exactly they do. Also, I clean up cache directory after `liveserver` task, so I have easy and fast way of starting with known state. I can add something similar here, if you feel that would be worthwhile.

justinmayer:
My limited testing yielded no problems, so I think we can merge and address any issues as they arise.

Many thanks for the useful enhancement, Mirek! 🥇

Issue #2678, opened 2020-01-03
Is it possible write-selected just doesn't work?


mlissner:
Issue #2601 seems similar, but I can't figure out how to make `--write-selected` work. I've tried:

```bash
↪ pelican content -s pelicanconf.py -d -D --write-selected output/index.html
↪ pelican content -s pelicanconf.py -d -D --write-selected /home/me/programming/website/output/index.html
↪ pelican content -s pelicanconf.py -d -D --write-selected 'output/*'
↪ pelican content -s pelicanconf.py -d -D --write-selected ["output/index.html"]
```

And a dozen other things. Whenever I try these, they just don't generate any markdown files, though they do copy over a ton of assets. As far as I can tell, it really doesn't work, even with the example that is in the documentation.

I'm on version 4.2. If there's a fix here I should know about, I'd love to hear it. Maybe I'm just using it wrong? But I'm using output/index.html, which seems pretty hard to screw up.

mlissner:
OK, no, I don't think it ever works, and I think #2601 (opened August 2019, marked stale Oct. 1, 2019) was totally right. 

In utils.py there's a function that compares the `write-selected` setting/command line arg vs the file that's currently being processed. I added a few print statements, but otherwise it looks like this:

```
89 def is_selected_for_writing(settings, path):
890     '''Check whether path is selected for writing
891     according to the WRITE_SELECTED list
892 
893     If WRITE_SELECTED is an empty list (default),
894     any path is selected for writing.
895     '''
896     if settings['WRITE_SELECTED']:
897         print("Setting is: %s" % settings['WRITE_SELECTED'])
898         print("Path is: %s" % path)
899         if path not in settings['WRITE_SELECTED']:
900             print("Path is not in settings.")
901         return path in settings['WRITE_SELECTED']
902     else:
903         return True
```

Unfortunately, path generated by Pelican as a relative path and the `write-selected` value is an absolute path, even if you put a relative one into your command.

So, I use a relative path in my command, like so:

    pelican content -s pelicanconf.py --write-selected 'output/legal-api-downloader/index.html'

The tweaked code above yields:

    Setting is: [u'/home/mlissner/Programming/free.law/output/legal-api-downloader/index.html']
    Path is: output/legal-api-downloader/apple//index.html
    Path is not in setting.

Also note that `path` has two slashes, so I guess this doesn't support the `slug` field when it ends in a slash either. 

[...]

PR #2848, opened 2021-02-10
[WIP] Read content selectively

mirekdlugosz:
This is quick and dirty version of new setting that allows pelican to **read** content selectively, which provides tangible improvement of site build speed. See https://github.com/getpelican/pelican/issues/2678#issuecomment-776614289 for rationale. I am putting it here mostly as a reminder for me, and also to give a little head-start to anyone willing to work on this.

Speed comparison on my website:
- baseline: 0.91 second
- WRITE_SELECTED: 0.85 second
- READ_SELECTED: 0.13 second

Things I would like to get done before opening it for possible merge:

- [ ] Make READ_SELECTED a list of paths/patterns, and compare specific path to each of them
- [ ] Smarter handling of static files (especially static files that are `{attach}`ed in read post, but are outside of read paths themselves)
- [x] Don't pollute STDOUT with "errors" about skipped files (I'm thinking of creating new exception for that, and modifying all generators to only `logger.debug` when this exception is seen)
- [ ] Document new setting
- [ ] Remove `WRITE_SELECTED` support

Also, it makes perfect sense to me that `--autoreload` would use this feature to generate only files that actually changed, but that might require changes to `FileSystemWatcher`, so I would leave it for another PR.

### DIFF
--- a/pelican/generators.py
+++ b/pelican/generators.py
@@ -808,13 +808,21 @@ class StaticGenerator(Generator):
                 if self._is_potential_source_path(f):
                     continue
 
-            static = self.readers.read_file(
-                base_path=self.path, path=f, content_class=Static,
-                fmt='static', context=self.context,
-                preread_signal=signals.static_generator_preread,
-                preread_sender=self,
-                context_signal=signals.static_generator_context,
-                context_sender=self)
+            try:
+                static = self.readers.read_file(
+                    base_path=self.path, path=f, content_class=Static,
+                    fmt='static', context=self.context,
+                    preread_signal=signals.static_generator_preread,
+                    preread_sender=self,
+                    context_signal=signals.static_generator_context,
+                    context_sender=self)
+            except Exception as e:
+                logger.error(
+                    'Could not process %s\n%s', f, e,
+                    exc_info=self.settings.get('DEBUG', False))
+                self._add_failed_source_path(f, static=True)
+                continue
+
             self.staticfiles.append(static)
             self.add_source_path(static, static=True)
         self._update_context(('staticfiles',))
--- a/pelican/generators.py
+++ b/pelican/generators.py
@@ -14,7 +14,7 @@ from jinja2 import (BaseLoader, ChoiceLoader, Environment, FileSystemLoader,
 from pelican.cache import FileStampDataCacher
 from pelican.contents import Article, Page, Static
 from pelican.plugins import signals
-from pelican.readers import Readers
+from pelican.readers import PelicanFileSkipped, Readers
 from pelican.utils import (DateFormatter, copy, mkdir_p, order_content,
                            posixize_path, process_translations)
 
@@ -621,9 +621,12 @@ class ArticlesGenerator(CachingGenerator):
                         context_signal=signals.article_generator_context,
                         context_sender=self)
                 except Exception as e:
-                    logger.error(
-                        'Could not process %s\n%s', f, e,
-                        exc_info=self.settings.get('DEBUG', False))
+                    if isinstance(e, PelicanFileSkipped):
+                        logger.debug('Skipping %s', f)
+                    else:
+                        logger.error(
+                            'Could not process %s\n%s', f, e,
+                            exc_info=self.settings.get('DEBUG', False))
                     self._add_failed_source_path(f)
                     continue
 
@@ -726,9 +729,12 @@ class PagesGenerator(CachingGenerator):
                         context_signal=signals.page_generator_context,
                         context_sender=self)
                 except Exception as e:
-                    logger.error(
-                        'Could not process %s\n%s', f, e,
-                        exc_info=self.settings.get('DEBUG', False))
+                    if isinstance(e, PelicanFileSkipped):
+                        logger.debug('Skipping %s', f)
+                    else:
+                        logger.error(
+                            'Could not process %s\n%s', f, e,
+                            exc_info=self.settings.get('DEBUG', False))
                     self._add_failed_source_path(f)
                     continue
 
@@ -817,9 +823,12 @@ class StaticGenerator(Generator):
                     context_signal=signals.static_generator_context,
                     context_sender=self)
             except Exception as e:
-                logger.error(
-                    'Could not process %s\n%s', f, e,
-                    exc_info=self.settings.get('DEBUG', False))
+                if isinstance(e, PelicanFileSkipped):
+                    logger.debug('Skipping %s', f)
+                else:
+                    logger.error(
+                        'Could not process %s\n%s', f, e,
+                        exc_info=self.settings.get('DEBUG', False))
                 self._add_failed_source_path(f, static=True)
                 continue
 
--- a/pelican/readers.py
+++ b/pelican/readers.py
@@ -63,6 +63,10 @@ METADATA_PROCESSORS = {
 logger = logging.getLogger(__name__)
 
 
+class PelicanFileSkipped(Exception):
+    pass
+
+
 def ensure_metadata_list(text):
     """Canonicalize the format of a list of authors or tags.  This works
        the same way as Docutils' "authors" field: if it's already a list,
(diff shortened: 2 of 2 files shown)


avaris:
> Speed comparison on my website:
> 
> baseline: 0.91 second
> WRITE_SELECTED: 0.85 second
> READ_SELECTED: 0.13 second

I'm interested in the details of this benchmark. What exactly did you skip?

Also, I'd be very surprised (and confused) if you at all skipped a content (such as an article) and end up with proper, fully generated site at the end. You'd skip generating the output for that article, sure, but you'd also miss that from the generated index/tag/author/etc type collection pages.

mirekdlugosz:
> Also, I'd be very surprised (and confused) if you at all skipped a content (such as an article) and end up with proper, fully generated site at the end.

That's right, and generating proper full site is not the goal when using this setting.

There are situations when, as a site author, you want to trade correctness or thoroughness for speed. Think about editing article before publishing, or changing single template - at that point, you only care about this one page you see right now. But you want to see results of your change as soon as possible, because you have to decide if this change moves you in right direction.

This is exactly the same set of use-cases that `WRITE_SELECTED` tried to cover. Reasons for this proposal are:

* `WRITE_SELECTED` is currently broken, so nobody can use it anyway
* skipping content before reading should provide at least the same benefit as skipping content before writing - that is in worst case, when you read file and put it raw in template. In all other cases, skipping before read should give better results, as it skips more actions
* it provides easier to use API to users - author is more likely to know source path of file than output path, because they directly work with source paths

Similar to `WRITE_SELECTED`, this is not the setting you would keep in your publishconf, as the resulting output dir is likely to differ from "cold" build output dir.

> I'm interested in the details of this benchmark. What exactly did you skip?

I have run this patch against source code of [my website](https://github.com/mirekdlugosz/website), with `READ_SELECTED = '2020'`. This picks up 2 articles out of 30.

For `WRITE_SELECTED` runs, I had it set to match single article output path.

avaris:
> WRITE_SELECTED is currently broken, so nobody can use it anyway

That's a reason to fix it, not keep it broken and add tangentially similar functionality instead :).

> skipping content before reading should provide at least the same benefit as skipping content before writing - that is in worst case, when you read file and put it raw in template. In all other cases, skipping before read should give better results, as it skips more actions

Yes and no. `WRITE_SELECTED` will generate specified output page as it would've been in the full site. The implementation in this PR generates all of site with only a subset of the content. Two not equal results. Sites could be read heavy (lots of content or complicated plugins) or write heavy (complicated theme). Benefits would shift one way or another based on site.

> it provides easier to use API to users - author is more likely to know source path of file than output path, because they directly work with source paths

This, I agree.

For the record, you can achieve pretty much identical effect with the existing capabilities:
```
ARTICLE_PATHS = ['2020']
PAGE_PATHS = []
STATIC_PATHS = []
```

mirekdlugosz:
>  Sites could be read heavy (lots of content or complicated plugins) or write heavy (complicated theme). Benefits would shift one way or another based on site.

Could you share an example of site that is write heavy and where `WRITE_SELECTED` (if it was working) would provide better results than `READ_SELECTED`?

> PS: About speeding up general read time, did you try [cache](https://docs.getpelican.com/en/latest/settings.html#reading-only-modified-content)?

Yes, I am using it in my current setup. It reduces site build time by about 40%. In absolute terms, with caching I can build my site in 600ms; with `READ_SELECTED`, I can build it in 130ms.

avaris:
OK, I didn't see caching in your current config files. Just wanted to mention it. Yes, it's not a replacement for this kind of functionality but just a performance boost in general.

I don't have any examples off the top of my head. All I was trying to say is that these two work marginally different (trim input vs trim output), and it is probably premature to judge performance based on one site.

By the way, did you check my proposed settings above? If your goal is to limit input content, then you can already do that with those `*_PATHS` settings. I tried with your site and I get comparable results: 0.97s for full site, 0.15s for `*_PATHS` modified site.

mirekdlugosz:
> All I was trying to say is that these two work marginally different (trim input vs trim output), and it is probably premature to judge performance based on one site.

I have run my patch against some sites listed in [Powered by Pelican wiki page](https://github.com/getpelican/pelican/wiki/Powered-by-Pelican).

* site build time reduction provided by `WRITE_SELECTED` (even when it works) is between negligible and underwhelming - in only 2 sites (out of 16) reduction was by 40% or more
* meanwhile, `READ_SELECTED` consistently brought down site build times by 80-90%
* not a single site benefited from `WRITE_SELECTED` more than it did from `READ_SELECTED`
* although one site failed to build when `READ_SELECTED` was in effect

You can read all the details in [report](https://rpubs.com/mirekdlugosz/pelican-read-write-selected).

> If your goal is to limit input content

The goal is to provide near-instantaneous build times during local site development - something that other platforms can do, something that users today expect and something that Pelican fails to deliver.

`READ_SELECTED` is one of steps in that direction. Ideally, it would be baked in things like `pelican --autoreload` and users would never have to touch this setting directly.

avaris:
> You can read all the details in report.

Thanks, this is useful.

>> If your goal is to limit input content

> The goal is to provide near-instantaneous build times during local site development - something that other platforms can do, something that users today expect and something that Pelican fails to deliver.

> READ_SELECTED is one of steps in that direction. Ideally, it would be baked in things like pelican --autoreload and users would never have to touch this setting directly.

Yes, by limiting input content... I don't get why you keep ignoring part of what I said. `pelican` definitely is equipped to control input content:

```console
$ git clone https://github.com/mirekdlugosz/website.git 2>/dev/null
$ pelican --version
4.5.4
$ pelican content -s pelicanconf.py
Done: Processed 30 articles, 0 drafts, 5 pages, 0 hidden pages and 0 draft pages in 0.93 seconds.
$ echo -e "ARTICLE_PATHS=['2020']\nPAGE_PATHS=[]\nSTATIC_PATHS=[]" >> pelicanconf.py
$ pelican content -s pelicanconf.py
Done: Processed 2 articles, 0 drafts, 0 pages, 0 hidden pages and 0 draft pages in 0.16 seconds.
```

We can talk about whether or not it covers all the use cases and come to a solution if it fails to meet certain needs. But flat out ignoring this is not helpful. I'm not trying to be difficult. I certainly appreciate the contribution and the work you did, but this would require quite a bit more effort to be stable and usable.

mirekdlugosz:
> I don't get why you keep ignoring part of what I said.

For one, I have assumed that `ARTICLE_PATHS` must specify a directory. On this assumption, that setting would work for my simplistic test case I have started with, but would fail for sites like [bytefish.de](https://github.com/bytefish/bytefish.de/tree/master/blog), which put all articles in single directory. I have since learned that `ARTICLE_PATHS` can also point to a file, making it much more viable option.

Second, I wanted to make it clear that "we should just fix `WRITE_SELECTED`" is not a solution. In general, this setting falls flat on delivering the promise of shortening site build times. For that reason I initially wanted to remove this feature, but I don't really need to do that.

> Yes, by limiting input content...

I'm unsure how to read that. Are you fundamentally opposed to idea of reducing build time by limiting the amount of content Pelican should work on (regardless of how implemented, through new setting or existing ones)?

If you are, then how do you think Pelican should be changed to provide shorter feedback loops during local site development? Or do you believe this is not a problem Pelican has?

avaris:
I feel like I'm failing to convey my point of view.

I never said "let's just fix `WRITE_SELECTED` and forget about this". All I said was "`WRITE_SELECTED` is broken" is a reason to fix and/or improve that (or in the extreme case remove it if it is unsalvageable). Because the output is different, so this feature is not a direct replacement for it. `WRITE_SELECTED` generates output page *as it would have been in the full site*. You may not care about that, but someone might. It *does* provide a performance benefit, even though it might look underwhelming in some cases.

No, I'm not opposed to people limiting their input content, whatever their end goal may be (why would I?). You can read it as "if you want to limit content, here is how you can do that *right now* with existing functionality".

Issue #2834, opened 2021-01-01
Version 4.5.2+ error


dhjensen:
- [X] I have read the [Filing Issues](https://docs.getpelican.com/en/latest/contribute.html#filing-issues) and subsequent “How to Get Help” sections of the documentation.
- [X] I have searched the [issues](https://github.com/getpelican/pelican/issues?q=is%3Aissue) (including closed ones) and believe that this is not a duplicate.

- **OS version and name**: Linux Mint 19.3
- **Python version**: 3.6
- **Pelican version**: 4.5.2+
- **Link to theme**: https://github.com/getpelican/pelican-themes/tree/master/pelican-bootstrap3
- **Links to plugins**: https://github.com/getpelican/pelican-plugins/tree/master/i18n_subsites
- **Link to your site**: https://dhjensen.dk
- **Link to your source**: 
- **Link to a [Gist](https://gist.github.com/) with the contents of your settings file**: [Gist](https://gist.github.com/dhjensen/787719d30358fed7656a3d94ec4aabb5)

## Issue

Trying to upgrade from version 4.5.0 to 4.5.3 this "Error" start to occur in the log:
Note that the HTML generation seem to work fine.

```bash
$ make html
"pelican" "/home/dhjensen/Projects/sites/dhjensen.dk/content" -o "/home/dhjensen/Projects/sites/dhjensen.dk/output" -s "/home/dhjensen/Projects/sites/dhjensen.dk/pelicanconf.py" 
ERROR: Generator None (i18n_subsites.i18n_subsites) cannot be loaded
Done: Processed 2 articles, 0 drafts, 1 page, 0 hidden pages and 0 draft pages in 0.60 seconds.
```

In the code for 4.5.2 there are some changes that must be related.  
I do not understand enough about the inner workings of pelican to identify if this is working as intended or actually a bug.

avaris:
See getpelican/pelican-plugins#1314

PR #2850, opened 2021-02-12
ignore None return value from get_generators signal

avaris:
Resolves: #2834

Ignores if nothing was returned from `get_generators` signal. Some plugins were using this signal primarily because it provided a convenient place in the run timeline and did not return any generator to register. They were cluttering log with error messages even though everything was working as expected.

- [+] Ensured **tests pass** and (if applicable) updated functional test output
- [+] Conformed to **code style guidelines** by running appropriate linting tools

### DIFF
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -181,10 +181,10 @@ class Pelican:
         for receiver, values in signals.get_generators.send(self):
             if not isinstance(values, Iterable):
                 values = (values,)
-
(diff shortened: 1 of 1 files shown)


justinmayer (approved):
Looks good to me. Many thanks for fixing this.

Once this and the Pygments 2.8-related PR are merged, we can issue a new release containing both changes (and any other minor commits since the release last month).

Issue #1965, opened 2016-05-28
Hidden status for articles


rr-:
I hesitated before creating this issue, since there are at least two workarounds for this:
- Changing file extension to something like `.md~`
- Just using `status: draft`

But `hidden` has a slightly different semantic value than `draft`, plus if pages can be _hidden_ (without the notion of being a _draft_), why can't some posts too? I believe combining `draft` and `hidden` status to be an alias of each other would be pretty reasonable.

saschwarz:
According to the documentation that is an option: http://docs.getpelican.com/en/3.6.3/content.html?highlight=status#pages

I tried using it for a 404 page as described: http://docs.getpelican.com/en/3.6.3/tips.html?highlight=status#custom-404-pages

Unfortunately an error is generated: ``ERROR: Unknown status 'hidden' for file ./404.rst, skipping it.``

ingwinlu:
what would a hidden post do?

rr-:
Personally I'd like to mark some old posts as hidden so that they no longer appear on the main site (obsolete information etc.), but I can access them in the file system.

iKevinY:
Semantically, I think there's two distinct ideas being discussed here. Like with pages, a `hidden` post would be generated but not included in the index page (and perhaps the archive as well). In other words, you could still link to a `hidden` post via `example.com/blog/2016/06/my-post/`, much like you can link to a `hidden` page via `example.com/page`.

On the other hand, there could also be an `unpublished` status, where Pelican would ignore the file completely. One possible use case for this could be people who keep their Pelican content in a Git repository that automatically builds on pushes; this would allow them to iteratively work on a post without pushing anything new to their live site. It's pretty niche, but I think having a `unpublished → draft → published` "workflow" could be valuable.

alexwlchan:
I’ve been working around this with some ropey template code and monkey-patching bits of the feed generator. I’d be happy to have a go at incorporating this into Pelican core, if nothing else so I can tidy up my local install. :-)

I don’t have much free time this weekend, but I’ll try to put a patch together at some point.

justinmayer:
@rr- / @alexwlchan: Is this something you would like to work on implementing?

Telofy:
+1. My hack for now: I’ve set the article to draft and then configured a permanent redirect from the old URL to the draft URL in the Nginx I use to serve the files.

avaris:
[...]

Issue #2865, opened 2021-04-16
Feature Request: Unlisted articles


GiovanH:
- [x] I have searched the [issues](https://github.com/getpelican/pelican/issues?q=is%3Aissue) (including closed ones) and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://docs.getpelican.com/) and believe that my question is not covered.
- [x] I am willing to lend a hand to help implement this feature. 

## Feature Request

Currently, pages can be given the `status: hidden` attribute, which publishes them to the standard location but doesn't include them in page listings. This is great for making unlisted pages, or error pages. 

Articles also have a `status: hidden` attribute, but unlike pages this keeps them from being included in the output at all. You can set articles as `drafts`, but all drafts get lumped together in the `drafts` folder by default

I propose an `unlisted` status for articles that would behave the same as pages' `hidden` status: they would be output to the standard path, but not be included by default in listings (including tags, categories, and chrono)

If this seems like a useful addition, I am interested in helping to implement it.

justinmayer:
Hey Gio. This is a duplicate of #1965, which nobody stepped forward to implement. Regarding the following comment…

> You can set articles as `drafts`, but all drafts get lumped together in the `drafts` folder by default

As @avaris [mentioned](https://github.com/getpelican/pelican/issues/1965#issuecomment-495945811) in the aforementioned issue, you should be able to use `DRAFT_URL/DRAFT_SAVE_AS` to place drafts where you want.

That said, I can't think of any reason why setting `Status: hidden` for articles shouldn't behave in a similar fashion as it does for pages, and your offer to implement that feature parity for articles is much appreciated. Would you please submit a pull request when you have a solution ready for review? 😁

GiovanH:
I would be worried about changing the behavior of “hidden”, which currently has a different effect than unlisted (although the behavior technically seems undefined?). Not as catastrophic as automatically revealing hidden posts publicly, but still possibly undesirable behavior

justinmayer:
If setting `Status: hidden` causes articles not to be generated at all, then that seems like unintended behavior that I don't feel we need to preserve.

Since there's already an existing issue for this topic, I'm closing this one in favor of the pre-existing issue.

PR #2866, opened 2021-04-17
Support hidden posts

GiovanH:
# Pull Request Checklist

Resolves: #1965

This adds `hidden_articles` and `hidden_translations` to the article generator, as described in #1965 and #2865. There are some very basic test cases, including test cases for drafts.


docs/content.rst
```diff
@@ -614,6 +614,14 @@ the ``DEFAULT_METADATA``::
 To publish a post when the default status is ``draft``, update the post's
 metadata to include ``Status: published``.
 
+Hidden posts
+=================
+
+Like pages, posts can also be marked as ``hidden`` with the ``Status: hidden``
+attribute. Hidden posts will be output to the final path as expected, but
```
avaris, line 621: > Hidden posts will be **outputted** ...

Is it also worth mentioning that these follow `ARTICLE_SAVE_AS` as opposed to `DRAFT_SAVE_AS`?

pelican/generators.py
```diff
@@ -675,8 +688,10 @@ def _process(arts):
         self.authors = list(self.authors.items())
         self.authors.sort()
 
-        self._update_context(('articles', 'dates', 'tags', 'categories',
-                              'authors', 'related_posts', 'drafts'))
+        self._update_context((
+            'articles', 'drafts', 'hidden_articles',
+            'dates', 'tags', 'categories',
+            'authors', 'related_posts'))
```
avaris, line 694: This would also require an entry in the docs for [themes/common variables section](https://github.com/getpelican/pelican/blob/master/docs/themes.rst#common-variables)

avaris: Completely unrelated but pretty sure the `related_posts` here is unused :). Probably, left over from ancient times where related plugin was in the core.

avaris (changes requested):
Looks good overall. Left some minor comments.

GiovanH (commented):
@avaris Updated the documentation so the file is outdated now but let me know which way you want to go on the wording.

> I think "output" is technically acceptable here, but we can definitely change it to "outputted" for consistency. Up to you!

avaris (approved):
Looks good, thanks!

oulenz (approved):
LGTM, thanks for implementing this!

MinchinWeb:
This is cool; this is a feature I would see myself making use of.

Out of curiosity, what happens to a category page if all the entries in it are hidden? Would the category still be listed on the "Categories" page?

And the same question for labels, authors, archives, and period archive pages.

---

Also, when working on a theme/in the templating engine, is there a (easy) way to get a list of all articles, including the hidden ones? (something like `dates`). Here is the usecase I'm imagining:

- I generate a near-daily hidden post of my travels (rendered by a separate, to-be-written plugin). Included is a map overview.
- on the daily archive page, at the top, I include the map overview image, with a link to the (hidden) post with more details. So I need the daily archive page to be generated, even if all the posts on that day are hidden.
- from my list of period archive pages, I want to be able to a link the sub-period archive pages, including ones that only have hidden posts on them.

For reference, this is the current code that I'm using to generate links to the period archive pages (with similar loops for monthly and daily links): https://github.com/MinchinWeb/seafoam/blob/archive-periods/seafoam/templates/period_archives.html#L71..L147

```
{# Top Row with links to Archives by Year #}
<div class="archives-quicklinks text-center col-xs-12">
    {% for article in dates %}
        {% if loop.index0 == 0 or (article.date.year != dates[loop.index0 - 1].date.year) %}
            {% if loop.index0 != 0 %} &bull; {% endif -%}
            {%- if article.date.year == period[0] and period | length == 1 -%}
                <span class="active">{{ article.date.year }}</span>
            {%- else -%}
                <a href=
                    {%- if YEAR_ARCHIVE_URL -%}
                        "{{ SITEURL -}} /
                            {{- article.date | merge_date_url (YEAR_ARCHIVE_URL) }}"
                    {%- else -%}
                        "{{ SITEURL -}} / {{- ARCHIVES_URL | default('archives.html') -}}
                            #{{ article.date.year }}"
                    {%- endif -%}
                >
                    {{- article.date.year -}}
                </a>
            {%- endif -%}
        {%- endif -%}
    {%- endfor %}
</div>
```

The above relies on `dates`, which will only include non-hidden posts, so I'd need to adjust that variable somehow.

Thanks for your help!

GiovanH:
> Out of curiosity, what happens to a category page if all the entries in it are hidden? Would the category still be listed on the "Categories" page?
> 
> And the same question for labels, authors, archives, and period archive pages.

The behavior here would be the the same as with draft posts: drafts and hidden posts are not automatically added to those indexes, so if the only posts that use a category/tag/author is a draft (or hidden), that category index page wouldn't be generated.

> Also, when working on a theme/in the templating engine, is there a (easy) way to get a list of all articles, including the hidden ones? (something like `dates`). 

There's not a built-in variable for "published articles and drafts by date" or "published articles and hidden articles by date", but it should be fairly easy to write the equivalent in pure jinja. If this is something your site does regularly it should be fairly easy for the user to write an extra filter like `dates_all` that's just defined inline in pelicanconf.

docs/content.rst
```diff
@@ -614,6 +614,14 @@ the ``DEFAULT_METADATA``::
 To publish a post when the default status is ``draft``, update the post's
 metadata to include ``Status: published``.
 
+Hidden posts
+=================
```
justinmayer, line 618: Perhaps shorten this header underline to match the length of the header text above it?

docs/content.rst
```diff
@@ -614,6 +614,14 @@ the ``DEFAULT_METADATA``::
...
+
+Like pages, posts can also be marked as ``hidden`` with the ``Status: hidden``
+attribute. Hidden posts will be output to `ARTICLE_SAVE_AS` as expected, but
```
justinmayer, line 621:
```
`ARTICLE_SAVE_AS` → ``ARTICLE_SAVE_AS``
```

justinmayer (commented):
I made two minor reST-formatting suggestions. Other than that, I think this ready to go! 💫

MinchinWeb:
> so if the only posts that use a category/tag/author is a draft (or hidden), that category index page wouldn't be generated.

Is there currently a way to generate these index pages? I suppose these pages (by default) would be empty, but in theory, they would also be hidden. In any case, this (generating these index pages) isn't something I think I can do from Jinja.

(If that's not something you want to tackle, that's fine, but any tips on how to implement it myself would be appreciated.)

> There's not a built-in variable for "published articles and drafts by date" or "published articles and hidden articles by date", but it should be fairly easy to write the equivalent in pure jinja. If this is something your site does regularly it should be fairly easy for the user to write an extra list like dates_all that's just defined inline in pelicanconf.

A built in variable was what I was hoping for. I could probably put something together as a [Jinja Filter](https://github.com/pelican-plugins/jinja-filters) if it came to that. The inline entry in `pelicanconf.py` is an interesting idea, one that I hasn't considered before; it does make packaging and installing my theme more involved, but perhaps it's worth the tradeoff.

Thanks for your explanation.

justinmayer:
Manually merged via add3628a6474f2bd145fe2ea55c5e5255e9970ba. Will issue a release soon containing this new feature. Many thanks to @GiovanH for the enhancement and to @oulenz, @avaris, and @MinchinWeb for reviewing. 🥇

PR #2868, opened 2021-04-19
Fix `invoke tests` task problem on Windows

MinchinWeb:
PTY is not supported on Windows.

# Pull Request Checklist

- [ n/a ] Ensured **tests pass** and (if applicable) updated functional test output

Just something I stumbled across while working on `jinja_filters`.

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -42,7 +42,8 @@ def docserve(c):
 @task
 def tests(c):
     """Run the test suite"""
-    c.run(f"{VENV_BIN}/pytest", pty=True)
+    PTY = True if os.name != "nt" else False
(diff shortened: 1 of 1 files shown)


justinmayer:
Thanks, @MinchinWeb. Out of curiosity, what happens when running this Invoke task on Windows when PTY is set to `True`?

MinchinWeb:
~~~
λ invoke tests
You indicated pty=True, but your platform doesn't support the 'pty' module!
~~~

So it crashes out without running any tests.

justinmayer (approved):
Thanks for the additional detail. Always nice to have that recorded for future reference. ✨

PR #2869, opened 2021-04-21
Add progress spinner animation during generation

justinmayer:
I thought it would be nice to enhance Pelican's console output aesthetics via the [Rich][] project, and adding an animated spinner during site generation is the most naive initial implementation I could conceive.

There are [many other ways](https://github.com/willmcgugan/rich#rich-library) in which we could use [Rich][] to enhance console output, including prettier text colors, nicer logging formatting, and more readable tracebacks.

I'd love to replace the `Generating...` spinner in this PR with per-task progress bars (generating, writing, etc.). If someone wants to submit a follow-up PR to implement that, I'd be most grateful!  🚀

[Rich]: https://github.com/willmcgugan/rich

### DIFF
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -13,6 +13,8 @@ from collections.abc import Iterable
 from pkgutil import extend_path
 __path__ = extend_path(__path__, __name__)
 
+from rich.console import Console
+
 # pelican.log has to be the first pelican module to be loaded
 # because logging.setLoggerClass has to be called before logging.getLogger
 from pelican.log import init as init_logging
@@ -35,6 +37,7 @@ except Exception:
 
 DEFAULT_CONFIG_NAME = 'pelicanconf.py'
 logger = logging.getLogger(__name__)
+console = Console()
 
 
 class Pelican:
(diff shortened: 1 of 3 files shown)


Lucas-C:
Interesting small aesthetics improvement! 😊

My humble feedback: is it worth pulling all those extra dependencies?
https://github.com/willmcgugan/rich/blob/master/pyproject.toml#L29

MinchinWeb:
> My humble feedback: is it worth pulling all those extra dependencies?

`colorama` would likely be required for any simple effort to provide cross platform color CLI output. I'll let others comment on the rest.

iKevinY (approved):
LGTM; I might've had the same comment as @Lucas-C re: whether it's worth pulling in dependencies, but I _do_ like a good terminal spinner. 😄

avaris:
It has a considerably small footprint than I anticipated.

justinmayer:
> My humble feedback: is it worth pulling all those extra dependencies?

Is it worth it, just to get an animated spinner, as provided by this very naive implementation? Probably not. But as I mentioned above, I think we could expand our Rich usage to a degree that would, in aggregate, be worth the extra dependencies. Besides, I don't think the extra dependencies really cost us very much, and if we later decide they aren't worth it, we can easily remove them.

I look forward to someone building on top of this effort to make the most of what Rich can offer us. ✨

justinmayer:
If you are suggesting that dependencies can be an attack vector, well sure they can — that’s not saying very much.

Do I think Pelican users should be worried about Rich being such an attack vector? Nope.

PR #2878, opened 2021-05-20
Fix FILENAME_METADATA doc to render correctly

jzc:
I was rewriting my website with pelican and did a blind copy paste of the (alleged) FILENAME_METADATA default value from the docs as a starting point, but I was having trouble getting what I needed working. Then, I realized that the regexp I copied was different from the actual default value, as it got incorrectly rendered on the docs due to some missing backslashes. whoops!

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -815,7 +815,7 @@ Metadata
 
    The default metadata you want to use for all articles and pages.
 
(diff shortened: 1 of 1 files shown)


avaris (approved):

justinmayer:
Many thanks to @jzc for the fix and to @avaris for reviewing. 🏅

PR #2882, opened 2021-06-08
Allow easy subclassing of Writer

MinchinWeb:
When you write a custom Writer, it gets called with `settings=None`. If you writer is simply a subclass of the built-in Writer, Pelican will through the error `CRITICAL: 'RELATIVE_URLS'`.

The source of the error is from `Pelican._get_writer()` in `__init__.py`.

With this in place, your custom writer can be as simple as:

~~~python

class CustomWriter(Writer):
    pass
~~~

I realize the above example doesn't do much, but it makes starting easier.

# Pull Request Checklist

Resolves: #issue-number-here 

- [?] Added **tests** for changed code

### DIFF
--- a/pelican/writers.py
+++ b/pelican/writers.py
@@ -25,7 +25,7 @@ class Writer:
         self._overridden_files = set()
 
         # See Content._link_replacer for details
-        if self.settings['RELATIVE_URLS']:
+        if "RELATIVE_URLS" in self.settings and self.settings['RELATIVE_URLS']:
             self.urljoiner = posix_join
         else:
             self.urljoiner = lambda base, url: urljoin(


MinchinWeb:
The tests seem to be failing (every test!) because Jinja 3 is throwing a deprecation warning: `DeprecationWarning: 'jinja2.Markup' is deprecated and will be removed in Jinja 3.1. Import 'markupsafe.Markup' instead.`

justinmayer:
@MinchinWeb: I pushed a test branch containing what I believe to be a fix for those test failures on Jinja 3+: https://github.com/getpelican/pelican/commit/c041bf2192dbf95cb2ac63f1ea979a0d955690b7  Does that resolve the issue for you?

MinchinWeb:
@justinmayer Yes! With your commit, the test pass again (at least locally).

It looks like `markup.Markup` exists on the 1.1 branch (see [here](https://markupsafe.palletsprojects.com/en/1.1.x/escaping/)), so there shouldn't be a need to bump the minimum versions of either Jinja or MarkupSafe.

I'm getting three tests that fail locally because I don't have *diff* installed, but that's a separate issue, on those tests fail with both Jinja 2 and 3.

justinmayer:
Above-mentioned commit has been merged to `master`.

avaris:
> When you write a custom Writer, it gets called with `settings=None`.

Where is it called with `None`?

https://github.com/getpelican/pelican/blob/c041bf2192dbf95cb2ac63f1ea979a0d955690b7/pelican/__init__.py#L215-L228

MinchinWeb:
@avaris Line 216.

`settings=None` is the default value for `settings`, and used  when, as in this case, the Pelican settings isn't passed to the class on initialization.

The traceback:

~~~
  File "/home/william/Code/gpx-reader/.venv/bin/pelican", line 8, in <module>
    sys.exit(main())
  File "/home/william/Code/gpx-reader/.venv/lib/python3.9/site-packages/pelican/__init__.py", line 527, in main
    pelican.run()
  File "/home/william/Code/gpx-reader/.venv/lib/python3.9/site-packages/pelican/__init__.py", line 121, in run
    writer = self._get_writer()
  File "/home/william/Code/gpx-reader/.venv/lib/python3.9/site-packages/pelican/__init__.py", line 207, in _get_writer
    writers = [w for _, w in signals.get_writer.send(self) if isinstance(w, type)]
  File "/home/william/Code/gpx-reader/.venv/lib/python3.9/site-packages/blinker/base.py", line 266, in send
    return [(receiver, receiver(sender, **kwargs))
  File "/home/william/Code/gpx-reader/.venv/lib/python3.9/site-packages/blinker/base.py", line 266, in <listcomp>
    return [(receiver, receiver(sender, **kwargs))
  File "/home/william/Code/gpx-reader/.venv/lib/python3.9/site-packages/pelican/writers.py", line 28, in __init__
    if self.settings['RELATIVE_URLS']:
KeyError: 'RELATIVE_URLS'
~~~

avaris:
Which class initialization? I feel like there is a confusion. `get_writer` signal needs to return a Writer *class*, **not** instance. Writer object instantiation happens at the end of that method (line 228). Also your traceback suggests that settings is in fact **not `None`**, it is a `dict` but missing `RELATIVE_URLS` key.

avaris (approved):

avaris:
I'm approving this since it's a good addition, however it is not the source of your error :).

MinchinWeb:
@justinmayer @avaris I think this is ready to merge... (thanks!)

justinmayer:
Thanks to @MinchinWeb for the enhancement and to @avaris for reviewing. 🏅

PR #2886, opened 2021-06-09
Fix formatting of Jinja example code. Fixes #2884

### LINKED ISSUE #2884: Example code for JINJA_FILTERS setting is misformatted
- [X] I have searched the [issues](https://github.com/getpelican/pelican/issues?q=is%3Aissue) (including closed ones) and believe that this is not a duplicate.

## Issue

In the [Settings documentation](https://docs.getpelican.com/en/latest/settings.html), the description of the `JINJA_FILTERS` setting includes some example code which is wrapped in `<p>…</p>` tags rather than `<code>…</code>` tags, which has the effect of suppressing display of important newlines.  The example code also uses smart quotes (`‘…’`) rather than upright quotes (`'…'`) which I suspect won't be recognized by the Python interpreter.

logological:
# Pull Request Checklist

Resolves: #2884 

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+Release type: patch
+
+Fix formatting of Jinja example code
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,3 +1,0 @@
-Release type: patch
-
-Fix formatting of Jinja example code
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -104,6 +104,7 @@ Basic settings
    should map the filtername to the filter function.
 
    Example::
+
     import sys
     sys.path.append('to/your/path')
 


justinmayer:
Hi Tristan. Thank you for the contribution. Two thoughts:

1. In the future, for these kinds of changes, there is no need to submit both an issue and a pull request. You can forgo the issue and just submit the pull request.
2. It seems this PR only adds a single newline and does not contain the changes described in the corresponding issue. Would you please review and advise?

logological:
My understanding is that the single newline is what fixes the issue.  ReStructured Text requires preformatted text to be separated from the previous paragraph with a double colon followed by a blank line, and it was the blank line that was missing from the documentation.

justinmayer:
Got it. Makes sense. Thanks for the fix, Tristan! 👍

Issue #2896, opened 2021-06-30
"Generating..." leaves artifacts in debug output


MinchinWeb:
A spinner was added post 4.6 (which is snazzy!), but it leaves an artifact on the following debug lines.  Ideally these lines would be overwritten rather than appended.

![image](https://user-images.githubusercontent.com/1548809/123890693-e507d080-d914-11eb-9535-97ceadecffd4.png)

- [x] I have read the [Filing Issues](https://docs.getpelican.com/en/latest/contribute.html#filing-issues) and subsequent “How to Get Help” sections of the documentation.
- [x] I have searched the [issues](https://github.com/getpelican/pelican/issues?q=is%3Aissue) (including closed ones) and believe that this is not a duplicate.

- **OS version and name**:  Windows 10
- **Python version**:  3.9.6
- **Pelican version**:   master at e800b23b4d9ec45663dbac6c5cb4775b0cacdb57 (post 4.6)
- **Link to theme**:  
- **Links to plugins**:  
- **Link to your site**:  
- **Link to your source**:  not yet on GitHub
- **Link to a [Gist](https://gist.github.com/) with the contents of your settings file**:

MinchinWeb:
Waiting on #2897

justinmayer:
Ostensibly addressed by #2897.

PR #2897, opened 2021-07-01
Rich logging

MinchinWeb:
Further to #2869, this expands *rich*'s usage by extending it to cover logging.

This has the side effect of somewhat simplifying the logging code we need "in house" as some of this is pushed to *rich*.

Current (v4.6.0) logging:

pelican-old-logging

Logging with this PR:

pelican-rich-logging

# Pull Request Checklist

Re formatting: On my side I'm showing quite a large number of files that would be reformatted by *black* and *isort*. I'm not going to make those changes here...

### DIFF
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -222,7 +222,7 @@ class Pelican:
 
         writer = writers[0]
 
-        logger.debug("Found writer: %s (%s)" % (writer.__name__, writer.__module__))
+        logger.debug("Found writer: %s (%s)", writer.__name__, writer.__module__)
         return writer(self.output_path, settings=self.settings)
 
 
@@ -536,7 +536,7 @@ def main(argv=None):
     except KeyboardInterrupt:
         logger.warning('Keyboard interrupt received. Exiting.')
     except Exception as e:
-        logger.critical("%s: %s" % (e.__class__.__name__, e))
+        logger.critical("%s: %s", e.__class__.__name__, e)
 
         if args.verbosity == logging.DEBUG:
             console.print_exception()
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -14,6 +14,7 @@ from pkgutil import extend_path
 __path__ = extend_path(__path__, __name__)
 
 from rich.console import Console
+from rich.logging import RichHandler
 
 # pelican.log has to be the first pelican module to be loaded
 # because logging.setLoggerClass has to be called before logging.getLogger
@@ -500,7 +501,8 @@ def listen(server, port, output, excqueue=None):
 def main(argv=None):
     args = parse_arguments(argv)
     logs_dedup_min_level = getattr(logging, args.logs_dedup_min_level)
-    init_logging(level=args.verbosity, fatal=args.fatal, name=__name__,
+    init_logging(level=args.verbosity, fatal=args.fatal,
+                 handler=RichHandler(console=console), name=__name__,
                  logs_dedup_min_level=logs_dedup_min_level)
 
     logger.debug('Pelican version: %s', __version__)
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -13,11 +13,9 @@ from collections.abc import Iterable
 from pkgutil import extend_path
 __path__ = extend_path(__path__, __name__)
 
-from rich.console import Console
-from rich.logging import RichHandler
-
 # pelican.log has to be the first pelican module to be loaded
 # because logging.setLoggerClass has to be called before logging.getLogger
+from pelican.log import console
 from pelican.log import init as init_logging
 from pelican.generators import (ArticlesGenerator,  # noqa: I100
                                 PagesGenerator, SourceFileGenerator,
@@ -38,7 +36,6 @@ except Exception:
 
 DEFAULT_CONFIG_NAME = 'pelicanconf.py'
 logger = logging.getLogger(__name__)
-console = Console()
 
 
 class Pelican:
@@ -502,8 +499,7 @@ def main(argv=None):
     args = parse_arguments(argv)
     logs_dedup_min_level = getattr(logging, args.logs_dedup_min_level)
     init_logging(level=args.verbosity, fatal=args.fatal,
-                 handler=RichHandler(console=console), name=__name__,
-                 logs_dedup_min_level=logs_dedup_min_level)
+                 name=__name__, logs_dedup_min_level=logs_dedup_min_level)
 
     logger.debug('Pelican version: %s', __version__)
     logger.debug('Python version: %s', sys.version.split()[0])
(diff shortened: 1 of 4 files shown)


avaris:
Shiny :sparkles: . I'll review once I have time to test this. By the way, does this help with #2896 ?

MinchinWeb:
Ah yes, this would close #2896!

The "Generating..." spinner is still there (if you're sharpeyed, you might see it in the above screen cast); the trick was to use the same *console*.

pelican/__init__.py
```diff
@@ -500,7 +501,8 @@ def listen(server, port, output, excqueue=None):
 def main(argv=None):
     args = parse_arguments(argv)
     logs_dedup_min_level = getattr(logging, args.logs_dedup_min_level)
-    init_logging(args.verbosity, args.fatal,
+    init_logging(level=args.verbosity, fatal=args.fatal,
+                 handler=RichHandler(console=console), name=__name__,
```
MinchinWeb, line 505: Pass the same *console* here so the logging and spinners play nice together.

avaris: I'd prefer if this was handled in the `log.py` (as part of default `init_logging`) and `__init__.py` would import the `console` from `log.py`.

pelican/__init__.py
```diff
@@ -538,9 +540,9 @@ def main(argv=None):
     except KeyboardInterrupt:
         logger.warning('Keyboard interrupt received. Exiting.')
     except Exception as e:
-        logger.critical('%s', e)
+        logger.critical("%s: %s" % (e.__class__.__name__, e))
 
         if args.verbosity == logging.DEBUG:
-            raise
+            console.print_exception()
```
avaris, line 546: This makes process finish "successfully" (return code 0). Should be non-zero. Probably easier to remove `else` and run that block unconditionally:
```python
        if args.verbosity == logging.DEBUG:
            console.print_exception()

        sys.exit(getattr(e, 'exitcode', 1))
```

pelican/__init__.py
```diff
@@ -538,9 +540,9 @@ def main(argv=None):
     except KeyboardInterrupt:
         logger.warning('Keyboard interrupt received. Exiting.')
     except Exception as e:
-        logger.critical('%s', e)
+        logger.critical("%s: %s" % (e.__class__.__name__, e))
```
avaris, line 543: Don't pre-format logs and let the logger handle them please :)
```python
logger.critical("%s: %s", e.__class__.__name__, e)
```

avaris (changes requested):
Left some comments, but overall looks really nice. Thanks!

pelican/__init__.py
MinchinWeb: This code will give you:
```
[21:27:20] CRITICAL ZeroDivisionError: division by zero
```

the old code will give you:
```
[21:29:08] CRITICAL division by zero
```

Ideally the logger should handle the formatting, but I feel like the default is failing us here as it looses the "type" of error that is being throw.

Are you certain you want to hide the error "type"?

pelican/__init__.py
MinchinWeb: done!

pelican/__init__.py
MinchinWeb: done!

MinchinWeb:
This also updates the formating of the debug message to match the Generators (below is with rich logging active):

```
           DEBUG    Found generator: ArticlesGenerator (internal)                                        __init__.py:211
           DEBUG    Found generator: PagesGenerator (internal)                                           __init__.py:211
           DEBUG    Found generator: GPXGenerator (pelican.plugins.gpx_reader)                           __init__.py:211
           DEBUG    Found generator: StaticGenerator (internal)                                          __init__.py:211
...
           DEBUG    Found writer: XMLWriter (pelican.plugins.xml_writer.writer)                          __init__.py:228
```

pelican/__init__.py
avaris: I am not hiding it? Difference between my suggestion and your is that, you pre-format the message using type and error text (`%` operator) before sending it to logger, mine sends those as arguments to the logger (`,` in between string and arguments).

MinchinWeb: bc21922cf217c4e05f9c28fbc30349271735798e resolves this. I'd missed the `,` vs `%`...

MinchinWeb:
I've also expanded the use the `rich.console` to the ``--print-settings` CLI option, particularly when there is an error in your `pelicanconf.py`.

General errors (as run through the `FatalLogger` / `LimitLogger` only display the error message but not the error class. I'm not sure how to "fix" this, but it was "broken" before I started...

avaris:
> General errors (as run through the `FatalLogger` / `LimitLogger`) only display the error message but not the error class. I'm not sure how to "fix" this, but it was "broken" before I started...

Which ones are those? If you mean `logger.error(...)` cases, then those are only logs, they are not necessarily associated with an exception, so no error class.

justinmayer:
_Unrelated note:_ Maybe we could add support for [Better Exceptions](https://github.com/qix-/better-exceptions), much like my friend Hynek [just did for Structlog](https://github.com/hynek/structlog/pull/330).

avaris:
Somewhat related note: The only remaining non-rich output is the `server.py` (and a [`print`](https://github.com/getpelican/pelican/blob/332be6e5c8005f699a708568939d31dd85e9cffb/pelican/__init__.py#L485) in it's counterpart in `listen` function from `__init__.py`).  To make the server use `console`, simply defining a `log_message` method in the [handler](https://github.com/getpelican/pelican/blob/332be6e5c8005f699a708568939d31dd85e9cffb/pelican/server.py#L42) and replace the [original `sys.stderr.write`](https://github.com/python/cpython/blob/95596d5921eeab9ae49f0dc31263a249013b6849/Lib/http/server.py#L561-L581) with `console.print` or `console.log` would be enough.

I was also thinking about making [`autoreload` message](https://github.com/getpelican/pelican/pull/2897/files#diff-211306d4da1b42c2361ad09ffa2f11ee9803c637df04f6a8ba482cd6ebec631bR436-R437) a status/spinner as well but `-lr` (`--listen --autoreload`) makes things problematic. Even if you "share" console, server messages disrupt status and results in similar issue as #2896. I'm 99% certain that the culprit is `multiprocessing` and in fact the `console` instances are not really same, but solution will be tricky.

justinmayer:
@avaris: Any chance you could take another look at this and update your review with any remaining items to be addressed, if any?

justinmayer:
@avaris: What do you think we should do in terms of moving this forward?

avaris (approved):
Thanks

avaris:
Sorry about that, I think this is good as is. We can always add further enhancements later.

justinmayer:
@avaris: No need to be sorry! You raised some good points, which hopefully we can use as guidance for further improvements in the future.

@MinchinWeb: Many thanks for these enhancements! Very cool, indeed. 💯

PR #2899, opened 2021-07-07
Really Allow Alternate Writers

MinchinWeb:
Further to #2882, the blinker signal returns both the *class of the custom Writer* and an *instance of the custom Writer*. Previously, Pelican was only keeping the instance, and this would fail the `isinstance(w, type)` check, which meant that Pelican would always default to the built-in Writer. This now keeps the class reference instead and actually allows one to provide a custom Writer.

This also updates the formating of the debug message to match the Generators (below is with rich logging active):

```
           DEBUG    Found generator: ArticlesGenerator (internal)                                        __init__.py:211
           DEBUG    Found generator: PagesGenerator (internal)                                           __init__.py:211
           DEBUG    Found generator: GPXGenerator (pelican.plugins.gpx_reader)                           __init__.py:211
           DEBUG    Found generator: StaticGenerator (internal)                                          __init__.py:211
...
           DEBUG    Found writer: XMLWriter (pelican.plugins.xml_writer.writer)                          __init__.py:228
```

I am working on a custom writer, and so will update here when I have it on GitHub so you can test it locally.

### DIFF
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -213,7 +213,7 @@ class Pelican:
         return generators
 
     def _get_writer(self):
-        writers = [w for _, w in signals.get_writer.send(self) if isinstance(w, type)]
+        writers = [w for w, _ in signals.get_writer.send(self) if isinstance(w, type)]
         num_writers = len(writers)
 
         if num_writers == 0:
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -224,7 +224,7 @@ class Pelican:
 
         writer = writers[0]
 
-        logger.debug("Found writer: %s", writer)
+        logger.debug("Found writer: %s (%s)" % (writer.__name__, writer.__module__))
         return writer(self.output_path, settings=self.settings)
 
 


MinchinWeb:
I working custom writer is available at https://github.com/MinchinWeb/gpx-reader/tree/483b1ca24d1f4ad800870c25f47d7c6e2f0608bf

Install the PR's version of Pelican, and the plugins at the above link, and then go to the "test-site" directory and run `pelican --debug`.

With the current code (i.e. before this PR), it will error out complaining about no "write_xml() method" (which is implemented by the custom writer). With this PR's code, you will notice a line "Found writer: EverythingWriter (pelican.plugins.everything_writer.writer)" and the site will generate to completion.

avaris:
> Further to #2882, the blinker signal returns both the class of the custom Writer and an instance of the custom Writer.

Are you sure, you are using it correctly?

You connect a function to `blinker` signal, and it's return value of that function is used. You're passing the Class to `connect`. That's not how it's supposed to work. Blinker is the one "calling" your class (i.e. instantiating it), because that's what it does. It calls all the connected functions with the attributes passed and collect return values. Original code is correct and here is how it's supposed to be used:

```python
from pelican import signals

class MyWriter(Writer):
    pass


def add_writer(pelican_instance):
    # use pelican_instance to setup stuff if needed
    return MyWriter


def register():
    signals.get_writer.connect(add_writer)
```

MinchinWeb:
> Are you sure, you are using it correctly?

Ahh... no, no idea... lol.

I could find very (no?) documentation specific to making a custom writer, and no implementation examples, so I've been at the *through stuff against the wall* stage intermittent with the *bang head against the wall* stage (all the while, trying to avoid getting code stuck in my hair ;) ). It doesn't help that I can't seem to get editable installs of Pelican to work, so I've gotten this far by using VSCode's debugger to step through Pelican's code and then my plugin's code to try and figure out how this is supposed to work.

Your code example works on my end, so I've implemented that and added it to the docs (#2901).

I'll close this for now. Thanks for your help and patience.

avaris:
> I could find very (no?) documentation specific to making a custom writer

Probably because the `get_writer` signal works pretty much identical to `get_generators` and that one has (some) documentation, but more docs is always a good thing :).

> It doesn't help that I can't seem to get editable installs of Pelican to work

Why not? What's the issue?

MinchinWeb:
> > It doesn't help that I can't seem to get editable installs of Pelican to work
> 
> Why not? What's the issue?

As I go back and look at this, I think it was just the appearance of not working. I usually check if an edit in the editable install stuck by editing the version number and running `program --version`, but with Pelican, the code version is determined by inspecting the version as installed by pip (which doesn't update until you re-run `pip install .`).

Also, editable installs don't work with *poetry*-managed packages, and as the plugins have been moved to *poetry*, I'd assumed that Pelican had been too when the editable installs appears not to work.

There was so much going on, I just didn't want to try and figure pelican's install status at the same time.

@avaris @justinmayer Would you support a PR that would swap the version number management so that the version is pulled from the main code body and have the installer/setup.py pull from the main code body. I would use code [like I did with Seafoam](https://github.com/MinchinWeb/seafoam/blob/master/setup.py#L34).

justinmayer:
FWIW, I tried my best to document how to set the development environment: https://docs.getpelican.com/en/latest/contribute.html#setting-up-the-development-environment

> Would you support a PR that would swap the version number management so that the version is pulled from the main code body and have the installer/setup.py pull from the main code body.

Hard to say when you haven't described what the perceived benefits would be. 😉 

Ideally I'd prefer that the version number is only specified in `pyproject.toml` and nowhere else. It's only listed in `setup.py` right now out of inertia.

The only reason Pelican hasn't been fully switched over to Poetry is (1) aforementioned Setuptools inertia, and (2) I thought it best to see how Poetry would evolve, including whether the editable install story might improve.

MinchinWeb:
@justinmayer 

> FWIW, I tried my best to document how to set the development environment: https://docs.getpelican.com/en/latest/contribute.html#setting-up-the-development-environment

Indeed, this is helpful. Unfortunately, I'm...special. Start with the fact that I'm working on Windows, and so the commands can't be run directly as listed (in particular, you don't `source` to activate a virtual environment, but run the activation script directly). Then add to that that I have about 4 plugins in play and the theme too, and it gets crazy in a hurry.

> Hard to say when you haven't described what the perceived benefits would be. 😉

That `pelican --version` would reflect the codebase installed rather than version number originally installed into pip. The biggest advantage is that `pelican --version` works are a very quick check that the install is editable and not a "fixed" install. I.e. change the  version number, and run `pelican --version` and a super quick proxy that your other changes have been picked up. When I'm dealing with some many moving pieces, it's helpful to have quick checks like this.

PR #2902, opened 2021-07-11
Support date format codes G, V, and u (used by ISO dates)

MinchinWeb:
Current behaviour:

~~~python-repl
>>> from pelican.utils import SafeDatetime
>>> from datetime import datetime
>>> "{a:%G}-{a:%V}-{a:%u}".format(a=SafeDatetime.now())
'%G-%V-%u'
>>> "{a:%G}-{a:%V}-{a:%u}".format(a=datetime.now())
'2021-27-7'
~~~

(`%G` gives you the ISO year, `%V` gives you the ISO week, and `%u` gives you the ISO day)

These codes aren't part of the C89 spec, but were added in Python 3.6 (which is the minimum officially supported version of Python for Pelican), so I don't expect any issues with supporting them.

(As a side, this also lead me to discover why `%-d` works in Pelican and nowhere else!)

**Usecase**: I want to create a weekly "period archive" and it would be simpliest to use the ISO dating, as that is already built into Python's standard library.

This also adds a test to make sure these are working as expected.

# Pull Request Checklist

- [ ? ] Updated **documentation** for changed code

### DIFF
--- a/pelican/tests/test_utils.py
+++ b/pelican/tests/test_utils.py
@@ -513,6 +513,9 @@ class TestUtils(LoggedTestCase):
         d = utils.SafeDatetime(2012, 8, 9)
         self.assertEqual(utils.strftime(d, '%-d/%-m/%y'), '9/8/12')
 
+        d = utils.SafeDatetime(2021, 1, 8)
+        self.assertEqual(utils.strftime(d, '%G - %-V - %u'), '2021 - 1 - 5')
+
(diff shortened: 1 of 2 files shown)


avaris (approved):
:+1:

justinmayer:
Thanks to @MinchinWeb for the enhancement and to @avaris for reviewing. 💫

PR #2904, opened 2021-08-01
Pickle.dump can raise TypeError (e.g. generators)

GiovanH:
Pickle.dump can raise TypeError (e.g. generators) but this wasn't caught, leading to a critical failure instead of the appropriate warning.

Didn't create an issue since this was such a minor logging change.


avaris (approved):

justinmayer:
Many thanks to @GiovanH for the fix and to @avaris for reviewing. 😊

Issue #2917, opened 2021-09-08
truncate_html_words() breaks with Jinja2's autoescape and the end_text being Markup()


galaxy4public:
- [x] I have read the [Filing Issues](https://docs.getpelican.com/en/latest/contribute.html#filing-issues) and subsequent “How to Get Help” sections of the documentation.
- [x] I have searched the [issues](https://github.com/getpelican/pelican/issues?q=is%3Aissue) (including closed ones) and believe that this is not a duplicate.

## Issue

When a template file contains something like the following:
```
{% set readmore -%}
     <span class="readmore">… <a href="{{ BASEURL }}/{{ article.url }}#continue" class="icon fa-forward">read </a></span>
{%- endset %}
{{ article.content|truncate_html_words(50, readmore)|safe }}
```
It works perfectly with the default Pelican settings and produces a properly truncated article with the addition of the custom end marker.  However, if one decides to enable Jinja2's `autoescape` functionality (which is coming to be a default setting pretty soon), by setting `JINJA_ENVIRONMENT` in `pelicanconf.py` to:
```
JINJA_ENVIRONMENT = {
    'autoescape': True,
}
```

The generated output of the template would suddenly contain the entire article's content in the escaped form while injected end marker will be in a valid HTML form.  If you try to rectify the situation by using filters, it will just invert parts -- in all times a part of the output of `truncate_html_words()` will be escaped.

This happens because when Jinja2 is configured for `autoescape` variable blocks (as in `{%set variable %}...{% endset %}` are automatically assigned to a safe class, which is `Markup()`.  This class ensures that if you add something unsafe to it (or add it to something unsafe), the result will be still safe from HTML point of view.

I think Pelican would benefit from using Jinja2's `autoescape` by default, since it makes the generated content less susceptible to XSS'es.

avaris:
I am a bit puzzled by this. `truncate_html_words` is an internal function and not at all exposed to `Jinja` side of the things.

galaxy4public:
Well, in my theme I needed a quick summary generated by keeping just few words.  I just checked my config, and you are correct -- the function itself is not available to Jinja2 by default.  I have the following in my config to make it visible:
```
from pelican.utils import truncate_html_words
JINJA_FILTERS.update({'truncate_html_words': truncate_html_words})
```
I guess, then it is not a bug :).  However, having such a filter helped me to do truncation at many places, I will find some other way to achieve it then.

galaxy4public:
[...]

PR #2918, opened 2021-09-08
Adding missing tests for truncate_html_words()

galaxy4public:
Added support for Jinja2 generated end marker when the passed variable
is an instance of Markup.  This makes the function compatible with
Jinja2's environments where autoescape is enabled.

# Pull Request Checklist

Resolves: #2917 

### DIFF
--- a/pelican/tests/test_utils.py
+++ b/pelican/tests/test_utils.py
@@ -6,9 +6,9 @@ import time
 from sys import platform
 from tempfile import mkdtemp
 
-import pytz
-
 from markupsafe import Markup
+
+import pytz
 
 from pelican import utils
 from pelican.generators import TemplatePagesGenerator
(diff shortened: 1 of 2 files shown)


galaxy4public:
I am looking at the failed tests and I am puzzled re: how my change could have affected these three tests.  Moreover, when I run `tox` on a clean clone of the repository (without my changes in), I am also getting these three test failed.  At the same time, I see that the latest commit in Jul passed all the tests.  Could somebody help me understand what's going on, please?

justinmayer:
There are two CI failures. The first one is indeed unrelated to your changes and is instead caused by a newer version of FeedGenerator, for which we have not yet updated our functional test output.

The second one is due to code style conformance checks. Did you follow the [development docs](https://docs.getpelican.com/en/latest/contribute.html#setting-up-the-development-environment) and run `invoke lint` before committing changes? If you make sure you have Pre-commit installed, it should catch those problems even if you forget to run `invoke lint`.

justinmayer:
… and the more I look at the [diff](https://github.com/getpelican/pelican/pull/2918/checks?check_run_id=3546166870#step:11:194) caused by the latest version of FeedGenerator, the more I wonder whether the `\n` that got introduced into the feed should be there. 🤔

galaxy4public:
Yeah, I also found that FeedGenerator has been updated on Aug 18 and the diff I made shows that the difference is actually the missing `<subtitle></subtitle>` -- we are passing `None` as the value, so most likely it was optimised out.

galaxy4public:
@justinmayer , I did fix linting issues (there is a commit just before your comment).  I was relying on my Vim being good with flake8 rules, but it was not, so when I've seen the linting errors, I quickly amended the file :).

galaxy4public:
@avaris correctly pointed out that I am tapping into an internal function (`truncate_html_words()`), which is not exposed to Jinja2 as a filter by default.  I agree with that assessment, so I propose to close the corresponding issue (#2917 ).  However, I would love to preserve the additional tests introduced in this PR.  Will it be OK if I drop the change to `truncate_html_words()` and the test for it, then rename the PR to be "Adding missing tests for truncate_html_words()" instead?  The tests are covering a situation when a custom `end_text` was provided, e.g. through the `SUMMARY_END_SUFFIX` parameter in `pelicanconf.py`.

justinmayer:
@galaxy4public: That seems to me a reasonable path forward on this topic. Would you please proceed as you suggested above?

galaxy4public:
@justinmayer , will do that (most likely this coming weekend since I am currently smashed with work :( )

justinmayer:
Hi @galaxy4public. Just checking in regarding your pull request. Do you think you could find some time to move it forward so we can get it merged?

galaxy4public:
@justinmayer, sorry, I was overloaded with other work lately.  Should be able to work on this by the end of this week.

PR #2922, opened 2021-09-22
Update docs with example on injecting articles

iUnknwn:
Updated the plugin documentation to add a recipe for adding articles programmatically when pelican is running.

I struggled to understand how to dynamically add content to Pelican (for example, if I wanted to fetch content via an API rather than from Markdown). Once I got it working, I wanted to update the docs so it was a bit easier for the next person. Hopefully this is the preferred way to inject dynamic content (as far as I can tell - this seems like it would be the correct signal). 

# Pull Request Checklist

N/A - updating documentation only - no code changes.

### DIFF
--- a/docs/plugins.rst
+++ b/docs/plugins.rst
@@ -301,6 +301,43 @@ Here is a basic example of how to set up your own writer::
         signals.get_writer.connect(add_writer)
 
 
+Modifying Generators to Inject Content
+---------------------------------------
+
+You can programatically inject articles or pages using plugins. This can be 
+useful if you plan to fetch articles from an API, for example. 
+
+Here is a very simple example of how we can build a plugin that injects
+custom articles, using the article `article_generator_pretaxonomy` signal::
+
+    from pelican import signals
+    from pelican.contents import Article
+    from pelican.readers import BaseReader
+    import datetime
+
+    def addArticle(articleGenerator):
+        settings = articleGenerator.settings
+
+        # Author, category, and tags are objects, not strings, so they need to
+        # be handled using BaseReader's process_metadata.
+        baseReader = BaseReader(settings) 
+
+        content = "I am the body of an injected article!"
+
+        newArticle = Article(content, {
+            "title": "Injected Article!",
+            "date": datetime.datetime.now(),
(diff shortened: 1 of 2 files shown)


docs/plugins.rst
```diff
@@ -301,6 +301,43 @@ Here is a basic example of how to set up your own writer::
...
+        # be handled using BaseReader's process_metadata.
+        baseReader = BaseReader(settings) 
+
+        content = "I am the body of an injected article!"
+
+        newArticle = Article(content, {
+            "title": "Injected Article!",
+            "date": datetime.datetime.now(),
+            "category": baseReader.process_metadata('category', 'fromAPI'),
+            "tags": baseReader.process_metadata("tags", 'tagA, tagB')
```
avaris, line 331: Random switch between single-quoted and double-quoted strings is triggering me a bit :).

avaris:
I like the idea but if go down this route we can end up with countless little recipes here (plugins can do *a lot*). Maybe it's better to move this to its own document (like `Example plugin recipes`) instead?

iUnknwn:
I'm happy to pull out the plugin examples into another page, if that would be better. Also happy to normalize the quotes. 

I do think that more documentation on the plugins overall is a good thing, though, both in terms of explaining what plugins can do (the docs focus on building new readers and generators, but not interacting with them), and showing how extendable pelican can be (discovering I could use hugo/liquid-style short codes by extending markdown was insanely helpful, but I never would have realized that was possible from the docs alone).

iUnknwn:
Just to follow up, would you like me to separate out the plugin examples and additional docs onto a separate page?

avaris:
I think it's still manageable as is, so let's keep it in same page. If you can normalize the quotes, I'll approve.

iUnknwn:
Sounds good. I normalized the quotes, and adjusted the section title to be a bit more simple/direct.

avaris (approved):
Thanks :+1:

justinmayer:
Merged manually via 9ec17507. Many thanks to @iUnknwn for the documentation addition and to @avaris for reviewing. 🏅

PR #2927, opened 2021-10-02
Use rich logging for Pelican server

MinchinWeb:
Further to #2897, this enables *rich* logging for `pelican.server.

image

Surprisingly (at least to me), all that was needed was to import `console` from `pelican.log`.

Note this is working for `python -m pelican.server`, but doesn't seem to work on `pelican -l`. The latter seems to be drawing its logging from elsewhere.

P.S. If this could be marked "hacktoberfest-accepted" (so it counts towards my Hacktoberfest Pull Requests), that would be appreciated!

### DIFF
--- a/pelican/server.py
+++ b/pelican/server.py
@@ -12,6 +12,7 @@ try:
 except ImportError:
     magic_from_file = None
 
+from pelican.log import console  # noqa: F401
 from pelican.log import init as init_logging
 logger = logging.getLogger(__name__)
 
@@ -103,6 +104,9 @@ class ComplexHTTPRequestHandler(server.SimpleHTTPRequestHandler):
             mimetype = magic_from_file(path, mime=True)
 
         return mimetype
+
+    def log_message(self, format, *args):
+        logger.info(format, *args)
 
 
 class RootedHTTPServer(server.HTTPServer):
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -476,6 +476,10 @@ def autoreload(args, excqueue=None):
 
 
 def listen(server, port, output, excqueue=None):
+    # set logging level to at least "INFO" (so we can see the server requests)
+    if logger.level < logging.INFO:
+        logger.setLevel(logging.INFO)
+
     RootedHTTPServer.allow_reuse_address = True
     try:
         httpd = RootedHTTPServer(
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -487,8 +487,8 @@ def listen(server, port, output, excqueue=None):
         return
 
     try:
-        print("\nServing site at: http://{}:{} - Tap CTRL-C to stop".format(
-            server, port))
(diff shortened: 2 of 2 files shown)


avaris:
* Linter is not happy (unused import)
* `pelican -l` bits are [here](https://github.com/getpelican/pelican/blob/f862d64b7a492f2af24e1228790a1d75371a1112/pelican/__init__.py#L478)
* This is a very limited support for `server`. All the messages the server outputs are rather "poor". It can be "enriched" by overriding [`log_message`](https://docs.python.org/3/library/http.server.html#http.server.BaseHTTPRequestHandler.log_message) method in the [subclass](https://github.com/getpelican/pelican/blob/f714f27c787a61e488ce922777ae322de7746cf6/pelican/server.py#L42) and using logging to display the output.

MinchinWeb:
Thanks for the tips @avaris ; I knew initially it was somewhat limited. I've fixed the three items you pointed to!

image

Previously the logging provided both the local address and the time. Time is still there (although without the date) but I've dropped the local address as it is provided on starting and should never change; plus is keeps the log messages more on point.

Are we good to merge?

pelican/server.py
```diff
@@ -99,6 +100,9 @@ def guess_type(self, path):
 
         return mimetype
 
+    def log_message(self, format, *args):
+        logger.info("%s" % format % args)
```
avaris, line 104: Extra formatting is unnecessary. This should be enough:
```python
logger.info(format, *args)
```

MinchinWeb: Fixed! The previous code is what is used upstream.

MinchinWeb:
I've also set the minimum logging level to INFO, so that the server requests are actually shown.

avaris (approved):
Thanks

justinmayer:
@MinchinWeb: I don't have time today to check this in detail, but can you confirm that the output is not hard-wrapped? c.f., https://rich.readthedocs.io/en/latest/console.html#soft-wrapping

I ask because I think I noticed the other day that the new Rich-powered logging is now hard-wrapped, and if it is indeed now hard-wrapped, I think we should change that behavior as well.

justinmayer:
Disregard my last comment. The culprit was an outdated `tasks.py` file that did not have PTY set correctly.

Many thanks to @MinchinWeb for continuing to make Pelican's output richer and to @avaris for the review. 👏

justinmayer:
@MinchinWeb: One more thing... Could you squash the first four commits into one commit, leaving the "set minimum logging level to INFO" commit as a separate commit? That way, there should be two discrete commits in this PR once your squash is completed. (Normally I would do this manually outside the PR, but I want to ensure you get your Hacktoberfest credit.) Thank you!

justinmayer:
@MinchinWeb: Any chance you could squash four of the five commits here, as noted above? I'd like to publish a follow-up release today, and it'd be nice to include this while ensuring you get Hacktoberfest credit.

justinmayer:
Thanks again, @MinchinWeb!

PR #2928, opened 2021-10-02
Server: actually return MIME type

minchinweb:
The original paths being provided had a tailing slash, and so everything was returning as "application/octet-stream" (at least on Windows)

# Pull Request Checklist

Resolves: #2880 

Refer to that issue for the discussion.

- [x] Ensured **tests pass** and (if applicable) updated functional test output
- [x] Conformed to **code style guidelines** by running appropriate linting tools
- [ n/a ] Added **tests** for changed code
- [ n/a ] Updated **documentation** for changed code

**P.S.** If this could be marked "hacktoberfest-accepted" (so it counts towards my Hacktoberfest Pull Requests), that would be appreciated!

PR #2929, opened 2021-10-02
Add MIME types for web fonts

minchinweb:
Building on #2928, it seems Python doesn't recognize/return the proper MIME-type for webfonts. Firefox does seems to work fine either way (as opposed to CSS...), but this seemed simple and proper to fix.

Before:

After:

image

# Pull Request Checklist

**P.S.** If this could be marked "hacktoberfest-accepted" (so it counts towards my Hacktoberfest Pull Requests), that would be appreciated!

### DIFF
--- a/pelican/server.py
+++ b/pelican/server.py
@@ -43,6 +43,21 @@ def parse_arguments():
 class ComplexHTTPRequestHandler(server.SimpleHTTPRequestHandler):
     SUFFIXES = ['.html', '/index.html', '/', '']
 
+    extensions_map = _encodings_map_default = {
+        # included in Python default implementation
+        '.gz': 'application/gzip',
+        '.Z': 'application/octet-stream',
+        '.bz2': 'application/x-bzip2',
+        '.xz': 'application/x-xz',
+
+        # web fonts
+        ".oft": "font/oft",
+        ".sfnt": "font/sfnt",
+        ".ttf": "font/ttf",
+        ".woff": "font/woff",
+        ".woff2": "font/woff2",
+    }
+
     def translate_path(self, path):
         # abandon query parameters
         path = path.split('?', 1)[0]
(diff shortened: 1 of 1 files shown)


MinchinWeb:
This would resolved by https://github.com/python/cpython/pull/20199, when it is merged into the Python core.

avaris:
Well, either commits from another  PR**should not** be here or these commit should go on to that PR instead. Now you're guaranteed to cause merge conflicts.

pelican/server.py
```diff
@@ -43,6 +43,21 @@ def parse_arguments():
...
+        '.xz': 'application/x-xz',
+
+        # web fonts
+        ".oft": "font/oft",
+        ".sfnt": "font/sfnt",
+        ".ttf": "font/ttf",
+        ".woff": "font/woff",
+        ".woff2": "font/woff2",
+    }
+
```
avaris, line 60: I think it will be more future-proof if we extend over the original map instead of overriding it:

```python
    extensions_map = {
        **server.SimpleHTTPRequestHandler.extensions_map,
        **{
            # web fonts
            ".oft": "font/oft",
            ".sfnt": "font/sfnt",
            ".ttf": "font/ttf",
            ".woff": "font/woff",
            ".woff2": "font/woff2",
        }
    }
```

I looked at the CPython code and as far as I can see `_encodings_map_default`[*] serves no purpose at all. Seems like `extensions_map` is expected to be modified and that just keeps initial values as backup if it was needed afterwards. And looks like it wasn't needed. So I guess we should leave that one alone.

[*]: Also, looks like it was a typo and should be `_extensions_map_default` instead in the CPython code (?)

avaris (approved):

justinmayer:
Thanks to @MinchinWeb for the enhancement and to @avaris for reviewing.

Issue #2841, opened 2021-01-07
drafts index page renders with incorrect links to draft articles if it doesnt have a trailing /


johnmathews:
- [x] I have read the [Filing Issues](https://docs.getpelican.com/en/latest/contribute.html#filing-issues) and subsequent “How to Get Help” sections of the documentation.
- [x] I have searched the [issues](https://github.com/getpelican/pelican/issues?q=is%3Aissue) (including closed ones) and believe that this is not a duplicate.

- **OS version and name**: MacOS 11.0.1
- **Python version**: 3.8.5
- **Pelican version**: 4.5.3
- **Link to theme**: my own theme
- **Links to plugins**: 
nb_markup,
'neighbors',
'pelican.plugins.render_math',
'pelican.plugins.sitemap',
'pelican.plugins.simple_footnotes',
'pelican.plugins.more_categories',
- **Link to your site**: [johnmathews.eu](https://johnmathews.eu)
- **Link to [Gist](https://gist.github.com/johnmathews/703ac149ba3bf4e53290514bbec7d561) of local settings file

## Issue
in my local settings file I have the following settings for viewing drafts (i believe this is the default)
``` python
DRAFT_SAVE_AS = 'drafts/{slug}.html'
DRAFT_URL = 'drafts/{slug}.html'
```
I launch a local server with: `pelican --listen --autoreload`

if I go to `localhost:8000/drafts` then I see the correct (complete) list of hyperlinks to each draft article, but the link urls are wrong. they are `localhost:8000/<article-slug>.html` but they should be `localhost:8000/drafts/<article-slug>.html`

If however, i goto `localhost:8000/drafts/` (note the trailing `/`) then the links are correct and clicking on them sends the browser to the article.

Notwithstanding the incorrect links, the drafts index page seems to be identical in both cases.

This seems like a bug to me, but perhaps I've done something weird that I haven't spotted?

avaris:
I'd check the relevant templates. Make sure they are in [this form](https://github.com/getpelican/pelican/blob/659c4cbf0f0cd823d5eb199220e5a087f167733d/pelican/themes/notmyidea/templates/index.html#L11): note the explicit `/` between `SITEURL` and content URL.

johnmathews:
I'm not using any templates for the drafts index view.

avaris:
[...]

PR #2931, opened 2021-10-04
Adjust suffix in server to allow redirection when needed

avaris:
Folders without index.html has to be redirected (`/foo` -> `/foo/`) for
directory listing to work properly. Skip '/' suffix if original path
does not have it so that base class can return a redirect.

# Pull Request Checklist

Resolves: #2841 

- N/A Updated **documentation** for changed code


justinmayer:
Many thanks for the fix, @avaris! 🏅

PR #2932, opened 2021-10-04
Remove shebang lines from generated pelicanconf.py and publishconf.py

BenSturmfels:
These configuration files don't need a #! line as they are not executed directly and not marked as executable. (In practise this doesn't cause any problems - it just came up in a Guix bug report because Guix transforms the #! lines.)

The "coding: utf-8" lines are also no longer required now that Pelican is Python 3 only.

### DIFF
--- a/pelican/tools/templates/pelicanconf.py.jinja2
+++ b/pelican/tools/templates/pelicanconf.py.jinja2
@@ -1,6 +1,3 @@
-#!/usr/bin/env python
-# -*- coding: utf-8 -*- #
-
 AUTHOR = {{author}}
 SITENAME = {{sitename}}
 SITEURL = ''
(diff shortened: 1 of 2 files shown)


BenSturmfels:
Here's the Guix bug report - not particularly interesting - just for context.

http://issues.guix.gnu.org/issue/27447

avaris:
I'm OK with this PR but I had serious trouble following the Guix thread :). `pelican-quickstart` didn't put the full path to python there, we have `#!/usr/bin/env python` (which is practically ubiquitous in python scripts). Why is it *our bug* that somehow shebang converted to something else *after* `pelican` is done with it?

avaris (approved):

BenSturmfels:
@avaris I think you can safely ignore the Guix thread, its automatic #! conversion just highlighted that these !# probably shouldn't exist. I'm definitely not claiming that's Pelican's fault. :)

justinmayer:
Many thanks to @BenSturmfels for the adjustment and to @avaris for reviewing.

Issue #2938, opened 2021-10-19
Booleans specified with  -e / --extra-settings do not evaluate to False


r4victor:
- [x] I have read the [Filing Issues](https://docs.getpelican.com/en/latest/contribute.html#filing-issues) and subsequent “How to Get Help” sections of the documentation.
- [x] I have searched the [issues](https://github.com/getpelican/pelican/issues?q=is%3Aissue) (including closed ones) and believe that this is not a duplicate.

- **OS version and name**: macOS Mojave
- **Python version**: 3.9.7
- **Pelican version**: 4.7.1
- **Link to theme**: default produced by `pelican-quickstart`
- **Links to plugins**: default produced by `pelican-quickstart`
- **Link to your site**: –
- **Link to your source**: https://github.com/r4victor/pelican/tree/boolean_overrides_demo/website
- **Link to a [Gist](https://gist.github.com/) with the contents of your settings file**: https://github.com/r4victor/pelican/blob/boolean_overrides_demo/website/pelicanconf.py

## Issue

1. Pick any boolean settings parameter that is set to `True` by default or in `pelicanconf.py`.
2. Try to override it to `False` with `-e / --extra-settings`.
3. Observe that parameter is still `True`.

For example, I set `DELETE_OUTPUT_DIRECTORY = True` in `pelicanconf.py` and then run:
```
pelican --debug -e DELETE_OUTPUT_DIRECTORY=false
```
I see how the old files in the `output` directory are deleted. Here's the debug log: https://github.com/r4victor/pelican/blob/boolean_overrides_demo/website/debug.txt

Setting `-e param=False` or `-e param=0` doesn't work as well. The only way is to set the parameter to an empty string like `-e param=`. This is definitely not what users are supposed to do according to the docs (https://docs.getpelican.com/en/latest/settings.html#settings).

It seems the problem was introduced when fixing issue #2789 and originates in the [`coerce_overrides()`](https://github.com/getpelican/pelican/blob/bb10d286a6ac2f283d332f7e80c5655b9997f827/pelican/settings.py#L664) function in `settings.py`:

```python
def coerce_overrides(overrides):
    if overrides is None:
        return {}
    coerced = {}
    types_to_cast = {int, str, bool}
    for k, v in overrides.items():
        if k not in DEFAULT_CONFIG:
            logger.warning('Override for unknown setting %s, ignoring', k)
            continue
        setting_type = type(DEFAULT_CONFIG[k])
        if setting_type not in types_to_cast:
[...]

PR #2939, opened 2021-10-19
Fix issue 2938

r4victor:
# Pull Request Checklist

Resolves: #2938 

- [x] Ensured **tests pass** and (if applicable) updated functional test output
- [x] Conformed to **code style guidelines** by running appropriate linting tools
- [x] Added **tests** for changed code
- [ ] Updated **documentation** for changed code

PR #2940, opened 2021-10-20
Fix parsing of settings passed to `--extra-settings`

r4victor:
# Pull Request Checklist

Resolves: #2938

It's a proper fix of #2938. The first proposed PR #2939 is only a partial solution.

Now `-e` / `--extra-settings` expect `KEY=VALUE` pairs where VALUE must always be in JSON notation. Harsh, but it should at least work correctly.

A side effect: I removed the `coerce_overrides()` function that warned when unknown setting are overridden. No more such warnings. I think it makes more sense to warn about unknown settings no matter how they are specified (flags or the settings file), and so it's better to be addressed in a separate issue.

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -9,11 +9,12 @@ line::
 If you used the ``pelican-quickstart`` command, your primary settings file will
 be named ``pelicanconf.py`` by default.
 
-You can also specify extra settings via ``-e`` / ``--extra-settings`` option
-flags, which will override default settings as well as any defined within
-settings files::
+You can also specify settings via ``-e`` / ``--extra-settings`` option
+flags. It will override default settings as well as any defined within the
+setting file. Note that values must follow JSON notation::
 
-    pelican content -e DELETE_OUTPUT_DIRECTORY=true
+    pelican content -e SITENAME='"A site"' READERS='{"html": null}' CACHE_CONTENT=true
+
 
 .. note::
 
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -1,4 +1,5 @@
 import argparse
+import json
 import logging
 import multiprocessing
 import os
@@ -24,7 +25,7 @@ from pelican.plugins import signals
 from pelican.plugins._utils import get_plugin_name, load_plugins
 from pelican.readers import Readers
 from pelican.server import ComplexHTTPRequestHandler, RootedHTTPServer
-from pelican.settings import coerce_overrides, read_settings
+from pelican.settings import read_settings
 from pelican.utils import (FileSystemWatcher, clean_output_dir, maybe_pluralize)
 from pelican.writers import Writer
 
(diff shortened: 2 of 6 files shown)


pelican/__init__.py
```diff
@@ -366,13 +380,13 @@ def parse_arguments(argv=None):
...
-                             'notation. (e.g., SETTING=none)',
+                             'override settings. VALUE must be in JSON notation: '
+                             'specify string values as SETTING=\'"some string"\'; '
+                             'booleans as SETTING=true or SETTING=false; '
+                             'None as SETTING=null.',
                         nargs='*',
-                        action=ParseDict
-                        )
+                        action=ParseOverrides,
+                        default={})
```
avaris, line 389: It doesn't matter much but technically the default should be an empty list (`[]`) because parser expects a list.

avaris (commented):
I have one very minor (stylistic) comment, otherwise looks good to me. Thanks!

pelican/__init__.py
r4victor: `add_argument`'s `default` is "The value produced if the argument is absent from the command line" (https://docs.python.org/3/library/argparse.html#the-add-argument-method). So it's not an input to `ParseOverrides` but the value of `args.overrides`.  Any empty iterable will work, but since `args.overrides` is a dict, the default should be an empty dict.

avaris: You're right, not sure why I was thinking it would be passed to `ParseOverrides`. Sorry.

avaris (approved):
Thanks :+1:

docs/settings.rst
```diff
@@ -9,11 +9,12 @@ line::
...
 be named ``pelicanconf.py`` by default.
 
-You can also specify extra settings via ``-e`` / ``--extra-settings`` option
-flags, which will override default settings as well as any defined within
-settings files::
+You can also specify settings via ``-e`` / ``--extra-settings`` option
+flags. It will override default settings as well as any defined within the
+setting file. Note that values must follow JSON notation::
+
+    pelican content -e SITENAME='"A site"' READERS='{"html": null}' DELETE_OUTPUT_DIRECTORY=true
```
justinmayer, line 16: I had imagined that each specified setting would need to be prefixed with `-e`, such as:
```
pelican content -e SITENAME='"A site"' -e READERS='{"html": null}' -e DELETE_OUTPUT_DIRECTORY=true
```

Is that not necessary with this implementation? If multiple `-e` prefixes are not necessary, can other option flags be added _after_ multiple `--extra-settings` parameters, such as the following?
```
pelican content -e SITENAME='"A site"' READERS='{"html": null}' -o /tmp/output -t my-theme --autoreload
```

justinmayer: Also, since Pelican already has a `-d` / `--delete-output-directory` option flag, perhaps we should replace `DELETE_OUTPUT_DIRECTORY=true` with another example of a settings override? It seems to me that particular example doesn't make much sense when there is a built-in option flag already available.

justinmayer (commented):
Many thanks for the fix, Victor. Would you let me know once you've taken a look at the two comments I made here?

docs/settings.rst
r4victor: > Is that not necessary with this implementation?

It's not necessary but it has nothing to do with this PR. This behaviour is determined by `argparse`'s `nargs='*'` (https://docs.python.org/3/library/argparse.html#nargs). 

> can other option flags be added after multiple --extra-settings parameters

They can. 

> perhaps we should replace DELETE_OUTPUT_DIRECTORY=true with another example of a settings override?

Agree. I replaced it with `CACHE_CONTENT=true`.

justinmayer: Got it. Many thanks for the explanation!

justinmayer:
@r4victor: It seems like a Typogrify-related test is failing. Any idea why that might be?

avaris:
Errors are related to the `docutils` update and should be fixed by #2949. For some reason, this is not using the current master, which is confusing to me.

justinmayer:
It is times like this that make me wonder whether we should add the [Automatic Rebase](https://github.com/marketplace/actions/automatic-rebase) action, which would allow maintainers to comment `/rebase` to quickly and easily rebase PRs.

justinmayer:
It seems GitHub just rendered my last comment moot by adding a button to [rebase PR branches](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-suggestions-to-update-pull-request-branches), which I just enabled for this repository and subsequently used on this pull request, so all lights are now green. ✅ 

Many thanks to @r4victor for the enhancement and to @avaris for reviewing. 🥇

PR #2944, opened 2021-11-02
chore: Simplify boolean `if` expression ✨

yezz123:
# Pull Request Checklist

Simplifies boolean if expressions by removing unnecessary explicit references to `True` or `False` states.

Rather than using an if-expression to evaluate a boolean, you can just use it directly. This is shorter and clearer, if expressions of this form take a while to mentally parse.


pelican/writers.py
```diff
@@ -83,13 +82,17 @@ def _add_item_to_the_feed(self, feed, item):
             unique_id=get_tag_uri(link, item.date),
             description=description,
             content=content,
-            categories=categories if categories else None,
+            categories=categories or None,
```
avaris, line 85: Personally, I think
```python
a if a else b
```
is easier to parse than
```python
a or b
```

pelican/plugins/_utils.py
```diff
@@ -53,28 +53,27 @@ def load_legacy_plugin(plugin, plugin_paths):
...
-        sys.modules[spec.name] = mod
+    # Avoid loading the same plugin twice
+    if spec.name in sys.modules:
+        return sys.modules[spec.name]
+    # create module object from spec
+    mod = importlib.util.module_from_spec(spec)
+    # place it into sys.modules cache
+    # necessary if module imports itself at some point (e.g. packages)
+    sys.modules[spec.name] = mod
+    try:
```
avaris, line 64: While correct, this can be debatable. Sometimes it's better to explicitly (and visually) separate two branches. For example here with the changes `raise` block kind of blends in with the rest of the code. With an explicit `else` it is more clear that `raise` is also one of the "main" endpoints of the function.

pelican/writers.py
yezz123: Somehow if we want to make the shortcode to keep it clean and understood.

pelican/plugins/_utils.py
ingwinlu: @yezz123 would you be willing to remove the mentioned part from the pr so we can get this merged?

justinmayer: @yezz123: Hi Yasser. Last call… Would you please remove this specific change from your PR so we can get it merged?

yezz123: Hello @justinmayer @ingwinlu Sorry for the late reply i just fix it

pelican/plugins/_utils.py
```diff
@@ -122,4 +122,4 @@ def get_plugin_name(plugin):
     if inspect.ismodule(plugin):
         return plugin.__name__
 
-    return type(plugin).__qualname__
+    return type(plugin).__qualname__
```
justinmayer, line 125: Last commit seems to have removed an EOF marker for some reason. Could you fix this?

justinmayer (approved):
Looks good to me.

@avaris / @ingwinlu: Any further comments on this?

boxydog:
@avaris - This is a small PR, but I don't see bugs. It passes tests. Merge?

avaris (approved):
Good to go, but probably with a squash merge. Not the best commit messages...

justinmayer (approved):
Many thanks to @yezz123 for the enhancements and to @boxydog and @avaris for reviewing. 🎖️

PR #2949, opened 2021-11-25
Use docutils.Node.findall instead of traverse

avaris:
`docutils.Node.traverse` is being deprecated as of `docutils==0.18.1`.

... and causing test errors.

# Pull Request Checklist


justinmayer:
Thanks for the fix, @avaris!

PR #2976, opened 2022-01-17
Simple theme fixes

pieqq:
A set of changes for the `simple` theme to be mobile-friendly and to use `h1` consistently throughout the template files. Please check individual commit messages for more info.

### DIFF
--- a/pelican/themes/simple/templates/article.html
+++ b/pelican/themes/simple/templates/article.html
@@ -22,7 +22,6 @@
 {% endblock %}
 
 {% block content %}
-<section id="content" class="body">
   <header>
     <h1 class="entry-title">
       <a href="{{ SITEURL }}/{{ article.url }}" rel="bookmark"
@@ -60,8 +59,7 @@
     </div>
     {% endif %}
   </footer><!-- /.post-info -->
-  <div class="entry-content">
+  <article>
     {{ article.content }}
-  </div><!-- /.entry-content -->
-</section>
+  </article>
 {% endblock %}
--- a/pelican/themes/simple/templates/article.html
+++ b/pelican/themes/simple/templates/article.html
@@ -24,9 +24,9 @@
 {% block content %}
 <section id="content" class="body">
   <header>
-    <h2 class="entry-title">
+    <h1 class="entry-title">
       <a href="{{ SITEURL }}/{{ article.url }}" rel="bookmark"
-         title="Permalink to {{ article.title|striptags }}">{{ article.title }}</a></h2>
+         title="Permalink to {{ article.title|striptags }}">{{ article.title }}</a></h1>
  {% import 'translations.html' as translations with context %}
  {{ translations.translations_for(article) }}
   </header>
--- a/pelican/themes/simple/templates/base.html
+++ b/pelican/themes/simple/templates/base.html
@@ -33,11 +33,11 @@
         {% endblock head %}
 </head>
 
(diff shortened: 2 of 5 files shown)


pauloxnet:
LGTM.
Do you think there would be other html5 syntax to integrate into the 'simple' template?

pieqq:
I think it's already using a lot of HTML5 tags (`header, footer, section, time` to name a few).

One comment I have is that the templates are using a lot of `id` and `class` that I don't necessarily find very useful by default (especially since the simple theme comes with no CSS). For instance, in `base.html`:

```html
<body id="index" class="home">
        <header id="banner" class="body">
(...)
        <footer id="contentinfo" class="body">
```

I don't really see the point of the `body` class, nor the `index` id. But I'm not sure if I should go ahead and change this, since I don't know how this theme is being used by others.

pauloxnet:
I agree with your perplexity about the presence of a lot of `id` and `class` since there is no CSS in simple theme.

pieqq:
So do you think I should rework this as well? I can trim **a lot** of things, since a lot of tags can be targeted by a specific CSS rule without having to rely on a specific `id` or `class`...

pauloxnet:
Yes, from my point of view, but I think it would be better to seek the advice of other core developers

pieqq:
OK, then I'll wait for others to comment on this and take action, if any :)

pieqq:
I went ahead and made some modifications to the `base.html` template. I think the rest is fine.

pauloxnet (approved):

pauloxnet:
LGTM

pieqq:
I've rebased this branch onto master. Not sure if this can be merged now?

pelican/themes/simple/templates/article.html
```diff
@@ -22,11 +22,10 @@
 {% endblock %}
 
 {% block content %}
-<section id="content" class="body">
   <header>
-    <h2 class="entry-title">
+    <h1 class="entry-title">
```
pauloxnet, line 26: If we're moving simple theme to a classless theme we can remove also this class and all the others.

pieqq: I decided to leave this class because in the `base.html` also has an `<h1>` tag within a `<header>` tag, so I thought it would still be easier to target the entry title using a dedicated class in CSS. What do you think?

pelican/themes/simple/templates/base.html
```diff
@@ -32,11 +33,11 @@
...
 
-<body id="index" class="home">
-        <header id="banner" class="body">
+<body>
+        <header>
                 <h1><a href="{{ SITEURL }}/">{{ SITENAME }}{% if SITESUBTITLE %} <strong>{{ SITESUBTITLE }}</strong>{% endif %}</a></h1>
-        </header><!-- /#banner -->
-        <nav id="menu"><ul>
+        </header>
+        <nav><ul>
```
cpitclaudel, line 40: Changes like this one will break websites that extend the `Simple` theme; what was the motivation?

cpitclaudel:
> I don't really see the point of the body class, nor the index id. 

Many websites extend the Simple theme and add their own CSS; removing these will break them.  Was anything gained in the removal?

justinmayer:
@cpitclaudel: Avoiding disruption for existing users is a priority for us. Rather than speaking in theoretical terms, could you point us to specific sites that have been affected by this change?

PR #3000, opened 2022-04-29
Fix small typo

canyon289:
Fixing a very small typo. I really enjoy using pelican, thanks much for continuing to maintain it


justinmayer:
Many thanks for the fix and the kind words, Ravin. Much appreciated! 😊

PR #3011, opened 2022-06-13
Fix and update pre-commit hooks

antonmosich:
flake8 is no longer in the pre-commit repo, and now has its own repository
I also ran `pre-commit autoupdate` which failed before to update all hooks to the latest version.
# Pull Request Checklist

---
Most of those are not really applicable in my case...


justinmayer:
Many thanks for the enhancement, Anton. I want to keep Flake8 pinned at < 4.0 due to a [dependency resolution situation](https://github.com/PyCQA/flake8/pull/1438) that may not be resolved for a while, so I've adjusted things here accordingly. Thanks again!

PR #3025, opened 2022-08-16
Docs: Hidden posts also excluded from author index

Kristinita:
### 1. Summary

I added, that `Status: hidden` exclude posts from `author` indexing too.

### 2. Argumentation

Specifying `tag` and `category`, but not specifying `author`, may confuse users. It’s not clear from the documentation what effect `Status: hidden` has on `author`. In my opinion, the documentation should give complete and clear information about the software product.

Thanks.


justinmayer (approved):
Many thanks for the documentation improvement, @Kristinita!

PR #3040, opened 2022-09-12
Reduce commands of makefile

Saphyel:
The makefile has some duplicate commands that could be removed if we use some defaults.

I also was thinking that probably with the conf.py file some of the argument should be optional

### DIFF
--- a/pelican/tools/templates/Makefile.jinja2
+++ b/pelican/tools/templates/Makefile.jinja2
@@ -66,10 +66,8 @@ help:
 	@echo '   make clean                          remove the generated files         '
 	@echo '   make regenerate                     regenerate files upon modification '
 	@echo '   make publish                        generate using production settings '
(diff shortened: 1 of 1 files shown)


justinmayer:
Hi Carlos. I get the impression that this isn't exactly a neutral change. One set of commands uses `localhost` by default, while the "global" variants use `0.0.0.0`, which can be accessible outside one's local workstation and which is either a benefit or a hazard, depending on one's perspective. Or I'm just tired and perhaps missing something obvious.

Put another way, if these changes were merged, how would someone who prefers to serve on `localhost` use the Make automation to do so?

Saphyel:
One option would be using the env var `SERVER` on the makefile

Or use the conf option ["BIND"](https://docs.getpelican.com/en/latest/settings.html ) so in this you can rerun the same command many times and if you want to change the conf file or edit that file (similar concept as using the .env file)

PR #3051, opened 2022-10-20
GitHub actions update

avaris:
* GitHub is [deprecating](https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/) `node12` actions in favor of `node16`, so updated the actions with the latest versions. There is only `pandoc` installer still on `node12` and throws warnings but that should be resolved soon within `@v2`.
* `set-output` is also [deprecated](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/), but they added a proper check for step results in the [context](https://docs.github.com/en/actions/learn-github-actions/contexts#steps-context), so that ended up simplifying deploy logic (yay!).
* I don't know what happened but there is a weird dependency hell going on with `flake8` and `importlib-metadata`. However, this only affects `python<3.8` (`flake8` uses built-in for later versions), I upgraded the `Lint` job python to 3.9. I think we can unpin `flake8` version but I couldn't remember why it was pinned so I left it alone.
* While I was at it, I also moved the `Build docs` and `Deploy` pythons to 3.9 as well, since 3.7 started to become ancient-ish and will be EOLed in less than a year.

# Pull Request Checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -29,13 +29,13 @@ jobs:
             python: "3.7"
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: Setup Python ${{ matrix.config.python }}
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.config.python }}
       - name: Set up Pip cache (Linux)
-        uses: actions/cache@v2
+        uses: actions/cache@v3
         if: startsWith(runner.os, 'Linux')
         with:
           path: ~/.cache/pip
@@ -43,7 +43,7 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-pip-
       - name: Set up Pip cache (macOS)
-        uses: actions/cache@v2
+        uses: actions/cache@v3
         if: startsWith(runner.os, 'macOS')
         with:
           path: ~/Library/Caches/pip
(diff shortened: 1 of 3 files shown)


justinmayer:
Many thanks for these improvements. As for `flake8`, for the last year the entire Python community has been dealing with the repercussions of their decision to pin `importlib-metadata` to < 4.3: https://github.com/PyCQA/flake8/pull/1438

So the easiest way to resolve the resulting dependency resolution hell you accurately described has been to pin `flake8` to < 4.0. That said, I'm open to other remediation methods.

avaris:
Yeah, I did catch up on the backstory a bit and it was more than I anticipated. Using `flake8` which did not pin `importlib-metadata` does not resolve anymore. The recent release of `importlib-metadata==5.0.0` removed a bunch of APIs and makes `flake8` crash instead.

I added another commit for resolution. Problem comes from `flake8` and `markdown` clashing in dependencies. And `markdown` comes with installing `pelican`, however it's not required for the `flake8` check. So, added `skip_install=true` to `tox` and unpinned `flake8`.

justinmayer (approved):
Yes, I tried to [broker the peace on the Python-Markdown side](https://github.com/Python-Markdown/markdown/issues/1200), but since that didn't pan out, the pin was the most expedient resolution. I'm glad you found a better solution. 👍

PR #3054, opened 2022-10-23
Use `(?P=)` to replace `\2` for intrasite link

renyuneyun:
# Pull Request Checklist

# Description

This PR converts a normal (Perl-style?) regex to refer back to a matching group `\2` to the Python-specific extended form `(?P=quote)`.
This PR does this change because this whole regex already uses Python-specific extensions, so it is natural to make them consistent, and it is easier to read as well (of course, requiring understanding the syntax).

(I just recently learned this syntax. I can be totally wrong on understanding the intention.)

### DIFF
--- a/pelican/contents.py
+++ b/pelican/contents.py
@@ -341,7 +341,7 @@ class Content:
 
             (?P<quote>["\'])      # require value to be quoted
             (?P<path>{}(?P<value>.*?))  # the url value
-            \2""".format(intrasite_link_regex)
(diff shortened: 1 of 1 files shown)


avaris (approved):
+1 on the readability.

BTW, it's more like positional vs named backreferences rather than Perl vs Python :).

justinmayer:
Many thanks to @renyuneyun for the enhancement and to @avaris for reviewing 🥇

PR #3055, opened 2022-10-26
Add Python 3.11 to CI and Tox

avaris:
* adds python 3.11 to CI and tox
* `setup-python` action added `pip` cache functionality with considerably less configuration. changed the workflow to use that.
* updates one failing test that was using `locale.getdefaultlocale()` which is deprecated in 3.11. on the plus side, new test should work on windows and no need to skip

<hr>


justinmayer (approved):
Love the simplified configuration. Bravo! 👏

Issue #2401, opened 2018-08-20
How about translate the Pelican documents into Chinese?


ertuil:
I am a programmer in China, and I have great interest in Pelican. I am willing to translate the documents into Chinese to let more people know this project for such work has never been done yet. Of course, I will obey the LICENSE of this project.
I have no idea whether I can finish this job but I think I will try my best if allowed. I request your permission now and if you share the same interest, please let me know.
Thank you very much.

jorgesumle:
I wanted to translate it into Esperanto as well. I managed to generate a POT catalog and translated some strings (see [relevant documentation from Sphinx](www.sphinx-doc.org/en/stable/intl.html?highlight=translation#translating-with-sphinx-intl) to learn how to do that), but it's a long work and it has a low priority for me right now.

I'd personally translate Pelican itself before working on the translation of the documentation, because some output is produced in English. We should add i18n support to Pelican first. If you need some help setting up the i18n support for the Sphinx documentation or Pelican, just ask: I'd be happy to help you.

andytt:
@jorgesumle I want to learn more about your work, but some how I can not open the link, it returns a 404 error.

jorgesumle:
Sorry for that. Here you are: http://www.sphinx-doc.org/en/stable/intl.html?highlight=translation#translating-with-sphinx-intl

justinmayer:
Hi @ertuil. Very sorry to have taken so long to respond to your generous offer. Would you still be willing to translate Pelican's documentation into Chinese?

ertuil:
Sure. But I am busy these days, perhaps I can do it soon later.

PR #3084, opened 2023-01-03
WIP: translate the documentation into Spanish

jorgesumle:
I'm translating  the documentation into Spanish. I think that translating changelog is not worth the effort, so I'm going to leave that out.

I'm using [this guide](https://docs.readthedocs.io/en/stable/guides/manage-translations-sphinx.html).

Related issue #2401


KrlitosForever:
@jorgesumle Conoces la razón de por qué nunca se incluyo? también estoy interesado en traducir.

Issue #1463, opened 2014-09-14
Copying themes should set correct permissions for the local "output" directory.


teh:
I have a theme in a location where its permissions are `-r--r--r--`. 

If I execute pelican a second time I get:

```
CRITICAL: [Errno 13] Permission denied: u'[redacted]/output/theme/css/reset.css'
```

Reason:

```
$ ls -l output/theme/css/reset.css 
-r--r--r-- 1 tom tom 1.1K Jan  1  1970 output/theme/css/reset.css
```

bjornfor:
This "permission denied" issue also affect users of the Nix package manager (http://nixos.org/nix/), because nix makes packages read-only after installation. Even worse, DELETE_OUTPUT_DIRECTORY = True doesn't work here, because the _directories_ are also read-only:

```
$ pelican content
Done: Processed 1 article(s), 0 draft(s) and 0 page(s) in 0.10 seconds.

$ pelican content
ERROR: Unable to delete directory /home/bfo/pelican-test/output/theme; [Errno 13] Permission denied: '/home/bfo/pelican-test/output/theme/images/icons/rss.png'
CRITICAL: [Errno 13] Permission denied: u'/home/bfo/pelican-test/output/theme/images/icons/rss.png'

$ find output/ -type d -exec chmod 755 '{}' \;

$ pelican content
Done: Processed 1 article(s), 0 draft(s) and 0 page(s) in 0.10 seconds.
```

So an explicit chmod 755 was needed on the directories below output/ to make DELETE_OUTPUT_DIRECTORY = True work.

madduck:
This is still an issue and I was wondering if there was any update or plans to fix this?

justinmayer:
@madduck: As far as I'm aware, not many folks run into this issue since not many folks have such restrictive permissions. Ergo, there hasn't been much movement on this topic. What specific problem are you running into?

madduck:
Well, I have `umask 077`, which means that files get checked out by Git as 0600, and then those files are copied with permissions preserved, hence not readable by the web server. I have to install a hack into my build scripts to cope with this. It's just a bit weird because `pelican` takes care of permissions for all other files just fine.

bjornfor:
And users installing Pelican from the Nix package manager are affected by this, as mentioned earlier in this thread, since all files installed by Nix are made immutable.

justinmayer:
[...]

PR #3101, opened 2023-02-17
Don't copy file file ownership, permissions and metadata

SuperSandro2000:
When building the default template with pelican installed through a distro package manager, the copied files are usually owned by root and likely have 644 or 444 set which aborts the build. Ownership and permissions should be based on the destination directory and umask, not on the referenced files.

# Pull Request Checklist

Resolves: #1463

### DIFF
--- a/pelican/utils.py
+++ b/pelican/utils.py
@@ -300,7 +300,7 @@ def copy(source, destination, ignores=None):
             logger.info('Creating directory %s', dst_dir)
             os.makedirs(dst_dir)
         logger.info('Copying %s to %s', source_, destination_)
-        copy_file_metadata(source_, destination_)
+        copy_file(source_, destination_)
 
     elif os.path.isdir(source_):
         if not os.path.exists(destination_):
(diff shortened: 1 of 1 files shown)


justinmayer:
Looks like tests are failing because the test environment is installing Pygments 2.14.0 instead of the version (`2.8.1`) we've pinned in `requirements/test.pip`. Perhaps I'll look into updating our test output as well as the pinned Pygments version.

Aside from that, it seems sensible to remove what looks to be an outdated function that worked around a bug that presumably no longer exists. How does this look to you, @avaris?

pelican/utils.py
```diff
@@ -330,20 +330,17 @@ def walk_error(err):
...
-def copy_file_metadata(source, destination):
-    '''Copy a file and its metadata (perm bits, access times, ...)'''
-
-    # This function is a workaround for Android python copystat
-    # bug ([issue28141]) https://bugs.python.org/issue28141
+def copy_file(source, destination):
+    '''Copy a file'''
     try:
-        shutil.copy2(source, destination)
+        shutil.copyfile(source, destination)
```
avaris, line 343: I think the point of using `shutil.copy2` was to actually preserve metadata, specifically creation/modification times, to assist in HTML/Browser caching. Otherwise static files will be re-dated with each site generation and browsers will consider that these files are new and re-request them even if they were not changed. Permission transfer is an unfortunate side effect of `shutil.copystat`, but we can pick and choose what to copy if we do it ourselves. So this needs to be followed by a [`os.utime`](https://docs.python.org/3/library/os.html#os.utime) call to preserve that behavior.

justinmayer:
@SuperSandro2000: Would you be willing to amend this PR to account for the feedback provided by @avaris?

justinmayer:
Hi @SuperSandro2000. Last call… Would you be willing to amend this PR to account for the feedback provided above so we can get your contribution merged?

SuperSandro2000:
Sorry, I didn't find the time in the last months to do this and I don#t think I will in the next weeks.

justinmayer:
Thanks for helping with this, Sandro. Merged via #3228.

PR #3115, opened 2023-04-10
Use default configuration while importing

mart-e:

Before this commit:

```
$ pelican-import --wpfile -o ./content/ WordPress.xml -m markdown --wp-attach
[12:30:53] WARNING  Feeds generated without SITEURL set properly may not be valid                                                                                                                                                                                      log.py:91
           WARNING  No timezone information specified in the settings. Assuming your timezone is UTC for feed generation. Check https://docs.getpelican.com/en/latest/settings.html#TIMEZONE for more information 
...
```
This is because the code calls read_settings() but no configuration file is provided.

After this commit:
 
No more warning.

### DIFF
--- a/pelican/tools/pelican_import.py
+++ b/pelican/tools/pelican_import.py
@@ -15,7 +15,7 @@ from urllib.request import urlretrieve
 
 # because logging.setLoggerClass has to be called before logging.getLogger
 from pelican.log import init
-from pelican.settings import read_settings
+from pelican.settings import DEFAULT_CONFIG
 from pelican.utils import SafeDatetime, slugify
 
 
@@ -285,8 +285,7 @@ def dc2fields(file):
 
     print("%i posts read." % len(posts))
 
-    settings = read_settings()
-    subs = settings['SLUG_REGEX_SUBSTITUTIONS']
+    subs = DEFAULT_CONFIG['SLUG_REGEX_SUBSTITUTIONS']
     for post in posts:
         fields = post.split('","')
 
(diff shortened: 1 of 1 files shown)


avaris:
I'm not sure people will run `pelican-import` after they already have a config file. So this might be a half solution. Alternatively, instead of `read_settings` it just uses [`DEFAULT_CONFIG`](https://github.com/getpelican/pelican/blob/86f62d0a92ad78df36aac9f5837d4f2715535421/pelican/settings.py#L24). Current code already relies on the default value anyway.

PS: There are more places where `read_settings` is used in that file.

mart-e:
@avaris you are right, I wasn't sure about adding yet another option so this is much easier and most likely enough for the import. Thanks for the review :)

pelican/tools/pelican_import.py
```diff
@@ -285,7 +285,7 @@ def dc2fields(file):
 
     print("%i posts read." % len(posts))
 
-    settings = read_settings()
+    settings = DEFAULT_CONFIG
     subs = settings['SLUG_REGEX_SUBSTITUTIONS']
```
avaris, line 289: I kind of want to nitpick the unnecessity of the variable `settings`. This is the only time it is used:

```python
subs = DEFAULT_CONFIG['SLUG_REGEX_SUBSTITUTIONS']
```

avaris (approved):
LGTM and thanks!

justinmayer:
Many thanks to @mart-e for the contribution and to @avaris for reviewing. 🥇

PR #3119, opened 2023-04-16
Replace deprecated pkg_resources

justinmayer:
`importlib.metadata.version()` appears to be the anointed replacement. Not sure if what I've done here is appropriate or the best solution, so feedback is most welcome.

See also: [Single-sourcing the package version](https://packaging.python.org/en/latest/guides/single-sourcing-package-version/)

### DIFF
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -1,4 +1,5 @@
 import argparse
+import importlib.metadata
 import json
 import logging
 import multiprocessing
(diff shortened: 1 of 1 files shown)


avaris:
I think this is OK but `importlib-metadata` will be a requirement for `python-3.7`. Alternatively, we can drop support for 3.7 which is EOL in a few months ([June 2023](https://endoflife.date/python)) anyway.

justinmayer:
Thanks for having a look. My primary motivation was to address deprecation warnings when running tests, which obviously is not a high priority. I think we can leave this PR open until after Python 3.7 is officially EOL, at which point we can merge it.

PR #3127, opened 2023-05-06
Don't specify unlimited feed size by default

mart-e:
When not set, the RSS feed contains every article. For site with many articles, this makes a very long feed that can be painful to process.
Having a feed with hundreds of articles, is rarely expected and is probably not the best value.

Set a high fallback value of 100 so it does not change for small sites.

Still allow to have infinite feed by setting `FEED_MAX_ITEM = False`

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -998,10 +998,10 @@ the ``TAG_FEED_ATOM`` and ``TAG_FEED_RSS`` settings:
    placeholder. If not set, ``TAG_FEED_RSS`` is used both for save location and
    URL.
 
-.. data:: FEED_MAX_ITEMS
+.. data:: FEED_MAX_ITEMS = 100
 
-   Maximum number of items allowed in a feed. Feed item quantity is
-   unrestricted by default.
+   Maximum number of items allowed in a feed. Setting to ``None`` will cause the
+   feed to contains every article. 100 if not specified.
 
 .. data:: RSS_FEED_SUMMARY_ONLY = True
 
--- a/pelican/writers.py
+++ b/pelican/writers.py
@@ -143,11 +143,9 @@ class Writer:
 
         feed = self._create_new_feed(feed_type, feed_title, context)
 
-        max_items = len(elements)
-        if self.settings['FEED_MAX_ITEMS']:
-            max_items = min(self.settings['FEED_MAX_ITEMS'], max_items)
-        for i in range(max_items):
-            self._add_item_to_the_feed(feed, elements[i])
+        # FEED_MAX_ITEMS = None means [:None] to get every element
+        for element in elements[:self.settings['FEED_MAX_ITEMS']]:
+            self._add_item_to_the_feed(feed, element)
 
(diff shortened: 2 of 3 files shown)


docs/settings.rst
```diff
@@ -1000,8 +1000,8 @@ the ``TAG_FEED_ATOM`` and ``TAG_FEED_RSS`` settings:
 
 .. data:: FEED_MAX_ITEMS
```
avaris, line 1001: Default should be here with `= 100`.

See `RSS_FEED_SUMMARY_ONLY` below for an example.

docs/settings.rst
```diff
@@ -1000,8 +1000,8 @@ the ``TAG_FEED_ATOM`` and ``TAG_FEED_RSS`` settings:
...
 
-   Maximum number of items allowed in a feed. Feed item quantity is
-   unrestricted by default.
+   Maximum number of items allowed in a feed. `False` for unrestricted. 100 if
```
avaris, line 1003: Another similar option `SUMMARY_MAX_LENGTH` uses `None` for disabling the cutoff. So it should be `None` here as well for consistency.

pelican/writers.py
```diff
@@ -143,11 +143,15 @@ def write_feed(self, elements, context, path=None, url=None,
...
         if self.settings['FEED_MAX_ITEMS']:
-            max_items = min(self.settings['FEED_MAX_ITEMS'], max_items)
-        for i in range(max_items):
-            self._add_item_to_the_feed(feed, elements[i])
+            last_item = self.settings['FEED_MAX_ITEMS']
+        elif isinstance(self.settings['FEED_MAX_ITEMS'], bool):
+            # explicitly set FEED_MAX_ITEMS = False
+            last_item = len(elements)
+        else:
+            last_item = 100
```
avaris, line 152: Default should go into [`settings.py`](https://github.com/getpelican/pelican/blob/86f62d0a92ad78df36aac9f5837d4f2715535421/pelican/settings.py#L43)

pelican/writers.py
```diff
@@ -143,11 +143,15 @@ def write_feed(self, elements, context, path=None, url=None,
...
+        for element in elements[:last_item]:
+            self._add_item_to_the_feed(feed, element)
```
avaris, line 154: With disabling changed to `None` and given the fact that `some_list[:None]` will take all the items, logic here will be sooo much simpler :).

```python
for element in elements[:self.settings['FEED_MAX_ITEMS']]:
    self._add_item_to_the_feed(feed, element)
```

That's all of it.

avaris (changes requested):

pelican/writers.py
mart-e: Nice the `[:None]` shortcut :+1:

avaris (approved):
Looks good, thanks!

justinmayer (approved):
Thanks to @mart-e for the contribution and to @avaris & @boxydog for reviewing.

PR #3114, opened 2023-03-29
[FIX] pelican_import with gfm instead of markdown

mart-e:
The markdown import of pandoc is their own flavour of markdown. It for instance uses fenced divs[1] which are not supported by python-markdown.
When importing content from Wordpress, there is several issues as explained in discussion 3113[2]
This change follows a discussion with pandoc developer[3]

_To discuss : not sure which one is best between commonmark and gfm_ :man_shrugging: 

[1] https://pandoc.org/MANUAL.html#divs-and-spans
[2] https://github.com/getpelican/pelican/discussions/3113
[3] https://fosstodon.org/@pandoc/110105559949588768

Take the following Wordpress blog post sample:
```html
<p><!-- wp:paragraph --></p>
<p>Paragraph content</p>
<p><!-- /wp:paragraph --></p>
<p><!-- wp:image {"align":"center","id":3747,"sizeSlug":"full"} --></p>
<div class="wp-block-image">
<figure class="aligncenter size-full"><img src="https://test.com/test.jpg" alt="" class="wp-image-3747" title="Some title"/><br />
<figcaption><em>Some caption</em></figcaption>
</figure>
</div>
<p><!-- /wp:image --></p>
```
Before this commit:
was imported as 

```md
`<!-- wp:paragraph -->`{=html}

Paragraph content

`<!-- /wp:paragraph -->`{=html}

`<!-- wp:image {"align":"center","id":3747,"sizeSlug":"full"} -->`{=html}

::: wp-block-image
<figure class="aligncenter size-full">
<img src="https://test.com/test.jpg" title="Some title"
class="wp-image-3747" /><br />

<figcaption><em>Some caption</em></figcaption>
</figure>
:::

`<!-- /wp:image -->`{=html}
```

After this change:
```md
<!-- wp:paragraph -->

Paragraph content

<!-- /wp:paragraph -->

<!-- wp:image {"align":"center","id":3747,"sizeSlug":"full"} -->

<div class="wp-block-image">

<figure class="aligncenter size-full">
<img src="https://test.com/test.jpg" title="Some title"
class="wp-image-3747" /><br />

<figcaption><em>Some caption</em></figcaption>
</figure>

</div>

<!-- /wp:image -->
```

# Pull Request Checklist

Resolves: https://github.com/getpelican/pelican/discussions/3113

- [x] Ensured **tests pass** and (if applicable) updated functional test output
- [x] Conformed to **code style guidelines** by running appropriate linting tools
- [ ] Added **tests** for changed code
[...]

PR #3141, opened 2023-06-04
Convert Wordpress caption to figure

mart-e:
In Wordpress, inserting image with a caption can look like 
```
[caption id="attachment_42" caption="Image Description"]<a ...><img ... /></a>[/caption]
[caption id="attachment_42"]<a ...><img ... /></a> Image Description[/caption]
[caption id="attachment_42"]<img ... > Image Description[/caption]
```

Replace by an HTML `<figure>` tag

### Pull Request Checklist

PS: feel free to suggest a better regex, this was [tested with a few examples](https://regex101.com/r/oTyq6b/1) taken from my blog.

### DIFF
--- a/pelican/tests/content/wordpressexport.xml
+++ b/pelican/tests/content/wordpressexport.xml
@@ -685,7 +685,52 @@ proident, sunt in culpa qui officia deserunt mollit anim id est laborum.]]></con
             <wp:meta_key>_edit_last</wp:meta_key>
             <wp:meta_value><![CDATA[3]]></wp:meta_value>
         </wp:postmeta>
-    </item>   
+    </item>
+    <item>
+        <title>Caption on image</title>
+        <link>http://thisisa.test/?p=176</link>
+        <pubDate>Thu, 01 Jan 1970 00:00:00 +0000</pubDate>
+        <dc:creator>bob</dc:creator>
+        <guid isPermaLink="false">http://thisisa.test/?p=176</guid>
+        <description></description>
+        <content:encoded><![CDATA[Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod
+tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam,
+quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo
+consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse
+cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non
+proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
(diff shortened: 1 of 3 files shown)


mart-e:
Don't get why the build is failing... When raising an error to see ouput, it's as if 1b360acafaa126ee52ca8472e816ec3f994c7c02 was not merged :thinking: 
https://github.com/getpelican/pelican/actions/runs/5169977273/jobs/9312580636

justinmayer:
It seems clear that https://github.com/getpelican/pelican/commit/1b360acafaa126ee52ca8472e816ec3f994c7c02 was indeed merged. It appears in `master` as well as in the branch attached to this PR: https://github.com/mart-e/pelican/commits/wp-caption-to-figure

mart-e:
@justinmayer ok after a lot more debugging, I finally understood, it's due to the pandoc version used in github actions (2.9.2) being different than the one on my machine (3.0.1).

If you check the output in the action below, you see that
```html
<p><figure>
<img src="/theme/img/xpelican.png.pagespeed.ic.Rjep0025-y.png"/>
<figcaption>This is a pelican</figcaption>
</figure></p>
```
is converted using `pandoc -f html+raw_html --to=gfm-smart --wrap=none` and results in:
```md
![This is a pelican](/theme/img/xpelican.png.pagespeed.ic.Rjep0025-y.png)
```
https://github.com/mart-e/pelican/actions/runs/5174375780/jobs/9320618059#step:8:78

I may have been a bit optimistic on #3114 solving all issues.

Only pandoc 3.0 has a proper support for `<figure>` tags:
https://pandoc.org/releases.html#pandoc-3.0-2023-01-18
```
Markdown writer: figures are output as implicit figures if possible, via HTML if the raw_html extension is enabled, and as Div elements otherwise.
HTML reader: <figure> elements are parsed as figures, with the caption taken from the respective <figcaption> elements.
```

The easiest for my issue would be to upgrade the required vrsion of pandoc but it's probably not the best as even [the latest ubuntu](https://packages.ubuntu.com/lunar/pandoc) is still in 2.17.

I see two better solutions:

- modify my test to be a bit less strict and must check that `[caption]` tags are converted (to whatever pandoc allows)
- instead of using `<figure>...<figcaption>Caption Text</figcaption></figure>`, I convert to an `<p></p><p>Caption Text</p>`

I tend to prefer the first one but it brings a different result based on the version of pandoc installed.

Let me know if ok with you.

justinmayer:
I agree that the first option sounds better. 👍

justinmayer (approved):
Many thanks for your work on this, Martin. We can revisit this topic once Pandoc 3.0+ is ubiquitous, but until then your solution strikes a good balance. 👍

Issue #2773, opened 2020-06-22
Period Archive Quirk


bnice5000:
I am working on a Bootstrap 4 theme. I am working to create the `period_archives.html`. It generates the subpages in the posts structure just fine. However, in the base `period_archives.html` page, the `{{ period }}` jinja2 variable does not work. I have a work around. It is pretty sloppy and it requires the jinja2 extension `do`. I would prefer to have direct access to the `period` variable on the base `period_archives` page so I don't have to code around the issue.

I will say that it has been a while since I have messed with theming in Pelican so I may be approaching the `period_archives` page incorrectly. If there is a better way to approach this issue or a better work around, i would be happy to have some help. Thanks in advance!

My workaround, assuming you have the jinja2 `do` extension turned on looks like this:

```jinja2
{% if output_file == 'period_archives.html' %}
  {% set period = [] %}
  {% for year, date_year in dates|groupby( 'date.year' )|sort(reverse=True)  %}
    {% do period.append(year) %}
  {% endfor %}
{% endif %}
```

- [x] I have searched the [issues](https://github.com/getpelican/pelican/issues?q=is%3Aissue) (including closed ones) and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://docs.getpelican.com/) and believe that my question is not covered.

avaris:
I'm a bit confused. What is `period_archive.html`? The template name is [`period_archives.html`](https://github.com/getpelican/pelican/blob/master/pelican/themes/notmyidea/templates/period_archives.html) (note the `s` at the end)

bnice5000:
> I'm a bit confused. What is `period_archive.html`? The template name is [`period_archives.html`](https://github.com/getpelican/pelican/blob/master/pelican/themes/notmyidea/templates/period_archives.html) (note the `s` at the end)

I missed the trailing `s` in the page name in my description. Per the code, it is named correctly in the project. I have edited the question to add the `s`'s. Hopefully that eliminated any confusion you are having.

avaris:
Can you show an example with the issue? I cannot reproduce it with template from default theme (the one I linked earlier). That one uses `period` variable and produces output as expected.

```console
$ cat pelicanconf.py
MONTH_ARCHIVE_SAVE_AS = 'posts/{date:%Y}/{date:%b}/index.html'
$ pelican content
WARNING: Feeds generated without SITEURL set properly may not be valid
[...]

PR #3148, opened 2023-06-18
Implement period_archives common context variable

djramones:
# Motivation

As discussed in #2773 (specifically [this comment](https://github.com/getpelican/pelican/issues/2773#issuecomment-650610307)), a context variable that lists the generated period archives along with their URLs should be useful for those who'd like to have an index to those archives. Before this, users had to manually write links to period archives in content or templates, or even expect site visitors to explore editing URLs on their browser address bar.

# Implementation

I moved code from `ArticlesGenerator.generate_period_archives()` to `generate_context()`. Initially, the new context variable only needed the list of `period`/`period_num` tuples and the corresponding URLs, but in order to retain the `generate_period_archives()` functionality, I decided to include the generation of `save_as` and filtered `articles`/`dates` collections.

Also, I decided it appropriate to set default patterns for time-period `*_ARCHIVE_URL` settings, so that anyone using the new context var just by setting `*_ARCHIVE_SAVE_AS` would get sensible URLs by default too.

I thought about providing a default template implementing the new variable (perhaps named `period_archives_index.html`), but this is probably more of a website- or theme-level detail, so I just put in a sample HTML snippet in the docs.

# Pull Request Checklist

Resolves: #2773


pelican/settings.py
```diff
@@ -90,11 +90,11 @@ def load_source(name, path):
...
         (2, '{name}{number}{extension}', '{name}{number}{extension}'),
     ],
-    'YEAR_ARCHIVE_URL': '',
+    'YEAR_ARCHIVE_URL': 'posts/{date:%Y}/',
     'YEAR_ARCHIVE_SAVE_AS': '',
-    'MONTH_ARCHIVE_URL': '',
+    'MONTH_ARCHIVE_URL': 'posts/{date:%Y}/{date:%b}/',
     'MONTH_ARCHIVE_SAVE_AS': '',
-    'DAY_ARCHIVE_URL': '',
+    'DAY_ARCHIVE_URL': 'posts/{date:%Y}/{date:%b}/{date:%d}/',
```
avaris, line 97: Why the default values?

pelican/generators.py
```diff
@@ -680,6 +634,74 @@ def _process(arts):
         self.dates.sort(key=attrgetter('date'),
                         reverse=self.context['NEWEST_FIRST_ARCHIVES'])
 
+        def _build_period_archives(sorted_articles):
+            period_archives = defaultdict(list)
```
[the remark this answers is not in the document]
djramones: Agreed that it should be moved out of `generate_context`, but I'm not sure about `utils` as the destination, because its purpose is too specific to generators, specifically `ArticlesGenerator` even. How about making it a `@staticmethod` right after `generate_context`?

pelican/settings.py
djramones: Mainly just to provide sensible defaults (that were already suggested elsewhere in the settings docs before this PR) to help a bit with configuring new Pelican sites. (With this PR, if users choose to enable period archives, they also need to configure the URLs, unlike before when those settings were only used for `PAGINATION_PATTERNS`).

I can see now though that having defaults could be more confusing, especially since the `*_ARCHIVE_SAVE_AS` settings should match.

Should I revert to blank defaults and just edit the docs accordingly?

pelican/generators.py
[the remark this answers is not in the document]
avaris: Sure, it can be a method in `ArticleGenerator`, not necessarily a `@staticmethod` though. It can be an internal method and keep using the `self.settings` as is.

pelican/settings.py
avaris: I believe so. These defaults won't do anything without `*_SAVE_AS` being defined as well. I think it would be confusing.

pelican/generators.py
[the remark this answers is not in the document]
djramones: I've pushed the changes for this. Didn't make it a static method, though I did turn settings into a parameter.

pelican/settings.py
djramones: I've pushed a commit to revert to blank defaults.

offbyone (approved):
This looks good.

Should you add an entry to the changelog for this, or will @justinmayer do that on release?

pelican/generators.py
```diff
@@ -483,64 +484,17 @@ def generate_period_archives(self, write):
...
-            'month': self.settings['MONTH_ARCHIVE_URL'],
-            'day': self.settings['DAY_ARCHIVE_URL'],
-        }
-
-        period_date_key = {
-            'year': attrgetter('date.year'),
-            'month': attrgetter('date.year', 'date.month'),
-            'day': attrgetter('date.year', 'date.month', 'date.day')
-        }
+        for granularity in list(self.period_archives.keys()):
```
boxydog, line 487: nit: `for granularity in self.period_archives`

doesn't need keys or list.

pelican/tests/test_generators.py
```diff
@@ -405,6 +405,103 @@ def test_per_article_template(self):
...
+        settings['YEAR_ARCHIVE_SAVE_AS'] = 'posts/{date:%Y}/index.html'
+        settings['YEAR_ARCHIVE_URL'] = 'posts/{date:%Y}/'
+        context = get_context(settings)
+        generator = ArticlesGenerator(
+            context=context, settings=settings,
+            path=CONTENT_DIR, theme=settings['THEME'], output_path=None)
+        generator.generate_context()
+        period_archives = generator.context['period_archives']
+        self.assertEqual(len(period_archives.items()), 1)
+        self.assertIn('year', period_archives.keys())
```
boxydog, line 435: nit: I'd just self.assertEqual, i.e.

```
self.assertEqual(
  {'year': {'period': [1970, 2014]}},
  period_archives)
```

or whatever the right comparison is.

It tests more, and makes it clearer what period_archives looks like.

pelican/tests/test_generators.py
```diff
@@ -405,6 +405,103 @@ def test_per_article_template(self):
...
+            'posts/{date:%Y}/{date:%b}/index.html'
+        settings['MONTH_ARCHIVE_URL'] = \
+            'posts/{date:%Y}/{date:%b}/'
+        context = get_context(settings)
+        generator = ArticlesGenerator(
+            context=context, settings=settings,
+            path=CONTENT_DIR, theme=settings['THEME'], output_path=None)
+        generator.generate_context()
+        period_archives = generator.context['period_archives']
+        self.assertEqual(len(period_archives.items()), 2)
```
boxydog, line 451: Same here, I'd `self.assertEqual` the whole thing.

pelican/tests/test_generators.py
```diff
@@ -405,6 +405,103 @@ def test_per_article_template(self):
...
+            'posts/{date:%Y}/{date:%b}/{date:%d}/index.html'
+        settings['DAY_ARCHIVE_URL'] = \
+            'posts/{date:%Y}/{date:%b}/{date:%d}/'
+        context = get_context(settings)
+        generator = ArticlesGenerator(
+            context=context, settings=settings,
+            path=CONTENT_DIR, theme=settings['THEME'], output_path=None)
+        generator.generate_context()
+        period_archives = generator.context['period_archives']
+        self.assertEqual(len(period_archives.items()), 3)
```
boxydog, line 468: Same here, I'd self.assertEqual all of period_archives

boxydog (approved):
I left a few nits, but I don't see anything blocking.

Since I am new to the codebase, I could be missing something.

pelican/tests/test_generators.py
avaris: I can understand this test style. `period_archives` contains actual `Article` objects. Not really possible to test complete equality.

avaris: But I agree, so I changed them as close to that as possible :).

avaris (approved):

justinmayer:
@offbyone: Normally we handle change-log entries via individual `RELEASE.md` files that should be included with PRs, but since we are merging a bunch of PRs into a single Pelican 4.9 release, I will handle this particular change-log bundle manually. You can see my collection so far in: https://github.com/getpelican/pelican/issues/3219

justinmayer (approved):
Many thanks to @djramones for the enhancement and to @boxydog, @offbyone, and @avaris for reviewing. 💯

PR #3149, opened 2023-06-19
Strip HTML tags from SITENAME inside title tags. Fixes #3147

### LINKED ISSUE #3147: Strip html tags from the SITENAME and AUTHOR when used in the head

## Feature Request

Hi, Pelican is great!

I wanted to give style to specific words in the `SITENAME` and `AUTHOR`, that are usually displayed at the top of the generated home page, which means adding html tags in those variables. However, this is inappropriate in the metadata and in the `<title>` fields of the html page (because it shows the raw text).

Specifically, my SITENAME is something like: `firstname <span class="smallcaps">lastname</span>`.

Is there a way to tolerate html markup in these variables, and tell Pelican or the jinja2 template to strip it away when needed?

Thanks!

Gullumluvl:
Resolves: #3147 for the default themes `simple` and `notmyidea`.

### DIFF
--- a/pelican/RELEASE.md
+++ b/pelican/RELEASE.md
@@ -1,0 +1,3 @@
+Release type: patch
+
+strip html tags from SITENAME inside title tags (in default themes).
--- a/pelican/RELEASE.md
+++ b/pelican/RELEASE.md
@@ -1,3 +1,0 @@
-Release type: patch
-
-strip html tags from SITENAME inside title tags (in default themes).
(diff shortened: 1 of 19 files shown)


pelican/themes/notmyidea/templates/archives.html
```diff
@@ -1,7 +1,7 @@
 {% extends "base.html" %}
 {% block content %}
 <section id="content" class="body">
-<h1>Archives for {{ SITENAME }}</h1>
+<h1>Archives for {{ SITENAME|striptags }}</h1>
```
avaris, line 4: This is within content. I don't think this needs to remove tags.

avaris (approved):

PR #3151, opened 2023-06-21
Use `watchfiles` as a file watching backend

RealOrangeOne:
This doesn't use polling unless absolutely necessarily, making it more efficient. It also reduces the amount of first-party code required, and simplifies working out which files are being watched.

# Pull Request Checklist


avaris:
I'll review when I get the time but before that you also need to add it to the requirements list in `setup.py`.

pelican/utils.py
```diff
@@ -753,167 +755,31 @@ def order_content(content_list, order_by='slug'):
...
 
-        Returns a boolean indicating if files are changed since last check.
-        Returns None if there are no matching files in the folder'''
+    watching_paths.extend(
+        os.path.join(content_path, path) for path in settings.get('STATIC_PATHS', [])
+    )
 
-        def file_times(path):
-            '''Return `mtime` for each file in path'''
+    watching_paths = [os.path.abspath(p) for p in watching_paths if p and os.path.exists(p)]
```
avaris, line 774: Might still be a good idea to warn if watched path does not exist.

avaris (changes requested):
Also, linter wants some changes.

avaris (approved):
Had to make some changes. 
* For not-existing paths it was erroring out, so had to take them out of watch list
* Changed the order of `pelican.run()`, otherwise it was waiting awkwardly for a change before it run at the start.

justinmayer (approved):
Many thanks to @RealOrangeOne and @avaris for this enhancement! ✨

PR #3154, opened 2023-06-24
Fix grammar errors/typos from docs/faq.rst

jorgesumle:


justinmayer (approved):
Many thanks for these fixes, Jorge! 🥇

PR #3159, opened 2023-07-01
Allow to cap the pagination

mart-e:
This way one could only generate one page of content by using `MAX_PAGINATION = 1`

My request is similar to what was tried at #2548.
On [my site](https://mart-e.be/ ) I have an index page with 10 articles and then, a `/archives` listing *all* articles without any pagination. I still want to use 10 for `DEFAULT_PAGINATION` but stop after the first page. As [mentioned by avaris](https://github.com/getpelican/pelican/pull/2548#issuecomment-490971787), the easiest is to have a new parameter.

An alternative could be to have a boolean value like `PAGINATE = False` to disable the iteration but it gets a bit confusing when mixing with `DEFAULT_PAGINATION`...

## Pull Request Checklist

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -1038,6 +1038,11 @@ You can use the following settings to configure the pagination.
    The maximum number of articles to include on a page, not including orphans.
    False to disable pagination.
 
+.. data:: MAX_PAGINATION = False
+
+   The maximum number of pages generated. ``False`` to paginate all available
+   pages.
+
 .. data:: PAGINATED_TEMPLATES = {'index': None, 'tag': None, 'category': None, 'author': None}
 
    The templates to use pagination with, and the number of articles to include
--- a/pelican/paginator.py
+++ b/pelican/paginator.py
@@ -47,6 +47,8 @@ class Paginator:
         if self._num_pages is None:
             hits = max(1, self.count - self.orphans)
             self._num_pages = int(ceil(hits / (float(self.per_page) or 1)))
+        if self.settings["MAX_PAGINATION"]:
+            self._num_pages = min(self.settings["MAX_PAGINATION"], self._num_pages)
         return self._num_pages
     num_pages = property(_get_num_pages)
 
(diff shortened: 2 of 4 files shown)


mart-e:
but that still generates orphan index pages in the background. Or I missed a way to stop it?

avaris:
[Turn off pagination](https://docs.getpelican.com/en/latest/settings.html#pagination) for index:

```python
# define PAGINATED_TEMPLATES and remove index from the default, if you want to keep others paginated
PAGINATED_TEMPLATES = {'tag': None, 'category': None, 'author': None}
```

PS: I'll look through the PR when I get a chance, but you don't need this feature :).

mart-e:
Oh I see what you meant now! I can just turn off pagination with `DEFAULT_PAGINATION = None` (I don't need pagination anywhere actually) and modify my index template to use `{% for article in articles[:10] %}` to limit.
I can definitely live with that way of limiting the articles :)

I don't know if my PR makes sense for other people and if it's worth to add a config parameter if we have a workaround. Feel free to close this PR if you believe it's too niche.

docs/settings.rst
```diff
@@ -1035,6 +1035,11 @@ You can use the following settings to configure the pagination.
    The maximum number of articles to include on a page, not including orphans.
    False to disable pagination.
 
+.. data:: MAX_PAGINATION = False
+
+   The maximum number of pages generated, not including orphans.
+   False to paginate all available pages.
```
avaris, line 1041: Orphans? They don't affect this.

Also, `False` should be codified: ``` ``False`` ```

mart-e: I was copy-pasting the above line. The concept of orphan pages is a bit vague for me but I can remove the reference if better.

avaris: Let's say, you have 31 articles with pagination 10. You'd end up with 4 pages where the last page would only contain 1 article. Orphans decide how many articles should be in that last page minimum. If it ends up below that threshold, last two pages are combined. So instead of 4 pages (10+10+10+1), you can get 3 pages (10+10+11).

mart-e:
Closing as can use the above workaround https://github.com/getpelican/pelican/pull/3159#issuecomment-1631039506

Issue #2956, opened 2021-12-03
SEO-friendly backlinks


Avamander:
- [X] I have read the [Filing Issues](https://docs.getpelican.com/en/latest/contribute.html#filing-issues) and subsequent “How to Get Help” sections of the documentation.
- [X] I have searched the [issues](https://github.com/getpelican/pelican/issues?q=is%3Aissue) (including closed ones) and believe that this is not a duplicate.

- **Pelican version**: 4.7.1

## Issue

This is a generic issue about the default Pelican theme, not specific to any generation.

Currently a bunch of the template pages contain backlinks to the "getpelican.com" domain, this does negatively impact its ranking as it's very numerous (looks spammy, bought) and you have no control over who links to you (could be a spammy domain).

## Proposal

Add `rel="nofollow"` (or `rel="ugc"`) to the backlinks in all templates. Suggest theme authors to do the same.

Avamander:
I suspect it's still relevant.

mart-e:
Hi,

You are talking about the "Proudly powered by pelican" in the bottom of the default theme, right?

> this does negatively impact its ranking as it's very numerous (looks spammy, bought) and you have no control over who links to you (could be a spammy domain).

What makes you think so? It's actually quite common to have this in default theme of CMS (e.g. the "powered by wordpress" everywhere and many don't have the  `rel=nofollow`) so I am not sure it's that bad.

PR #3162, opened 2023-07-25
Add rel='nofollow' to all external hardcoded links in templates

willthong:
# Pull Request Checklist

Resolves: #2956 
 
The HTML tag (eg `<a href="https://getpelican.com">`) in our templates negatively impacts a Pelican-generated site's ranking for Search Engine Optimisation purposes. This can be solved by adding the Google-developed `rel="nofollow"` attribute to external links so that the website's owner does not have to provide link authority to the target. This change also had to be reflected in the tests so that they would pass.

### DIFF
--- a/pelican/tests/output/basic/a-markdown-powered-article.html
+++ b/pelican/tests/output/basic/a-markdown-powered-article.html
@@ -58,10 +58,10 @@
 
         <footer id="contentinfo" class="body">
                 <address id="about" class="vcard body">
-                Proudly powered by <a href="https://getpelican.com/">Pelican</a>, which takes great advantage of <a href="https://www.python.org/">Python</a>.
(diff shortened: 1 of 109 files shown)


justinmayer:
As a general practice, I think it's a good idea to always specify the “why” — the motivation behind the proposed changes — in the commit body/description. That way folks can better understand the rationale, without having to take guesses as to the proponent's state of mind. 😉 

I can take a guess as to why, but in the interest of clarity, would you amend the commit body and the PR description here with your thoughts regarding the motivation/rationale?

avaris (approved):

justinmayer (approved):
Many thanks to @willthong for the enhancement and to @avaris for reviewing. 🥇

PR #3165, opened 2023-07-27
Replace `pytz` dependency in tests

willthong:
# Pull Request Checklist

Resolves: #2958 (something like a further resolution because it now includes a resolution within the tests) 

External dependency `pytz` has already been removed from the main code, but this PR also removes it from the tests.
Why have I changed the test timezone from CET to Europe/London? The tests were written with the [W3 examples](http://www.w3.org/TR/NOTE-datetime) in mind, but not accounting for the summertime offset. As such, they originally assumed that datetime(year=1997, month=7, day=16, hour=19, minute=20, tzinfo=ZoneInfo("CET")) would evaluate to 1997-07-16T19:20+01:00. But it doesn't; in summertime (July), CET is actually +02:00 and an example of a +01:00 timezone would be "Europe/London". With the shift to the built in ZoneInfo module, the test therefore needs to be adapted.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,4 +1,0 @@
-Release type: patch
-
-External dependency `pytz` has already been removed from the main code, but this PR also
-removes it from the tests.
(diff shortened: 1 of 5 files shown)


avaris (approved):
Looks good to me, thanks!

Issue #3178, opened 2023-08-19
Import yields incorrect link for Tumblr image posts


boxydog:
- [x] I have read the [Filing Issues](https://docs.getpelican.com/en/latest/contribute.html#filing-issues) and subsequent “How to Get Help” sections of the documentation.
- [x] I have searched the [issues](https://github.com/getpelican/pelican/issues?q=is%3Aissue) (including closed ones) and believe that this is not a duplicate.

- **OS version and name**: OS X 12.6.7
- **Python version**: 3.11.4
- **Pelican version**: 4.8.0
- **Link to theme**: default
- **Links to plugins**: default
- **Link to your site**: n/a
- **Link to your source**: n/a
- **Link to a [Gist](https://gist.github.com/) with the contents of your settings file**: n/a

## Issue

Importing tumblr blog entry with an image produces an incorrect link.

For example, https://www.tumblr.com/boxydog/188887119699/a-nice-visualization-of-whats-been-getting produces the link all spread out one letter at a time (see below). I bet the python code is iterating over a string when it thinks it's iterating over a list.

First 50 lines of the .rst file:

```
$ head -50 boxydog.com/2019-11-07-a-nice-visualization-of-whats-been-getting.rst 
blog.bluecrossmn.com
####################
:date: 2019-11-07 15:26:40
:author: boxydog
:category: photo
:tags: economics
:slug: 2019-11-07-a-nice-visualization-of-whats-been-getting
:status: published

<

i

m

g

a

l

t

=

"

"

s

r

c

=

"

h

t

t

p

s

:
```

PR #3187, opened 2023-08-26
Fix 3178

kzisme:
# Pull Request Checklist

Resolves: #3178  

I wasn't able to get the standard unit tests working locally, but it's a very small change.  

I attempted to run the unit tests described [here](https://github.com/getpelican/pelican/wiki/Git-Tips#making-your-changes) but I think I was missing some sort of initial setup.  

I was able to re-produce this bug via the tumblr API (although I ran into another error when the tool was going through the specific blog in this issue - which was related to expecting a str literal but found a bool).  

[here](https://gist.github.com/kzisme/87d1592cd914be62306475b8678d9dd4) is the API response from tumblr for the specific post in question along with the command I ran to reproduce, image link, and blog link.

### DIFF
--- a/pelican/tools/pelican_import.py
+++ b/pelican/tools/pelican_import.py
@@ -468,9 +468,8 @@ def tumblr2fields(api_key, blogname):
                     fmtstr = '<img alt="%s" src="%s" />'
                 content = ''
                 for photo in post.get('photos'):
-                    content += '\n'.join(
-                        fmtstr % (photo.get('caption'),
-                                  photo.get('original_size').get('url')))
+                    content += fmtstr % (photo.get('caption'),
+                                  photo.get('original_size').get('url'))
                 content += '\n\n' + post.get('caption')
             elif type == 'quote':
                 if format == 'markdown':
--- a/pelican/tools/pelican_import.py
+++ b/pelican/tools/pelican_import.py
@@ -468,9 +468,9 @@ def tumblr2fields(api_key, blogname):
                     fmtstr = '<img alt="%s" src="%s" />'
                 content = ''
                 for photo in post.get('photos'):
-                    content += fmtstr % (photo.get('caption'),
-                                  photo.get('original_size').get('url'))
-                content += '\n\n' + post.get('caption')
+                    content += '\n'.join(fmtstr % (photo.get('caption'),
+                                                   photo.get('original_size').get('url'))
+                                        for photo in post.get('photos'))
             elif type == 'quote':
                 if format == 'markdown':
                     fmtstr = '\n\n&mdash; %s'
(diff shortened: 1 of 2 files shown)


pelican/tools/pelican_import.py
```diff
@@ -468,9 +468,8 @@ def get_tumblr_posts(api_key, blogname, offset=0):
                     fmtstr = '<img alt="%s" src="%s" />'
                 content = ''
                 for photo in post.get('photos'):
-                    content += '\n'.join(
-                        fmtstr % (photo.get('caption'),
-                                  photo.get('original_size').get('url')))
+                    content += fmtstr % (photo.get('caption'),
+                                  photo.get('original_size').get('url'))
```
avaris, line 472: I think `'\n'.join(...)` was to keep one image per line if there are multiple and I'd say it's better to keep it in place for nicer output. But the implementation was wrong. So, the whole thing, including `for`, probably should be something like:

```python
content += '\n'.join(fmtstr % (photo.get('caption'),
                               photo.get('original_size').get('url'))
                     for photo in post.get('photos'))
```

PS: Also, you need to lineup your indentation properly to make linter happy.

kzisme: Ahh, that makes sense - I only tested with one image.  Sorry for the slight delay in getting back to this!

I had/have Pylint setup on my end - is there a different linter that is used for Pelican?  If so, can you point me in the direction to get it working properly?  I think I resolved everything in your comment, and have it indented properly, but let me know if there is anything else glaring.

justinmayer: At the moment, Pelican uses Flake8. You can see the two linter failures in the CI output: https://github.com/getpelican/pelican/pull/3187/checks#step:5:20

The dev environment setup docs are here: https://docs.getpelican.com/en/latest/contribute.html#setting-up-the-development-environment

pelican/tools/pelican_import.py
```diff
@@ -468,10 +468,9 @@ def get_tumblr_posts(api_key, blogname, offset=0):
                     fmtstr = '<img alt="%s" src="%s" />'
                 content = ''
                 for photo in post.get('photos'):
-                    content += '\n'.join(
-                        fmtstr % (photo.get('caption'),
-                                  photo.get('original_size').get('url')))
-                content += '\n\n' + post.get('caption')
+                    content += '\n'.join(fmtstr % (photo.get('caption'),
+                                                   photo.get('original_size').get('url'))
+                                        for photo in post.get('photos'))
```
boxydog, line 473: Is this right? It looks like you're iterating across post.get('photos') on line 473, but it's already doing that on line 470.

boxydog:
I would close this, as equivalent functionality was merged in #3215.

justinmayer:
Thanks, Kevin, for the contribution, the equivalent of which has now been merged. 🌟

PR #3189, opened 2023-08-28
Add GitHub Actions workflow for GitHub Pages

seanh:
Add a GitHub Actions workflow that users can use to publish their Pelican sites to GitHub Pages by running `pelican` on GitHub Actions, without having to run `pelican` locally or push the output directory to a branch. See: https://github.com/getpelican/pelican/discussions/3174 for some discussion.


.github/workflows/github_pages.yml
```diff
@@ -0,0 +1,64 @@
+name: Deploy to GitHub Pages
```
seanh, line 1: GitHub themselves provide starter workflows for publishing to GitHub Pages using various static site generators (unfortunately not including Pelican), the way I've designed this workflow is similar to these: https://github.com/actions/starter-workflows/tree/main/pages

.github/workflows/github_pages.yml
```diff
@@ -0,0 +1,64 @@
...
+        uses: actions/checkout@v3
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.11"
+      - name: Configure GitHub Pages
+        id: pages
+        uses: actions/configure-pages@v3
+      - name: Install requirements
+        run: pip install "pelican[markdown]"
```
seanh, line 37: Might want to add a workflow option for the `pelican[markdown]` string here so that users can override it to supply their own optional packages or to use a specific version of pelican. Users would do something like this in their `pelican.yml` files:

```yml
with:
  requirements: "pelican[markdown]==4.8.0 typogrify"
```

.github/workflows/github_pages.yml
```diff
@@ -0,0 +1,64 @@
...
+      - name: Print Pelican settings
+        run: pelican --print-settings > .pelican_settings.json
+        shell: bash
+      - name: Cache
+        uses: actions/cache@v3
+        with:
+          key: ${{ runner.os }}-${{ hashFiles('.pelican_settings.json') }}
+          path: ${{ inputs.cache-path }}
```
seanh, line 45: I'm using a hash of the output of `pelican --settings` as the cache key, so if any of the user's Pelican settings have changed that should invalidate the cache, even if the user is using something other than `publishconf.py` for their settings file or if their settings file imports other files etc. See https://github.com/actions/cache#creating-a-cache-key

docs/tips.rst
```diff
@@ -110,6 +123,98 @@ branch of your GitHub repository::
...
+           settings: "my_settings.py"
+
+* If you've changed your Pelican ``OUTPUT_PATH`` setting from the default you
+  should pass an ``output-path`` setting to tell GitHub Actions about your
+  custom output path:
+
+   .. code-block:: yaml
+
+       with:
+         output-path: "__output__/"
```
seanh, line 185: It might be possible for the workflow to automatically determine the value of this and other settings from the output of `pelican --print-settings` but I found this awkward to implement: I thought I could just do something like `pelican --print-settings | jq '.OUTPUT_PATH'` but the output of `--print-settings` doesn't appear to be valid JSON (?) and pelican also prints warnings at the top of the output even with `--quiet`.

docs/tips.rst
```diff
@@ -110,6 +123,98 @@ branch of your GitHub repository::
...
+   .. code-block:: yaml
+
+       deploy:
+         uses: getpelican/pelican/.github/workflows/github_pages.yml@main
+         permissions:
+           contents: "read"
+           pages: "write"
+           id-token: "write"
+         with:
+           settings: "my_settings.py"
```
seanh, line 176: Might be better to just have a single table documenting all the available `with:` options and their default values

seanh:
When I have a bit more time I can update [my demo site](https://github.com/seanh/pelican-github-pages-demo/actions/runs/6003395421) to use the `github_pages.yml` workflow from this branch, so that'll work as a test/demo of this PR (the demo site currently uses the one from my <https://github.com/seanh/pelican-github-pages> repo)

.github/workflows/github_pages.yml
seanh: Hmm, I'm not sure if this is correct. GitHub Actions caches are immutable, they can't be updated. So we need to make sure that whenever the input contents change the cache key changes so that we create a new cache. I think the cache key needs to include hashes of everything that goes into creating the site, including the `content` directory, the settings file and any other files that it imports or depends on, and anything else that affects the result of rendering the site.

docs/tips.rst
```diff
@@ -110,6 +123,98 @@ branch of your GitHub repository::
...
+large sites.  Just pass a ``cache: true`` option to the workflow. If you've
+changed the value of Pelican's ``CACHE_PATH`` setting then you must also pass a
+``cache-path`` option to tell GitHub Actions the location of your cache
+directory. For example:
+
+.. code-block:: yaml
+
+    with:
+      cache: true
+      cache-path: "__cache__"
```
seanh, line 206: If we decide to keep this caching function the workflow should probably use `--cache-path` for force the cache path to a known value, then this `cache-path` option won't be needed

docs/tips.rst
seanh: Ah, the workflow should just use `--output` to force the output path to a known value

.github/workflows/github_pages.yml
avaris: > something other than publishconf.py

Not true. Any `pelican` command that does not specify the config file will look for and load `pelicanconf.py`. So, `pelican --print-settings` will dump settings defined in `pelicanconf.py` if it exists or just default values if it doesn't.

.github/workflows/github_pages.yml
avaris: I'm not sure how I feel about hosting a workflow file under `.github/workflows` that is *not* used in this repository, just to make it reusable. I'm just saying this before I start reviewing, which will have to wait for some free time for me.

seanh: Reusable workflows have to be under `.github/workflows/` otherwise GitHub won't recognise them. Unfortunately they can't be in a subdirectory of `.github/workflows/` either ([docs](https://docs.github.com/en/actions/using-workflows/reusing-workflows#creating-a-reusable-workflow)).

It could be moved into a `.github/workflows/` directory of a separate `getpelican/reusable-workflows` repo, but the nice thing about having the workflow be in the `getpelican/pelican` repo itself is that you can change the workflow and its associated documentation at the same time, in one PR.

.github/workflows/github_pages.yml
```diff
@@ -0,0 +1,61 @@
...
+        uses: actions/checkout@v3
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.11"
+      - name: Configure GitHub Pages
+        id: pages
+        uses: actions/configure-pages@v3
+      - name: Install requirements
+        run: pip install ${{ inputs.requirements }}
```
seanh, line 40: These feels a bit dodgy: the user could publish a malicious package containing a `pelican` command to PyPI and use `inputs.requirements` to have this workflow install that package, then when the workflow runs `pelican` below it does something malicious. But I think it's okay because the workflow runs in the context of the user's repo not this repo, so the user could only do something malicious to themselves. And this option seems necessary in order to allow the user to install optional things like markdown, typogrify and plugins.

.github/workflows/github_pages.yml
```diff
@@ -0,0 +1,61 @@
...
+      - name: Build Pelican site
+        run: |
+          pelican \
+            --settings "${{ inputs.settings }}" \
+            --extra-settings SITEURL='"${{ steps.pages.outputs.base_url }}"' \
+            --output "${{ inputs.output-path }}"
```
seanh, line 46: For `--output` instead so that any `OUTPUT_PATH` in the user's Pelican settings is ignored. This is necessary because the upload-artifact step below needs to know what the output path is. The user can still customize the output path by passing an `output-path` input to the workflow

.github/workflows/github_pages.yml
seanh: Done

.github/workflows/github_pages.yml
seanh: :facepalm: Yes, of course. I think the cache key should just be based on the git commit SHA then, but for now I've removed it from this PR: it can be done in a separate PR since it's a tricky problem on its own

docs/tips.rst
seanh: Done

docs/tips.rst
seanh: Removed the caching stuff from this PR

docs/tips.rst
seanh: Done

seanh:
@avaris I've done a bit more work on this and it's ready for review now:

* Removed the caching stuff from this PR since it's a bit of a tricky problem in its own right. If desired I can send a follow-up PR to add caching so we can discuss and accept or reject it separately
* Added a `requirements` input to the workflow to enable users to install things like markdown, typogrify and plugins
* Added `settings` and `output-path` inputs to the workflow, corresponding to `pelican`'s `--settings` and `--output` options
* Tidied up the docs by using a table to document all the workflow inputs at once

I've also changed https://github.com/seanh/pelican-github-pages-demo/ to use the workflow from this PR, so see that repo for a working example

avaris:
Thanks @seanh, I will certainly take a look when my schedule permits, but it's very unlikely for the next couple weeks :(.

seanh:
Hey, no worries :+1:

.github/workflows/github_pages.yml
lioman: > I'm not sure how I feel about hosting a workflow file under `.github/workflows` that is _not_ used in this repository, just to make it reusable. I'm just saying this before I start reviewing, which will have to wait for some free time for me.

That is a valid point. On the other side, I think it would be good to have (even if I don't use it). Many projects on GH just use a different approach: They create a dedicated repo for that action in same org.
One could do the same and have the minimal documentation as a pelican site deployed as github page from that said repo.
Benefits:
  - this repo is not cluttered with unused code
  - the action is tested and used with it's own documentation
 
Con: 
  - another repo to maintain

.github/workflows/github_pages.yml
```diff
@@ -0,0 +1,61 @@
...
+    inputs:
+      settings:
+        required: true
+        description: "The path to your Pelican settings file (`pelican`'s `--settings` option), for example: 'publishconf.py'"
+        type: string
+      requirements:
+        required: false
+        default: "pelican"
+        description: "The Python requirements to install, for example to enable markdown and typogrify use: 'pelican[markdown] typogrify'"
+        type: string
```
avaris, line 13: I think there is an undocumented feature of writing `"-r requirements.txt"` here given that repo contains that file, right? Might be worth mentioning :).

avaris (approved):
I added the additional documentation. Good to merge, thanks!

.github/workflows/github_pages.yml
avaris: I added a commit for this.

justinmayer (approved):
Many thanks to @seanh for the contribution and to @avaris for reviewing and documenting.

PR #3197, opened 2023-10-02
Update python test versions

lioman:
- add 3.12
- remove 3.7 since it is out of support
- use 3.11 as stable base version for linter and docs

# Pull Request Checklist

Resolves: #issue-number-here 

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,8 +16,6 @@ jobs:
       matrix:
         config:
           - os: ubuntu
-            python: "3.7"
-          - os: ubuntu
             python: "3.8"
           - os: ubuntu
             python: "3.9"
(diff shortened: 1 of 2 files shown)


lioman:
Ok this pr is probably a bit ahead of time. 3.12 is just released and GitHub has not updated their available python versions yet.

pauloxnet:
It can be better to have two separate PRs: one to remove outdated Python versions and one to add Python 3.12 support (which maybe Github will provide in the meantime).

justinmayer:
We should probably also remove all other references to Python 3.7. Some of the ones I found are:

- setup.py
- pyproject.toml
- docs/contribute.rst
- docs/install.rst
- docs/quickstart.rst

Regarding the latter three places in the docs, I think we should choose *one* canonical place to document the minimum supported Python version, and then link to that from other places so that every time a Python version goes EOL we can minimize the number of changes we have to make.

lioman:
Good idea - I will try to get this done

lioman:
I close this in favor of #3199 and another yet to be created one

PR #3200, opened 2023-10-04
Only run 'Deploy' action on main repository

lioman:
Deploy action will always fail on forks as the token is not there, this happens on the fork especially, when you are syncing with, main one. This will improve the setup for collaborators.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -97,7 +97,7 @@ jobs:
     environment: Deployment
     needs: [test, lint, docs]
     runs-on: ubuntu-latest
-    if: github.ref=='refs/heads/master' && github.event_name!='pull_request'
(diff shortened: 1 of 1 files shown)


justinmayer (approved):
Thanks for the enhancement, @Lioman. I wish it were possible to globally disable a GitHub Action job from running on any fork _without_ having to specify the origin org+repository name explicitly, but that doesn't seem possible. There is a pre-existing discussion thread about precisely that topic, to which [I posted a comment with my own suggestions](https://github.com/orgs/community/discussions/9098#discussioncomment-7185361).

For now, I think explicitly naming `"getpelican/pelican"`, as is done here, is the only way to do it, so we might as well use that method until a better solution comes along.

lioman:
Just researched it, and I was kind of surprised, that there is no easier/cleaner solution then this.

PR #3209, opened 2023-10-23
Documentation correction

TheHorusHawk:
In content.html of documentation, section 'Linking to internal content', examples for ReStructured and Markdown now have the right filenames (md for Markdown and rst for ReStructured.)

Checked new version of documentation locally via invoke docserve as per the guidelines.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,3 @@
+Release type: patch 
+
+In content.html of documentation, section 'Linking to internal content', examples for ReStructured and Markdown now have the right filenames (md for Markdown and rst for ReStructured.)
--- a/docs/content.rst
+++ b/docs/content.rst
@@ -260,8 +260,8 @@ In this example, ``article1.rst`` could look like this::
 
     See below intra-site link examples in reStructuredText format.
 
-    `a link relative to the current file <{filename}../article2.md>`_
-    `a link relative to the content root <{filename}/article2.md>`_
+    `a link relative to the current file <{filename}../article2.rst>`_
+    `a link relative to the content root <{filename}/article2.rst>`_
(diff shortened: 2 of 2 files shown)


justinmayer:
I think there is a misunderstanding about what is being described in the documentation. `article1.rst` and `article2.md` are two files referenced in the file "tree" just above the section in which you made changes. In the first example, `article1.rst` is linking to `article2.md`, and then vice-versa in the second example.

In short, I believe the documentation is already correct in this regard, and the proposed changes would be incorrect insofar as they would refer to files that do not exist in the example file tree above the examples.

TheHorusHawk:
I see now. I didn't realise how it connected to the tree above.

As a basic user, who writes his website purely in markdown, I was looking for a simple example on how to link between files. I realise you were showing an advanced use-case, in that the files can be linked internally even if they use different markup languages.

The documentation is indeed correct. As a beginner to SSGs, it would have benefited me to see a simpler use-case.

justinmayer:
I can see where you are coming from. I'm not sure I would call it "advanced" — the idea was to demonstrate the functionality while being inclusive of folks who use reST as well as folks who use Markdown. And some people do indeed mix the two, so it's useful (as you said) to show that links are agnostic to the type of markup the target source file is written in.

In any case, I imagine that at this point the internal source file link behavior is clear. 😊

TheHorusHawk:
It is clear, Justin, and, to be honest, even when I thought it was a mistake, it was pretty clear what I had to change to make it work, and, you're right, it shows it's agnostic to the markup and that's good to know.

I agree with you now that the documentation is fine.

Issue #3203, opened 2023-10-11
Deprecate EOL version and centralize information about it


lioman:
- [x] I have searched the [issues](https://github.com/getpelican/pelican/issues?q=is%3Aissue) (including closed ones) and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://docs.getpelican.com/) and believe that my question is not covered.
- [x] I am willing to lend a hand to help implement this feature. 

## Feature Request
As a pelican developer I just want to support current Python versions, so that I have not to test and cover more versions than necessary. 
As a pelican user I want to have a quick way to understand, which versions are actually supported.

## Description
In PR #3197 there was the proposal to ditch 3.7 as it is [EOL](https://peps.python.org/pep-0537/#schedule-last-security-only-release).  This. Issue tries to achieve the same but also centralize the information about supported versions. Currently, it seems that the version is mentioned at least in these locations:
- `setup.py`
- `pyproject.toml`
- `docs/contribute.rst`
- `docs/install.rst`
- `docs/quickstart.rst`

## Proposal

use `pyproject.toml` as the single place to define that information and link it in other places (or ditch them if not needed anymore).

lioman:
I'm willing to implement this, just opened it as a place for discussion

pauloxnet:
+1 for me

lioman:
My first open point for discussion:
We can either use poetry for building the package or just use setuptools and modernize it by using `pyproject.toml` to configure it. I have experience with both and both work. Are there any opinions about it?

pauloxnet:
I usually prefer the KISS approach and use `setuptools` but I'd ask @justinmayer about this point.

lioman:
About the KISS approach: I'm not sure what is the simplest thing.
1. using the old tool (`setuptools`) and transferring it to `pyproject.toml`
2. using **one** tool (`poetry`)

justinmayer:
[...]

PR #3220, opened 2023-10-28
Make build configuration compatible with PEP621

lioman:
# Pull Request Checklist

Resolves: #3203 

This solves the issue by using pdm as a build tool to resolve dependencies and adding  `pyproject.toml` with PEP621 compliant configuration.

## Todo

- [x] adapt documentation to read `pyproject.toml`
- [x] check if files to build are added (transfer Manifest.in config)
- [x] double check entry points. etc from setup.py
- [x] remove setup.py 


justinmayer:
In the interests of keeping PR scope as small as possible, I may cherry-pick https://github.com/getpelican/pelican/commit/c48f5f1ac4916a2119e50795c1e3230899a70d41 and apply it before reviewing the rest of this PR.

pyproject.toml
```diff
@@ -84,5 +55,49 @@ version-header = "="
...
+    "tox<4.0,>=3.13",
+    "flake8<4.0,>=3.8",
+    "flake8-import-order<1.0.0,>=0.18.1",
+    "invoke<3.0,>=2.0",
+    "isort<6.0,>=5.2",
+    "black<20.0,>=19.10b0",
+]
+
+[tool.pdm.build]
+includes = []
```
justinmayer, line 89: Is this `[tool.pdm.build]` section required? I don't see it listed in the relevant docs: https://pdm.fming.dev/latest/reference/build/

pyproject.toml
```diff
@@ -1,17 +1,30 @@
-[tool.poetry]
+[project]
+authors = [{ name = "Justin Mayer", email = "authors@getpelican.com" }]
+license = { text = "AGPLv3" }
+requires-python = ">=3.8,<4.0"
```
justinmayer, line 4: In other projects I usually see a minimum of 3.8.1, so there must have been something not quite right with the initial 3.8 release, and we should probably follow their lead.

pyproject.toml
```diff
@@ -1,17 +1,30 @@
...
+dependencies = [
+    "blinker>=1.4",
+    "docutils>=0.16",
+    "feedgenerator>=1.9",
+    "jinja2>=2.7",
+    "pygments>=2.6",
+    "python-dateutil>=2.8",
+    "rich>=10.1",
+    "unidecode>=1.1",
+    "backports-zoneinfo<1.0.0,>=0.2.1; python_version < \"3.9\"",
```
justinmayer, line 14: Is that escaping required? Could it not be `python_version<3.9` instead?

.pdm-python
avaris, line 1: I don't know `pdm` that well, but I suspect this file isn't needed and would likely be `.gitignore`d?

justinmayer (changes requested):
Many thanks for working on this, @lioman! In addition to the in-line comments, I have two additional thoughts:

1. I see you have chosen to use PDM for the build backend. I have only using Hatchling in combination with PDM, and so I have no experience with PDM's build backend. But I imagine it works just fine and has the ostensible advantage of not needing an additional dependency (i.e., Hatchling). So I am down to try it and see how it goes.
2. I have a few mild preferences regarding the order of items in `pyproject.toml`, so I will add a commit with some proposed changes. Those changes may also include a couple of the suggested changes I made in my in-line comments.

justinmayer:
Also, the c48f5f1ac commit should be dropped from this PR, as I already added it to `master`. Then, this PR should be rebased on current `master`, which should fix the failing test currently shown here.

.pdm-python
lioman: Sounds reasonable. Will add that to ignore

.pdm-python
```diff
@@ -0,0 +1 @@
+/Users/eliaskirchgaessner/Development/FOSS/pelican/.venv/bin/python
```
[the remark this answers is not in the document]
lioman: That happens when you are in a hurry. Will remove the unnecessary files.

pdm.lock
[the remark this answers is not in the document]
offbyone: I would prefer that the locking happen on the static site side, not pelican; pelican is a library/tool in this form.

justinmayer:
1. Looks like c48f5f1 is still present. That commit should be dropped from this PR.
2. There is some commented-out configuration that should probably be removed?
3. Some other PRs were merged, so there are conflicts that need to be resolved.

pyproject.toml
```diff
@@ -1,21 +1,24 @@
...
     "Framework :: Pelican",
+    "Intended Audience :: End Users/Desktop",
+    "License :: OSI Approved :: GNU Affero General Public License v3 or later (AGPLv3+)",
     "Operating System :: OS Independent",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
```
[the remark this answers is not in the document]
lioman: One can mention explicitly every supported version but keeping it shorter and just saying "3" would be fine for me.
Any opinions @justinmayer I just copied the existing entries

pyproject.toml
lioman: It was just the start to include additional files that are mentioned in the `Manifest.in`. But these files don't exist in the existing release anyhow, so I removed it again.

pyproject.toml
lioman: no one need either escaping or `"...< '3.9'"`

pyproject.toml
```diff
@@ -84,5 +68,36 @@ version-header = "="
 version-strings = ["setup.py"]
 build-system = "setuptools"
```
lioman, line 69: @justinmayer I think this need a change as the version will be only in this file and build-system will be changed too. Is this possible with `autopub`?

pyproject.toml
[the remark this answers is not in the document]
justinmayer: @offbyone: Is that how PyPI behaves, though? I know PyPI shows supported versions if you mention them explicitly (such as https://pypi.org/project/pelican-search/), but don't know if they do that if you just say "3".

pyproject.toml
justinmayer: I don't understand. Can you elaborate?

lioman: Two things: 
I don't know autopub and if it is needed. If I should build a pipeline, my deploy step would be running `pdm build` and `pdm publish` based on branch/tag criteria. So my question would be what does autopub adds to that pipeline?
Second question:
version string is set to setup.py a file the will be removed in one of the next commits as all information is in pyproject.toml. The build system needs to be changed as well. Since you are the author of autopub, you can give me perhaps a hint on what to add there or if this should/ can be dropped.

justinmayer: We use [AutoPub](https://justinmayer.com/projects/autopub/) to automatically publish releases. You can read more about it via that link, including why I created it. I believe it obviates the need for calling `pdm build` and `pdm publish`, neither of which are likely necessary in an AutoPub context.

You can see the existing `autopub build` step in the GitHub Actions workflow. That invocation runs [`python -m build`](https://github.com/autopub/autopub/blob/d89ef5a7a894fbfbcefe2d46e1e4dbc2b08fdc81/autopub/build_release.py#L14) ([docs for `build`](https://pypa-build.readthedocs.io/en/latest/)), a build front-end that uses the build backend specified in `pyproject.toml` to build the `sdist` and wheel. We use the PyPI Trusted Publisher system to actually publish the package to PyPI ([relevant configuration](https://github.com/getpelican/pelican/blob/fad2ff7ae3cd0ea8b974db5fe42de7383da679c1/.github/workflows/main.yml#L127-L129)), which is why `pdm publish` is not necessary.

Going forward the version number should only ever exist in `pyproject.toml` and will be automatically incremented by AutoPub based on the `Release type:` value specified in `RELEASE.md` files included in pull requests.

This might be easier in an interactive conversation. Can you come find me in the chat room? https://web.libera.chat/?#pelican

.github/workflows/main.yml
```diff
@@ -51,23 +51,18 @@ jobs:
...
+          cache: true
+          cache-dependency-path: ./pyproject.toml
       - name: Install dependencies
         run: |
-          poetry env use "3.9"
-          poetry install --no-interaction --no-root
+          pdm install
       - name: Run linters
-        run: poetry run invoke lint --diff
+        run: pdm lint --diff
```
justinmayer, line 65: Shouldn't this be `pdm run invoke lint --diff` ?

pyproject.toml
```diff
@@ -24,57 +27,36 @@ classifiers = [
...
+Documentation = "https://docs.getpelican.com"
 
-[tool.poetry.scripts]
+[project.scripts]
 pelican = "pelican.__main__:main"
 pelican-import = "pelican.tools.pelican_import:main"
-pelican-plugins = "pelican.plugins._utils:list_plugins"
 pelican-quickstart = "pelican.tools.pelican_quickstart:main"
 pelican-themes = "pelican.tools.pelican_themes:main"
+pelican-plugins = "pelican.plugins._utils:list_plugins"
```
justinmayer, line 59: I'm not sure why this was moved. Perhaps put it back where it was, so it's in alphabetical order?

pyproject.toml
```diff
@@ -83,8 +65,40 @@ git-email = "52496925+botpub@users.noreply.github.com"
 changelog-file = "docs/changelog.rst"
 changelog-header = "###############"
 version-header = "="
-version-strings = ["setup.py"]
-build-system = "setuptools"
+version-strings = ["pyproject.toml"]
```
justinmayer, line 68: This line is not necessary. The `version-strings` AutoPub directive is only for _additional_ version strings that should be incremented, aside from the canonical one in `pyproject.toml`.

justinmayer (changes requested):
Hi @lioman. I made some more recommended fixes for some problems I found.

.github/workflows/main.yml
lioman: No, this is not needed as there is a script named `lint` configured

pyproject.toml
justinmayer: This is marked as `resolved`, but the spurious line seems to still be present.

docs/contribute.rst
```diff
@@ -15,16 +15,16 @@ Setting up the development environment
 ======================================
 
 While there are many ways to set up one's development environment, the following
-instructions will utilize Pip_ and Poetry_. These tools facilitate managing
+instructions will utilize Pip_ and pdm_. These tools facilitate managing
```
justinmayer, line 18: Let's use the canonical capitalized name when referring to PDM in all places (except when it's clearly a CLI invocation like `pdm install`).

pyproject.toml
```diff
@@ -24,52 +27,31 @@ classifiers = [
...
+    "blinker>=1.4",
+    "docutils>=0.16",
+    "feedgenerator>=1.9",
+    "jinja2>=2.7",
+    "pygments>=2.6",
+    "python-dateutil>=2.8",
+    "rich>=10.1",
+    "unidecode>=1.1",
+    "backports-zoneinfo<1.0.0,>=0.2.1;python_version<'3.9'",
+    "watchfiles",
```
justinmayer, line 41: We should probably specify a minimum version for `watchfiles`, as we do for all the other dependencies:

    "watchfiles>=0.21.0",

pyproject.toml
```diff
@@ -83,8 +65,41 @@ git-email = "52496925+botpub@users.noreply.github.com"
...
+    "pytest-xdist<3.0,>=2.0",
+    "tox<4.0,>=3.13",
+    "flake8<4.0,>=3.8",
+    "flake8-import-order<1.0.0,>=0.18.1",
+    "invoke<3.0,>=2.0",
+    "isort<6.0,>=5.2",
+    "black<20.0,>=19.10b0",
+    "ruff>=0.1.3,<1.0.0",
+    "tomli;python_version<'3.11'",
+]
```
justinmayer, line 101: 1. Should we replace the long `dev` group with a more-specific set of `docs`, `lint`, and `test` groups? ([example](https://github.com/pelican-plugins/featured-image/blob/38dbae18cf684657670f0b375dd1a8a68a5323a5/pyproject.toml#L39-L50))
2. Except for Jinja2, Markdown, Pygments, and Furo, I think we can remove the upper bounds. More often than not, upper bounds cause more problems than they solve. (See example of the [Black project dropping upper bounds on dependencies](https://github.com/psf/black/pull/2718).)
3. Given that we are going to use Ruff, I don't think we need `flake8`, `flake8-import-order`, or `isort` any longer — Ruff contains all the functionality of those tools.

justinmayer (changes requested):
Great work, Lioman. We're getting close! I commented on a few areas where I think we can still make improvements.

pyproject.toml
lioman: yes I think so too and thought about it. To keep this PR a bit smaller I propose to do it in an additional PR

pelican/tests/build_test/test_wheel.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+@pytest.mark.skipif(
+    "not config.getoption('--check-wheel')",
+    reason="Only run when --check-wheel is given",
+)
+def test_wheel_contents(pytestconfig):
+    """
+    This test, should test the contents of the wheel to make sure,
+    that everything that is needed is included in the final build
```
justinmayer, line 13: I suggest removing the two commas in this sentence, as well as adding a period at the end.

pelican/tests/build_test/test_wheel.py
```diff
@@ -0,0 +1,28 @@
...
+    """
+    wheel_file = pytestconfig.getoption("--check-wheel")
+    assert wheel_file.endswith(".whl")
+    files_list = ZipFile(wheel_file).namelist()
+    ## Check is theme files are copiedto wheel
```
justinmayer, line 18: Maybe a single `#` instead of two? Also, there are some errors. I would change the text to:

    # Check if theme files are copied to wheel

pelican/tests/build_test/test_wheel.py
```diff
@@ -0,0 +1,28 @@
...
+    simple_theme = Path("./pelican/themes/simple/templates")
+    for x in simple_theme.iterdir():
+        assert str(x) in files_list
+
+    ## Check is tool templatesare copiedto wheel
+    tools = Path("./pelican/tools/templates")
+    for x in tools.iterdir():
+        assert str(x) in files_list
```
justinmayer, line 26: I think this comment was copy+pasted and not relevant to what appears below it?

pyproject.toml
justinmayer: I agree that deferring the separation into development dependency groups makes sense, but I think my comments in (2) and (3) are probably best resolved in the context of this PR, don't you think?

justinmayer (changes requested):
A few more comments. So sorry! I'm sure this is annoying. Thank you for your patience!

justinmayer:
Hey @Lioman. As I just mentioned in the chat room, I didn't realize my request to remove the upper bound on dependencies would result in that huge mass-formatting of other files. I think it's best to revert that and go back to what you had before that change (eb052cae096c27dba40ccf3a467b197a7bc91e1b, presumably). I am *really* sorry.

lioman:
@justinmayer removed those commits. I will create an additional PR with those commits and suggest to apply it, just before the release, to minify the friction.

.github/workflows/main.yml
```diff
@@ -51,23 +51,18 @@ jobs:
...
-          cache: "poetry"
-          cache-dependency-path: "pyproject.toml"
+          python-version: 3.9
+          cache: true
+          cache-dependency-path: ./pyproject.toml
       - name: Install dependencies
         run: |
-          poetry env use "3.9"
-          poetry install --no-interaction --no-root
+          pdm install
```
avaris, line 63: I think this can be:
```
pdm install --no-default --dev
```

avaris (approved):

justinmayer (approved):
Many thanks to @lioman for the hard work on (and patience with) this endeavor. Much appreciated! 👏

PR #3225, opened 2023-10-28
Add a task test-coverage to generate a coverage report

boxydog:
# Pull Request Checklist

### DIFF
--- a/docs/contribute.rst
+++ b/docs/contribute.rst
@@ -75,6 +75,9 @@ via::
 
     invoke tests
 
+(For more on Invoke, see ``invoke -l`` to list tasks, or
+https://pyinvoke.org for documentation.)
+
 In addition to running the test suite, it is important to also ensure that any
 lines you changed conform to code style guidelines. You can check that via::
 
--- a/tasks.py
+++ b/tasks.py
@@ -45,6 +45,13 @@ def tests(c):
 
 
 @task
+def coverage(c):
+    """Generate code coverage of running the test suite."""
+    c.run(f"{VENV_BIN}/pytest --cov=pelican", pty=PTY)
+    c.run(f"{VENV_BIN}/coverage html", pty=PTY)
+
+
+@task
 def black(c, check=False, diff=False):
     """Run Black auto-formatter, optionally with --check or --diff"""
     check_flag, diff_flag = "", ""


docs/contribute.rst
```diff
@@ -75,6 +75,9 @@ via::
 
     invoke tests
 
+(For more on invoke, see `invoke -l` to list tasks, or
+[pinvoke.org](pyinvoke.org) for documentation.)
```
justinmayer, line 79: This is reST, not Markdown, so you'll need double-ticks. Also, I'd include the scheme. In short:
```
(For more on Invoke, see ``invoke -l`` to list tasks, or
[pyinvoke.org](https://pyinvoke.org) for documentation.)

File: tasks.py [hunk 2]
```diff
@@ -44,6 +44,13 @@ def tests(c):
     c.run(f"{VENV_BIN}/pytest", pty=PTY)
 
 
+@task
+def test_coverage(c):
```
justinmayer, line 48: Any reason not to name this task `coverage`? (Shorter, doesn't cause contention with `inv tests` completion)

justinmayer (changes requested):
Thanks, @boxydog! I made a few corrections and a suggestion.

docs/contribute.rst
justinmayer: Oops. Because it's reST, you'll also need to change the link syntax.

tasks.py
boxydog: Sure, changed to "coverage"

docs/contribute.rst
boxydog: Ah thanks. Changed to accept your comment, and look okay in rST (the `[pyinvoke.org]` didn't work).

tasks.py
```diff
@@ -44,6 +44,13 @@ def tests(c):
     c.run(f"{VENV_BIN}/pytest", pty=PTY)
 
 
+@task
+def coverage(c):
+    """Generate code coverage of running the test suite"""
```
justinmayer, line 49: Sorry! One more thing… I think the linter will complain (not yet, but at some point in the future) without a period at the end of the doc-string, so perhaps get ahead of that and add it now?

boxydog: Added, just to get this merged.

But, I wouldn't bother with stuff like that. If the linter isn't complaining, leave it.

Then, fix the linter, and fix all the stuff at once. Every other task in this file doesn't have a period, so there will be a lot of stuff to fix.

justinmayer (approved):
Thank you for the contribution, @boxydog! ✨

tasks.py
justinmayer: Fair point, and duly noted 😊

Issue #3216, opened 2023-10-26
More pre-commit filters, e.g. isort


boxydog:
- [X] I have searched the [issues](https://github.com/getpelican/pelican/issues?q=is%3Aissue) (including closed ones) and believe that this is not a duplicate.
- [X] I have searched the [documentation](https://docs.getpelican.com/) and believe that my question is not covered.
- [X] I am willing to lend a hand to help implement this feature. 

## Feature Request

I submitted a pull request that passed pre-commit but failed lint checks, because of import order.

[isort](https://pycqa.github.io/isort/docs/configuration/pre-commit.html) is a pre-commit filter that I think would've fixed my import order. I feel like we should add it.

Some others I'd recommend, in rough order:

- pyupgrade
- autoflake
- black
- pylint

- pre-commit check-case-conflict, check-docstring-first, check-merge-conflict, check-symlinks, debug-statement, mixed-line-ending, forbid-crlf, forbid-tabs

I understand that adding each of these would result in code shift getting the project to pass. My approach would be to add them all, see which are easy to keep, and drop the rest for now. Time box to a couple hours.

It has this problem that it would conflict with every other PR, so I would want permission to get it in fairly quickly, otherwise resolving conflicts becomes too painful.

Below is a sample pre-commit from one of my projects. It's not a perfect fit, but I could adjust it. For example, versions. I was targeting newer pythons, but could tune it back to python 3.7, which seems to be the oldest one you test.

```
# See http://pre-commit.com/#python
# See https://github.com/pre-commit/pre-commit-hooks
# Run 'pre-commit install' to install the pre-commit hooks
repos:

# See also https://adamj.eu/tech/2021/09/16/introducing-django-upgrade/
- repo: https://github.com/asottile/pyupgrade
  rev: v3.15.0
  hooks:
    - id: pyupgrade
      args: [ --py39-plus ]

- repo: https://github.com/pre-commit/pre-commit-hooks
  rev: v4.5.0
  hooks:
  - id: check-added-large-files
  - id: check-ast
  - id: check-case-conflict
  - id: check-docstring-first
  - id: check-merge-conflict
  - id: check-symlinks
  - id: debug-statements
  - id: detect-private-key
  # black handles quoting
  # - id: double-quote-string-fixer
  - id: end-of-file-fixer
  - id: mixed-line-ending
  - id: trailing-whitespace
    exclude: (.csv|.tsv)$
  - id: pretty-format-json
[...]

PR #3217, opened 2023-10-26
Fix pre-commit run --all

boxydog:
# Pull Request Checklist

Resolves: #issue-number-here 

- [ ] Ensured **tests pass** and (if applicable) updated functional test output
- [X] Conformed to **code style guidelines** by running appropriate linting tools
- [X] Added **tests** for changed code
- [X] Updated **documentation** for changed code

Very first part of #3216.

- Ran `pre-commit run --all`
- Added pre-commit github action

PR #3227, opened 2023-10-29
Format all

offbyone:
Resolves: #3216 

# Pull Request Checklist

Once this PR is _mostly_ approved, I will add a `.git-blame-ignore-revs` file with the revision of the bulk format commit.

After this, pre-commit should pass. This would supersede #3217 since that's included in here as well.


offbyone:
I pulled down the docs that were generated and they look fine; I was most worried about those, because docs generation often goes sideways after updates.

boxydog:
If this is ruff-format, there are at least two issues here: 1) ruff-format says it is unstable, I would advise against relying on an unstable tool, 2) now every PR still outstanding will have difficult conflicts to resolve.

I wouldn't do this.

boxydog:
Also: this does not quite supercede #3217, which also includes running `pre-commit run --all` in the github main.yaml.

offbyone:
The third commit in this PR is the outcome of running `pre-commit run --all` 

As for ruff's stability, I'll include https://astral.sh/blog/the-ruff-formatter#production-ready; from my experience with it, I agree with their claim that it is solid enough to use for production codebases.

boxydog:
Sounds like you understand the implications. Works for me.

boxydog:
Ah, but it still doesn't have the github check for pre-commit --all. It looks like this:

```
      # see https://github.com/pre-commit/action/#using-this-action
      - name: pre-commit checks
        uses: pre-commit/action@v2.0.0
        env:
          # it's okay for github to commit to main/master
          SKIP: no-commit-to-branch
```

boxydog:
That would make github check that pre-commit fully passes. That's what's in #3217, but not here yet.

offbyone:
oh, I missed that. Good call out. Would you like me to add that to this pool request, or do you think you'd like to merge 3217 independently?

boxydog:
i would recommend adding the fully-pass check in the same PR as fixing it (with the reformatting), so it doesn't get broken again. whether that happens here, or in 3217, or somewhere else, no opinion.

I just think the pre-commit fixes and the github check that pre-commit passes should be in the same PR to shut the door for good.

boxydog:
justin added `uses: pre-commit/action@v3.0.0`, looks good for me.

Looks like it has to be run through pre-commit once more, tho.

justinmayer:
@offbyone: For reasons that are boring, I needed to do some surgery on this PR, and since I can't re-write history in your fork, I created a new PR that contains these commits in a cleaned-up form: #3231

PR #3236, opened 2023-11-01
Add sphinxext-opengraph to pyproject dev requirements

avaris:
`sphinxext-opengraph` is required to build the docs, but it was missing from `pyproject.toml`


justinmayer (approved):
Nice catch. Thanks!

PR #3239, opened 2023-11-03
More Ruff checks, and make it fix

boxydog:
# Pull Request Checklist

Resolves: #3216

Note: there are no manual python fixes in this PR, it's all ruff.

Also, making ruff and ruff-format both fix means if there are pre-commit failures, you can just run it again and it might pass. This is a common pattern for me: run commit once (auto-fixes), run again (commits).


pelican/__init__.py
```diff
@@ -19,15 +19,15 @@
 
 # pelican.log has to be the first pelican module to be loaded
 # because logging.setLoggerClass has to be called before logging.getLogger
```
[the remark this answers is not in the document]
avaris: And if this is auto "fixed", it needs to be configured to not do that again...

pyproject.toml
```diff
@@ -109,3 +109,86 @@ source-includes = [
 [build-system]
 requires = ["pdm-backend"]
 build-backend = "pdm.backend"
+
+[tool.ruff.lint]
+# default rules include pyflakes ("F")
+# see https://docs.astral.sh/ruff/configuration/#using-pyprojecttoml
+# "F" contains autoflake, see https://github.com/astral-sh/ruff/issues/1647
+# add more rules
+extend-select = [
```
avaris, line 118: I'd much rather have this be `select` and explicitly list everything. I don't like the possibility of `ruff` changing their defaults and we are suddenly running something we did not intent to.

pelican/__init__.py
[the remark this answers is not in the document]
boxydog: While it's good that you pointed this out, having a unit test would be even better.

I also added a function set_logger_class(), to call for the action explicitly instead of relying on a side effect.

The auto-fixing does not get rid of that function, so that is fixed.

pyproject.toml
boxydog: Changed.

pelican/__init__.py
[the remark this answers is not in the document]
avaris: I don't consider this as "fixed". I should not need to "invent" interfaces to make a linter happy. This (`# noqa: I100`) was deliberate, so that the import order checks would not trigger.  If `ruff` auto fixer thinks that it can "remove noqa and reorder imports", then it is not safe to use.

```
from pelican.generators import (
    ArticlesGenerator,  # noqa: I100
```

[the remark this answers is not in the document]
boxydog: I'm not sure what to say.

Apparently we decided to use ruff.

Getting it to complain without fixing seems like a major flaw. No one will be able to auto-fix without this coming up again. So everyone will have to manually fix things until ruff stops complaining? That seems to put a lot of unneeded burden on developers and reviewers, slowing things down because of all the back and forth.

Relying on the side-effect of an import is more dangerous and hard to see than calling it out explicitly with a function, especially with no unit test.

So I don't understand where to go next.

@justinmayer ?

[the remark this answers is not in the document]
avaris: > Relying on the side-effect of an import is more dangerous and hard to see than calling it out explicitly with a function, especially with no unit test.

This is beside the point. If auto fixing can result in behavior change (as in this example), it is simply unacceptable. It is *not safe* to use it.

> So I don't understand where to go next.

`# noqa` has a purpose. It means "I know this doesn't obey the rules but it is an exception. Ignore it.". I don't know why, but seems like `ruff` made up its own code for it ([I001](https://docs.astral.sh/ruff/rules/#isort-i)) instead of following what's been used in flake8. You can try changing the original to ` # noqa I100, I001` and see if that makes this bad behavior goes away.

boxydog:
I added "noqa: I001" and the import line stays in place.

Now when I try to look at the pull request to see tests, github is giving
me 404s half the time, so I think it's flaky at the moment.

pyproject.toml
```diff
@@ -109,3 +109,101 @@ source-includes = [
...
+]
+
+
+[tool.ruff.lint.extend-per-file-ignores]
+
+"pelican/__init__.py" = [
+    # allow imports after a call to a function, see the file
+    "E402"
+]
+"**/management/commands/*" = ["T20"]
```
avaris, line 209: What does this glob refer to?

boxydog: Good eye. That is cut-and-paste from a Django project. Removed.

avaris:
Given the quirks so far, I will have to go through the diff in detail to make sure nothing strange is going on.

boxydog:
Is this PR ready?

boxydog:
> Given the quirks so far, I will have to go through the diff in detail to make sure nothing strange is going on.

Sorry, I missed this message. Works for me.

boxydog:
Rebased to current main. Thoughts?

pyproject.toml
```diff
@@ -110,3 +110,100 @@ source-includes = [
...
+    "PLW2901",
+    "SLF001",
+    "RUF001",
+    "PLR2004",
+    "PLR0912",
+    "PLR0913",
+    "RUF005",
+    "RUF012",
+    "PLR0915",
+    "INP001",
```
avaris, line 179: I don't get this. `INP` has only one rule `INP001`. What's the point of adding checks for it but ignoring the only thing it checks. I didn't look at all of them but there are other similar ones. e.g. `ISC` has 3 rules but two off them are ignored.

boxydog: I just did the quick thing: added each thing in the list, suppressed those subrules with violations, figuring there were probably other rules being checked. It's like closing the gate, so we don't introduce new violations.

You've found a couple of weird cases, it's true, but I don't understand how you'd like to move forward.

I thought it was probably useful to enforce a bunch of the rules quickly. Perhaps you disagree, but what now?

avaris: I just want to understand the logic behind some checks. 

* `INP` for example disallows implicit namespace packages. That's not going to work. We have namespace packages and will probably add more. Why should we disallow that? I understand the reason behind the rule: "Hey, maybe you forgot to add `__init__.py`?" but that's not applicable here. We intentionally use implicit namespace packages.
* `T20` disallows `print` (or its cousin `pprint`), which is a very project specific convention. We clearly have scripts that `print` stuff (`pelican-quickstart` etc). Either we don't care about this rule or we switch from `print` to something else. Until then, this rule is effectively not enforceable.
* `ISC` is one thing I can agree with but we are ignoring most of them :).

I understand that to un-ignore some of these rules, perhaps some manual fixes are needed. That is fine but then I'd treat this as a "work-in-progress" and the ignored rules more like a "to-do list" :).

boxydog: I really appreciate your careful review. Thanks.

My overall goal is to make some progress here, merge something. In particular, the auto-fix ("--fix") and auto-format.

I agree the ignored rules are like a TODO list. I was hoping to merge that TODO list, since it represents some knowledge that might be useful to lay out in the codebase.

However, if you don't want to merge that, then another possibility is to take out the extra rules and the suppressions.

What is the quickest path to moving forward?

boxydog: Another possibility would be to not include any rules that have suppressions, and drop INP and T20. Not sure where that would leave the list, I'd have to try it.

avaris: At least a comment next to the ignored rule stating what it does would be useful. It would have saved me some work during review and I'd guess it'll be useful when someone picks up these rules to make them pass. I already started doing that, might was well put it here to make your job easier :).

```python
ignore = [
    # suppression in order of # of violations:
  "B007",  # unused-loop-control-variable
  "T201",  # print
  "PLW2901",  # redefined-loop-name
  "SLF001",  # private-member-access
  "RUF001",  # ambiguous-unicode-character-string
  "PLR2004",  # magic-value-comparison
  "PLR0912",  # too-many-branches
  "PLR0913",  # too-many-arguments
  "RUF005",  # collection-literal-concatenation
  "RUF012",  # mutable-class-default
  "PLR0915",  # too-many-statements
  "INP001",  # implicit-namespace-package
  "RUF015",  # unnecessary-iterable-allocation-for-first-element
  "PLR1722",  # sys-exit-alias
  "ISC001",  # single-line-implicit-string-concatenation
  "C408",  # unnecessary-collection-call
  "B904",  # raise-without-from-inside-except
  "UP031",  # printf-string-formatting
  "PLR5501",  # collapsible-else-if
  "PERF203",  # try-except-in-loop
  "B006",  # mutable-argument-default
  "PLR1714",  # repeated-equality-comparison
  "PERF401",  # manual-list-comprehension
  # TODO: these only have one violation each:
  "SLOT000",  # no-slots-in-str-subclass
  "PYI024",  # collections-named-tuple
  "PLW0603",  # global-statement
  "PIE800",  # unnecessary-spread
  "ISC003",  # explicit-string-concatenation
  "EXE002",  # shebang-missing-executable-file
  "C401",  # unnecessary-generator-set
  "B028",  # no-explicit-stacklevel
  "B008",  # function-call-in-default-argument
]
```

avaris: Also probably drop these comments:
> suppression in order of # of violations:
> TODO: these only have one violation each:

While they might be accurate at the time, these kind of comments have a very strong tendency to become incorrect and/or misleading quickly.

boxydog: Good idea. Done.

boxydog: "Done" for adding the comments.

For dropping, how about I add "in Dec 2023"? That would preserve the info (these were in order) without lying.

If I were to work on the TODO list, I'd start with the ones that had one violation each in 2023, it might be helpful to have them marked.

boxydog: I added the "in Dec 2023". But I'm happy to drop those comments entirely if you'd prefer.

justinmayer: I agree that comments next to rules are helpful. As I've mentioned before, the set of rules and ignores that I've used for some of my plugin code-style check experimentation is an example of that practice: https://github.com/pelican-plugins/featured-image/blob/3d3473269c27dfd3ae123a73aebeb1af4d8b3dd0/pyproject.toml#L58-L91

justinmayer (approved):
@boxydog: You kindly asked, “What is the quickest path to moving forward?”, and yet the moving forward has not been very quick. Please accept my sincere apologies for that, as well as my gratitude for your patience. I made a few changes to bring the branch current with code that has been merged to `master` in the interim.

If you are still willing to work on the TODO list as you indicated a few months ago, then I for one would be happy to merge this PR in its current form. Would you be amenable to that?

@avaris: Any follow-up thoughts on the current state of this PR?

_(Note: Windows test failures in CI are unrelated — see #3299)_

boxydog:
> I would be happy to merge this PR in its current form. Would you be amenable to that?

Sounds good to me.

boxydog:
> If you are still willing to work on the TODO list

Oh, I'd be willing to work on TODOs as well. I have no particular order in mind, tho. Open to suggestions. I'd probably just pick a couple marked as TODO, uncomment them, and see what ruff does.

boxydog:
Alternatively, I could try to pick some of the TODOs at the bottom of the "ignore" list and see if I can get rid of them. (The ones at the bottom have only one violation apiece.) Again, I'm open to suggestions.

boxydog:
So where is this PR? I have no planned work. At one point you had said you'd merge it as-is.

I'd suggest adding `--exit-non-zero-on-fix` to the pre-commit filter (so it fails in the github check if anything had to be fixed). I can do that if you wish. Other than that, I'd merge it.

If you wished, I could work on other ruff changes in a later PR.

justinmayer:
Thank you for your patience, @boxydog. If you would be so kind as to make the pre-commit hook addendum you suggested (`--exit-non-zero-on-fix`), we can get this merged and proceed with the other Ruff-related changes in subsequent PRs.

boxydog:
I added `--exit-non-zero-on-fix` to ruff .pre-commit-config.yaml.

Then I noticed the version was old, so I changed it to version v0.4.4.

Then I ran `pre-commit run --all` because it should have no fixes to pass github tests. Ugh, there were fixes, perhaps from new unformatted code from the merge from master? So I committed those fixes.

Then the lint test failed. I think the linter is using v0.1.5 (from pyproject.toml).

So, I flipped the ruff version back to v0.1.5, because I want to make minimal changes.

However, I'd recommend you move to a more recent version in future.

justinmayer:
> Then I ran `pre-commit run --all` because it should have no fixes to pass github tests. Ugh, there were fixes, perhaps from new unformatted code from the merge from master? So I committed those fixes.

Possibly. Or perhaps new fixes were introduced between v0.1.5 and v0.4.4?

> However, I'd recommend you move to a more recent version in future.

I agree that we should upgrade to the latest version of Ruff in a subsequent pull request.

Looking at the check status of this PR, it seems a test is failing, presumably due to the "fix" Ruff made to `pelican/tests/support.py`. This doesn't particularly give me confidence that we can comfortably rely on including Ruff's fixing behavior in a CI context, but I'm willing to give Ruff the benefit of the doubt that such occurrences will (hopefully) be ironed out once we upgrade to v0.4.4 and beyond.

Returning back to the present moment, perhaps for now we revert the changes to that file and add some kind of ignore/exclusion related to that particular fix?

boxydog:
> Possibly. Or perhaps new fixes were introduced between v0.1.5 and v0.4.4?

Ah, right you are.

> Returning back to the present moment, perhaps for now we revert the changes to that file and add some kind of ignore/exclusion related to that particular fix?

Yes. So I did the following:

- git reset HEAD~1
- git checkout pelican (throwing away the ruff v0.4.4 fixes)
- git commit
- git push -f

So now the change just added --exit-non-zero-on-fix

boxydog:
A test is still failing. It's only failing on windows, an environment I don't have set up.

It's this:

```
E   AssertionError: 0 != 1 : expected 1 occurrences of '_old_attribute has been deprecated since 3.1.0 and will be removed by version 4.1.3.  Use _new_attribute instead', but found 0
```

boxydog:
In this PR, ruff did change `deprecated_attribute`, but only the type hinting:

image

So I don't understand why this change would break the test.

boxydog:
You might try just re-running those tests? I don't think I have permissions to do it.

boxydog:
Given that the re-running of tests now failed on macos, which previously succeeded, that makes me think it's a flaky test. Perhaps the behavior of that deprecated_attribute decorator depends on import order?

boxydog:
I'm going to create another branch that's not in this pull request and try a few variants to see if I can probe more precisely what's failing. For one thing, the test uses a method without `self`, maybe that's irritating something.

boxydog:
https://github.com/boxydog/pelican/pull/2 is a pull request with only changes to pyproject.toml, no python changes.

It still has that failing test (`test_deprecated_attribute`).

That says to me that it's a flaky test.  I am not sure why it's flaky, but I'm surprised it has not failed before.

boxydog:
https://github.com/boxydog/pelican/pull/3 is a pull request that just adds one blank line to pyproject.toml, and it has the same flaky test failing.

boxydog:
https://github.com/boxydog/pelican/actions/runs/9239087370/job/25417800210?pr=4 reveals a bit more of the story (from https://github.com/boxydog/pelican/pull/4). I print some debugging to stderr:

```
BEFORE value = self._old_attribute
top of deprecated_attribute._warn
bottom of deprecated_attribute._warn with message: _old_attribute has been deprecated since 3.1.0 and will be removed by version 4.1.3.  Use _new_attribute instead.
AFTER value = self._old_attribute
count_logs msg '_old_attribute has been deprecated since 3.1.0 and will be removed by version 4.1.3.  Use _new_attribute instead' level '30' count 0 all messages: []
```

This shows:

1. deprecated_attribute._warn is running, and is adding that message at the right time (between BEFORE and AFTER)
2. "messages", the captured log messages in LogCountHandler, is empty ("[]") when checked in count_logs.

I don't understand why that should be, especially since the test often succeed. What about LogCountHandler is flaky? This is a subtle bug somewhere.

LogCountHandler is a pelican class in support.py, added as a handler in pelican's LoggedTestCase. LogCountHandler derives from logging.handlers.BufferingHandler, a python class. So BufferingHandler is sometimes not receiving the log messages?

Is there something about the order the tests are called in that makes adding the handler not work? Help?

boxydog:
I will try rewriting the unit test to use mocks instead, if that's acceptable.

justinmayer:
No objection from my side. Thank you for digging into that, @boxydog! 😊

stanislavlevin:
`test_deprecated_attribute` is not flaky, it started failing with `pytest 8.2.0` (specifically https://github.com/pytest-dev/pytest/pull/12089).

With that change **collection** of the aforementioned test triggers the warning:
```
(py3) [builder@localhost pelican]$ python3 -m pytest pelican/tests/test_utils.py::TestUtils::test_deprecated_attribute --log-level=warning --log-cli-level=warning --collect-only
================================ test session starts =================================
platform linux -- Python 3.12.2, pytest-8.2.0.dev6+g1a5e0eb71, pluggy-1.5.0
rootdir: /usr/src/RPM/BUILD/pelican
configfile: tox.ini
plugins: anyio-4.4.0, cov-5.0.0, xdist-3.6.1
collecting ... 
-------------------------------- live log collection ---------------------------------
WARNING  pelican.utils:utils.py:215 _old_attribute has been deprecated since 3.1.0 and will be removed by version 4.1.3.  Use _new_attribute instead.
collected 1 item                                                                     

<Dir pelican>
  <Package pelican>
    <Package tests>
      <Module test_utils.py>
        <UnitTestCase TestUtils>
          <TestCaseFunction test_deprecated_attribute>

============================= 1 test collected in 0.20s ==============================
```

and during actual testing the warning **is not emitted second time** because of `LimitFilter` that `Remove duplicates records`.

Something like this helps:
```diff
diff --git a/pelican/tests/test_utils.py b/pelican/tests/test_utils.py
index 22dd8e38..bfdcc50a 100644
--- a/pelican/tests/test_utils.py
+++ b/pelican/tests/test_utils.py
@@ -23,9 +23,18 @@ from pelican.tests.support import (
 from pelican.writers import Writer
 
 
-class TestUtils(LoggedTestCase):
+class ClassDeprAttr:
     _new_attribute = "new_value"
 
+    @utils.deprecated_attribute(
+        old="_old_attribute", new="_new_attribute", since=(3, 1, 0), remove=(4, 1, 3)
+    )
+    def _old_attribute():
+        return None
+
+
+class TestUtils(LoggedTestCase):
+
     def setUp(self):
         super().setUp()
         self.temp_output = mkdtemp(prefix="pelicantests.")
@@ -34,15 +43,10 @@ class TestUtils(LoggedTestCase):
         super().tearDown()
         shutil.rmtree(self.temp_output)
 
-    @utils.deprecated_attribute(
-        old="_old_attribute", new="_new_attribute", since=(3, 1, 0), remove=(4, 1, 3)
-    )
-    def _old_attribute():
-        return None
-
     def test_deprecated_attribute(self):
-        value = self._old_attribute
-        self.assertEqual(value, self._new_attribute)
+        test_class = ClassDeprAttr()
+        value = test_class._old_attribute
+        self.assertEqual(value, test_class._new_attribute)
         self.assertLogCountEqual(
             count=1,
             msg=(
```

boxydog:
@justinmayer - #3310 indeed passes tests, so perhaps you can merge that unit test fix, then we can merge it in here, and get this PR in?

justinmayer:
I merged the test fix into `master`, merged those changes into this branch, and now all tests pass.

Any objections if I squash all the commits in this PR into a single commit before merging this PR?

Assuming there are no objections, after merging this PR, I will add that commit hash to [`.git-blame-ignore-revs`](https://github.com/getpelican/pelican/blob/master/.git-blame-ignore-revs) so the code style changes do not show up in `git blame` views.

Which reminds me… In the future we should try to split code formatting PRs into two commits: one for the code style configuration changes, and the other for the application of that new configuration to the code itself. That way, only the latter commit's hash can be added to [`.git-blame-ignore-revs`](https://github.com/getpelican/pelican/blob/master/.git-blame-ignore-revs), allowing the code style configuration changes to remain visible in `git blame` views.

justinmayer:
Thanks for sharing your thoughts. I usually prefer a small number of commits, squashing to a single commit when appropriate, but I sometimes like to ask folks (particularly when there's been a lot of effort invested) if they have preferences before merging.

I do my best to keep things clean but am not a Git-history-purist. I do prefer to keep `git blame` as functional as possible, however.

I agree that speed is helpful and do my best in that regard. That is sometimes hampered by (1) how much time I can devote to open source, and (2) how long it takes to get feedback from other folks, as I prefer to have something resembling consensus rather than making all decisions unilaterally. I realize that last point can inhibit speed and try my best to strike a good balance.

justinmayer:
I added the relevant commit hash to `.git-blame-ignore-revs` via: c46063cf

boxydog:
> I agree that speed is helpful and do my best in that regard.

I agree that open source is a complicated balance of many factors. I'm glad you are still shepherding pelican. Some open source maintainers just give up because it's too much time and hassle for the return.

boxydog:
Closing this because #3311 did the same thing, but with git commits the way Justin wanted.

PR #3241, opened 2023-11-12
Upgrade code base for Python 3.8 and above

justinmayer:
This upgrades the Pelican code base for Python 3.8 and above. Performed via:

	pipx run pyupgrade --py38-plus pelican/**/*.py

### DIFF
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -273,7 +273,7 @@ class PrintSettings(argparse.Action):
                         )
                     )
                 else:
-                    console.print("\n{} is not a recognized setting.".format(setting))
+                    console.print(f"\n{setting} is not a recognized setting.")
                     break
         else:
             # No argument was given to --print-settings, so print all settings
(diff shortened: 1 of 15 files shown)


pauloxnet (approved):
LGTM

justinmayer:
Follow-up question… Is this the kind of commit that should be added to [`.git-blame-ignore-revs`](https://github.com/getpelican/pelican/blob/master/.git-blame-ignore-revs)? Or is there some reason to have these changes show up in `git blame`?

pauloxnet:
> Follow-up question… Is this the kind of commit that should be added to [`.git-blame-ignore-revs`](https://github.com/getpelican/pelican/blob/master/.git-blame-ignore-revs)? Or is there some reason to have these changes show up in `git blame`?

In other similar projects (e.g. Django) is common to ignore this type of commit.

avaris (approved):
Looks good. I am fine with excluding this from blame.

pauloxnet (approved):

justinmayer:
@pauloxnet & @avaris: Thanks for having a look!

PR #3246, opened 2023-11-13
Add `tzdata` requirement on Windows

MinchinWeb:
Further to https://github.com/getpelican/pelican/pull/3161, `tzdata` is required on Windows, even when using a version of Python that includes `zoneinfo` in the standard library. This is because Windows doesn't ship with the "standard" timezone database; c.f. [PEP 615](https://peps.python.org/pep-0615/#sources-for-time-zone-data).

Without this, Pelican 4.9.0 will spit out a series of errors (one for each entry), like:

```
[10:53:22] ERROR    Could not process .\20121024-push-a-hg-repo-to-github.md                                   log.py:94
                    'No time zone found with key America/Edmonton'
```

and then finally has a fatal error:

```
           CRITICAL ZoneInfoNotFoundError: 'No time zone found with key UTC'                             __init__.py:666
```

I also note that `tzdata` is listed as a requirement for [testing](https://github.com/getpelican/pelican/blob/master/requirements/test.pip#L6); it may no longer be needed there.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+Release type: patch
+
+* Ensure ``tzdata`` dependency is installed on Windows
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,6 +40,7 @@ dependencies = [
     "unidecode>=1.3.7",
     "backports-zoneinfo>=0.2.1; python_version < \"3.9\"",
     "watchfiles>=0.21.0",
+    "tzdata; sys_platform == 'win32'",
 ]
 
 [project.optional-dependencies]
--- a/requirements/test.pip
+++ b/requirements/test.pip
@@ -3,7 +3,6 @@ Pygments==2.14.0
 pytest
 pytest-cov
 pytest-xdist[psutil]
-tzdata
 
 # Optional Packages
 Markdown==3.5.1


avaris:
> I also note that tzdata is listed as a requirement for [testing](https://github.com/getpelican/pelican/blob/master/requirements/test.pip#L6); it may no longer be needed there.

I was confused a bit. I was going to say it should show up in tests but that's why tests were passing :). I think it should be removed from `test.pip`.

MinchinWeb:
> I think it should be removed from `test.pip`.

done!

MinchinWeb:
FYI, here's where `tzdata` was added to the testing requirements: https://github.com/getpelican/pelican/commit/91d9ef7a7085bf9cf8aa341f23236970e9c6c2b6

>  Add tzdata as dependency in test requirements
>
> Otherwise yields the following error with Python 3.10 on Windows:
>
> zoneinfo._common.ZoneInfoNotFoundError: 'No time zone found with key UTC'

avaris (approved):

justinmayer (approved):
Sorry for this snafu. When I first saw this error, I read something on the Internet that gave me the (clearly-mistaken) impression that the `tzdata` dependency was something that only needed to be addressed when _testing_ on Windows and that was not otherwise necessary to _use_ Pelican on Windows. Mea culpa.

Many thanks to @MinchinWeb for the fix and to @avaris for reviewing.

MinchinWeb:
No worries @justinmayer ! You and avaris do a great job of keeping the project moving forward, and make it easy to contribute, and this is still a timely fix. Mostly I just got a little chuckle that I wasn't able to help with your 3.9 release sprint, but instead found this the next day.

Issue #3244, opened 2023-11-12
Pelican Webassets Cache Breaks Github Pages Deployment (File Permissions)


vbharadwaj-bk:
- [x] I have read the [Filing Issues](https://docs.getpelican.com/en/latest/contribute.html#filing-issues) and subsequent “How to Get Help” sections of the documentation.
- [x] I have searched the [issues](https://github.com/getpelican/pelican/issues?q=is%3Aissue) (including closed ones) and believe that this is not a duplicate.

- **OS version and name**:  Ubuntu
- **Python version**: 3.10
- **Pelican version**: 3.9
- **Link to theme**: Base
- **Links to plugins**: [pelican-webassets](https://pypi.org/project/pelican-webassets/)

## Issue
Plugins such as Pelican-webassets create files in the output directory that do not have appropriate file permissions when deploying to Github pages (specifically in the folder `output/theme/.webassets-cache`). See this page for the required permissions for the "Upload Artifact" action <https://github.com/actions/upload-pages-artifact#file-permissions>). This causes the Github pages upload action to fail with the message "Check file permissions". The relevant action from the Pelican GH-pages workflow is below.
```
      - name: Upload artifact
        uses: actions/upload-pages-artifact@v2
        with:
          path: ${{ inputs.output-path }}
```
It's easy to fix this by removing the offending files (which probably shouldn't be deployed as part of the website anyway), but a more general fix suggested by the "Upload Artifact" page is to add the action
```
      - name: Fix permissions
        run: |
          chmod -c -R +rX "output/" | while read line; do
            echo "::warning title=Invalid file permissions automatically fixed::$line"
          done
```
which changes the required file permissions (`output` should be replaced with the Pelican output directory). I can create a pull request that adds this action to the workflow script if there is enough interest. On the other hand, this issue was a useful signal that there were extraneous files to delete, and it may be worth just adding a note to the webassets repo README warning about this issue. Thoughts welcome.

justinmayer:
@seanh: As the author of the PR (#3189) that added the GitHub Pages deployment workflow, do you have any thoughts about this issue?

avaris:
Fix looks fine to me.

> it may be worth just adding a note to the webassets repo

I didn't test this but, it looks configurable via [`cache`](https://webassets.readthedocs.io/en/latest/environment.html#webassets.env.Environment.cache) option:

```python
WEBASSETS_CONFIG = [
    ("cache", False),  # disable, or
[...]

PR #3248, opened 2023-11-18
Add GH pages action to fix file permissions

vbharadwaj-bk:
# Pull Request Checklist

Resolves: #3244 

- [ ] Added **tests** for changed code (verified by building a test site and checking the GH pages workflow)
- [ ] Updated **documentation** for changed code (N/A)

Adds a Github action that fixes file permissions before uploading (such as cache files created by Pelican Webassets if the cache is not disabled). Raises a warning for each file without proper permissions.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,3 +1,0 @@
-Release type: patch
-
-Added Github pages action to fix files with invalid permissions in output file directory.
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+Release type: patch
+
+Added Github pages action to fix files with invalid permissions in output file directory.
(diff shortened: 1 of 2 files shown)


RELEASE.md
```diff
@@ -0,0 +1,3 @@
+Release type: patch
```
avaris, line 1: I don't think there is need for a release. This workflow lives only on github.

justinmayer: I came here to say the same thing and agree that the release file should be removed from this PR.

vbharadwaj-bk: Understood, file removed.

seanh:
This fix seems fine to me 👍 If upload-pages-artefact chokes on certain file permissions in the output directory, I see no reason why the workflow shouldn't fix those permissions to ensure that doesn't happen

avaris (approved):

justinmayer (approved):
Thanks to @vbharadwaj-bk for the enhancement and to @avaris and @seanh for reviewing. 🎖️

PR #3249, opened 2023-11-22
Feature: Add setting to append `ref` parameter to links in feeds

GiovanH:
This is a feature I'm experimenting with for personal use, as it's useful for analytics purposes to see if people are using the rss feeds.

This is more of a proposal at this stage. Is this something people would be interested in?


GiovanH:
Updated with documentation.

avaris (approved):

GiovanH:
@justinmayer Is there any further work needed here or is this ready to merge?

justinmayer (changes requested):
Hi @GiovanH. My apologies for the delay in reviewing. I've looked at this several times, but something always felt… not quite right… but I could never put my finger on _what_ didn’t feel right.

I think I figured it out:

1. The [`rel` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel) is for use within HTML documents. I don't think I have ever seen a `?rel=[…]` URL parameter, and I can’t find any documentation about such a parameter anywhere on the Internet. I have, on the other hand, seen `?ref=[…]` URL parameters used to indicating referrer information. Perhaps you meant `ref` and not `rel`? If so, perhaps change the URL parameter, and also change the setting name to `FEED_APPEND_REFERRER`?
2. For the parameter _value_, I suggest using the generic `feed` instead of `rss`, as a feed can be in either Atom or RSS format.

If I am wrong on any of the above counts, by all means feel free to point me to information to the contrary! 😊

GiovanH:
I agree with you on both counts.

justinmayer (approved):
Thank you for the feature contribution, @GiovanH! ✨

Issue #3037, opened 2022-09-05
Warning and Error logs messages show wrong source


copperchin:
- [x] I have read the [Filing Issues](https://docs.getpelican.com/en/latest/contribute.html#filing-issues) and subsequent “How to Get Help” sections of the documentation.
- [x] I have searched the [issues](https://github.com/getpelican/pelican/issues?q=is%3Aissue) (including closed ones) and believe that this is not a duplicate.

- **OS version and name**: `Ubuntu 18.04.4 LTS (bionic)`  ; via WSL 1 on Windows 10
- **Python version**:  `3.8.0`
- **Pelican version**:  `4.8.0`

## Issue
Any time an error or warning message is logged, the logger reports the message as coming from the log.py instead of the true source of the message.

Very simple test is to add the following to ``pelican.conf``:
```python
import logging
logging.warning("This is a warning")
logging.error("This is an error.")
```

Then run the pelican listener in the terminal:
`pelican -l`

which yields:
```
[10:24:40] WARNING  This is a warning                                       log.py:91
           ERROR    This is an error.                                       log.py:96
```

whereas I'd expect this:
```
[10:19:03] WARNING  This is a warning                                       pelicanconf.py:2
           ERROR    This is an error.                                       pelicanconf.py:3
```

This appears to be because `FatalLogger` overrides the Logger ``error`` and ``warning`` methods: first checking whether to raise a RuntimeError, then making a call to the log method with ``super()``.

I'm proposing a fix which would work by making the fatal check in the logger handler, rather than redefining the Logger's logging methods.

justinmayer:
This issue should be fixed by the above-linked PR. Thanks again, @MinchinWeb and @copperchin!

PR #3257, opened 2023-11-26
Log the original calling location, rather than the wrapper function

MinchinWeb:
I've noticed that when Pelican logs warnings and errors, the file number and line number are always the same. It turns out we wrap the calling of warnings and errors (but not other logging levels) to allow warnings to be fatal, etc, but it is thus reporting the location of that wrapper function, rather than the "true" caller.

This isn't as helpful as it could be when trying to understand what is going on under the covers.

This is a small patch that allows the logger to show the "true" caller.

Before:

```
[20:40:37] WARNING  [Seafoam] Development Mode is active. Image Process has been disabled.                     log.py:94
           WARNING  AUTHOR_URL is set to None                                                                  log.py:94
Done: Processed 4 articles, 0 drafts, 0 hidden articles, 0 pages, 0 hidden pages and 0 draft pages in 0.51 seconds.
```

After:

```
[20:51:24] WARNING  [Seafoam] Development Mode is active. Image Process has been disabled.              initialize.py:51
           WARNING  AUTHOR_URL is set to None                                                         urlwrappers.py:116
Done: Processed 4 articles, 0 drafts, 0 hidden articles, 0 pages, 0 hidden pages and 0 draft pages in 0.48 seconds.
```

The required feature was added in Python 3.8.
c.f. https://stackoverflow.com/a/55998744/4276230

### DIFF
--- a/pelican/log.py
+++ b/pelican/log.py
@@ -85,13 +85,15 @@ class FatalLogger(LimitLogger):
     warnings_fatal = False
     errors_fatal = False
 
+    # adding `stacklevel=2` means that the displayed filename and line number
+    # will match the "original" calling location, rather than the wrapper here
     def warning(self, *args, **kwargs):
-        super().warning(*args, **kwargs)
+        super().warning(*args, stacklevel=2, **kwargs)
         if FatalLogger.warnings_fatal:
             raise RuntimeError("Warning encountered")
 
     def error(self, *args, **kwargs):
-        super().error(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


MinchinWeb:
This would fix #3037. C.f. #3038.

boxydog:
Cool.

justinmayer (approved):

avaris (approved):

justinmayer:
@MinchinWeb: Many thanks for this fix! 👏

PR #3258, opened 2023-11-26
fix: use relative urls when github publish enabled

snosratiershad:
# Pull Request Checklist

I noticed without using relative paths GitHub can't fetch static files. so I enabled it when github config is enabled.


snosratiershad:
rebased with main. so there is no conflict.

PR #3266, opened 2023-12-12
Better logging if writing a tag/category/author page fails

minchinweb:
Add a short logging message if for some reason an invalid tag is provided, and Pelican instead tries to overwrite the index page.

Before:

```
           CRITICAL RuntimeError: File D:/Code/obsidian-vault/zz_output/tags/index.html is to be         __init__.py:666
                    overwritten
```

After:

```
[21:56:21] ERROR    Trying to write tag page for "**".                                                 generators.py:580
           CRITICAL RuntimeError: File D:/Code/obsidian-vault/zz_output/tags/index.html is to be         __init__.py:666
                    overwritten
```

(For some reason, the tag `**` is mapped onto the index page. Perhaps that is another issue that should be fixed...)

Fix is applied when writing Tag, Category, and Author pages. Let me know if it should be added other places.

# Pull Request Checklist

- [-] Updated **documentation** for changed code

### DIFF
--- a/pelican/generators.py
+++ b/pelican/generators.py
@@ -586,9 +586,10 @@ class ArticlesGenerator(CachingGenerator):
                 )
             except RuntimeError:
                 if not tag.slug:
-                    logger.warning(
+                    logger.info(
                         'Tag "%s" has an invalid slug; skipping writing tag page...',
                         tag,
+                        extra={"limit_msg": "Further tags with invalid slugs."},
                     )
                     continue
                 else:
@@ -616,9 +617,10 @@ class ArticlesGenerator(CachingGenerator):
                 )
             except RuntimeError:
                 if not cat.slug:
-                    logger.warning(
+                    logger.info(
                         'Category "%s" has an invalid slug; skipping writing category page...',
                         cat,
+                        extra={"limit_msg": "Further categories with invalid slugs."},
                     )
                     continue
                 else:
@@ -646,9 +648,10 @@ class ArticlesGenerator(CachingGenerator):
                 )
             except RuntimeError:
                 if not aut.slug:
-                    logger.warning(
+                    logger.info(
                         'Author "%s" has an invalid slug; skipping writing author page...',
                         aut,
+                        extra={"limit_msg": "Further authors with invalid slugs."},
                     )
                     continue
                 else:
--- a/pelican/generators.py
+++ b/pelican/generators.py
@@ -570,57 +570,69 @@ class ArticlesGenerator(CachingGenerator):
         tag_template = self.get_template("tag")
         for tag, articles in self.tags.items():
             dates = [article for article in self.dates if article in articles]
-            write(
-                tag.save_as,
-                tag_template,
-                self.context,
-                tag=tag,
-                url=tag.url,
-                articles=articles,
-                dates=dates,
-                template_name="tag",
-                blog=True,
-                page_name=tag.page_name,
-                all_articles=self.articles,
-            )
+            try:
+                write(
+                    tag.save_as,
+                    tag_template,
+                    self.context,
+                    tag=tag,
+                    url=tag.url,
+                    articles=articles,
+                    dates=dates,
+                    template_name="tag",
+                    blog=True,
+                    page_name=tag.page_name,
+                    all_articles=self.articles,
+                )
+            except RuntimeError as e:
+                logger.error('Trying to write Tag page for "%s".' % (tag))
+                raise e
 
     def generate_categories(self, write):
         """Generate category pages."""
         category_template = self.get_template("category")
         for cat, articles in self.categories:
             dates = [article for article in self.dates if article in articles]
-            write(
-                cat.save_as,
-                category_template,
-                self.context,
-                url=cat.url,
-                category=cat,
-                articles=articles,
-                dates=dates,
-                template_name="category",
-                blog=True,
-                page_name=cat.page_name,
-                all_articles=self.articles,
-            )
+            try:
+                write(
+                    cat.save_as,
+                    category_template,
+                    self.context,
+                    url=cat.url,
+                    category=cat,
+                    articles=articles,
+                    dates=dates,
+                    template_name="category",
+                    blog=True,
+                    page_name=cat.page_name,
+                    all_articles=self.articles,
+                )
+            except RuntimeError as e:
+                logger.error('Trying to write Category page for "%s".' % (cat))
+                raise e
 
     def generate_authors(self, write):
         """Generate Author pages."""
         author_template = self.get_template("author")
         for aut, articles in self.authors:
             dates = [article for article in self.dates if article in articles]
-            write(
-                aut.save_as,
-                author_template,
-                self.context,
-                url=aut.url,
-                author=aut,
-                articles=articles,
-                dates=dates,
-                template_name="author",
-                blog=True,
-                page_name=aut.page_name,
-                all_articles=self.articles,
-            )
+            try:
+                write(
+                    aut.save_as,
+                    author_template,
+                    self.context,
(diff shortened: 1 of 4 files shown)


pelican/generators.py
```diff
@@ -560,57 +560,69 @@ def generate_tags(self, write):
...
+                    articles=articles,
+                    dates=dates,
+                    template_name="tag",
+                    blog=True,
+                    page_name=tag.page_name,
+                    all_articles=self.articles,
+                )
+            except RuntimeError as e:
+                logger.error('Trying to write Tag page for "%s".' % (tag))
+                raise e
```
avaris, line 579: * It's not "trying" anymore. It tried and now failed, maybe `"Unable to write..."` instead
* Please don't format logs in place (see: [docs](https://docs.getpelican.com/en/latest/contribute.html#logging-tips)): `logger.error("...", tag)`
* You don't do anything with exception (`as e`), so same as:
```python
try:
    # ...
except RuntimeError:
    # ...
    raise
```

Same for the others.

avaris:
> For some reason, the tag ** is mapped onto the index page.

Because, slug is empty and depending on `TAG_URL/TAG_SAVE_AS` you'll get "weird" results:

```python
>>> from pelican.urlwrappers import Tag
>>> from pelican.settings import DEFAULT_CONFIG
>>> t = Tag('**', DEFAULT_CONFIG)
>>> t.slug
''
```

And that's because `slugify` will strip/replace "questionable" characters. Since this will become part of the filename in the output, and filesystems can be picky.

Off the top of my head, I cannot think of any reason for wanting an empty slug intentionally. So if no objections, we could error for empty slug.

MinchinWeb:
I've updated it as per your comments! This is the current message:

```
[20:13:46] WARNING  Tag "**" has an invalid slug; skipping writing tag page...                                 log.py:89
```

I have it throw a warning (rather than a critical error) if the slug is invalid, but throws the "regular" critical error if it's run into some other problem.

I've "un- pre- formatted" the log messages, and dropped the `as e`.

Overall, it makes sense that slugify would strip out questionable character; I was just annoyed that Pelican would break but not really tell me what was wrong or how to fix it; I hope this explains how to fix the issue to the next user.

pelican/generators.py
```diff
@@ -560,57 +560,90 @@ def generate_tags(self, write):
...
+                    page_name=tag.page_name,
+                    all_articles=self.articles,
+                )
+            except RuntimeError:
+                if not tag.slug:
+                    logger.warning(
+                        'Tag "%s" has an invalid slug; skipping writing tag page...',
+                        tag,
+                    )
+                    continue
```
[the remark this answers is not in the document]
MinchinWeb: So during Tag creation?

This issue came up for me while working on a replacement Markdown reader, so I wouldn't want it to be tied directly to the reader, but when tags are created could work.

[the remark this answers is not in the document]
avaris: Yes, e.g. here at the end: https://github.com/getpelican/pelican/blob/2d75ca8391bccb79bdac3fc7fd609a3a337e6fd8/pelican/urlwrappers.py#L13-L17

Something like:
```python
if not self.slug:
    logger.warning("%s `%s` will result in empty slug", self.__class__.__name__, self._name)
```

PS: Changed my mind and lowered it to a warning. Probably doesn't need to be an error.

MinchinWeb:
I've added a warning when trying to generate a slug from a name that won't give a valid slug, and downgraded the former warning to an INFO message. So the displayed messages (when run with `--verbose`) now look like:

```
           WARNING  Unable to generate valid slug for Tag "#**".                                               log.py:89
[...]
           INFO     Tag "#**" has an invalid slug; skipping writing tag page...                        generators.py:579
```

pelican/generators.py
```diff
@@ -560,57 +560,93 @@ def generate_tags(self, write):
...
+                    all_articles=self.articles,
+                )
+            except RuntimeError:
+                if not tag.slug:
+                    logger.info(
+                        'Tag "%s" has an invalid slug; skipping writing tag page...',
+                        tag,
+                        extra={"limit_msg": "Further tags with invalid slugs."},
+                    )
+                    continue
```
avaris, line 584: With empty slugs already generating warnings, is this special case necessary anymore? Why not just do the `else` block?

MinchinWeb: I think it's helpful to still keep this, because it makes it very explicity what Pelican is (or isn't doing), and it's not entirely obvious that an invalid slug would keep the page from being generated.

Also, it mirrors the logging for all the pages that *are* written.

MinchinWeb:
Hi @avaris !

This seems to have been lost in the shuffle.

Do you want me to remove the logging (of the non-generated pages), or is it good to merge as is?

Thanks!

MinchinWeb:
Hey @avaris ! Let me know if anything more is needed to merge this! Thanks!

avaris:
Apologies, offline life get in the way of things. I'll try to get back to this ASAP.

MinchinWeb:
> Apologies, offline life get in the way of things. I'll try to get back to this ASAP.

No worries at. My offline life has kept me from anything Pelican-related for the last couple months too...

minchinweb:
Hey @avaris , just wanted to check in on this. I'm maintaining a local fork with this on my side, but would love to just see it merged in.... Thanks for your help.

minchinweb:
Just wanted to bump this, hoping to get it merged! It's one of the patches I'm maintaining locally, and I was going to get up a new pull request till I remembered this is still here waiting...

Let me know if changes are needed; I'm happy to make them.

pelican/generators.py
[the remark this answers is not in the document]
minchinweb: Done! (Though I called the exception `FileOverwriteFailedError` because I thought that would be clearer)

minchinweb:
(rebased to resolve merge conflicts)

pelican/writers.py
```diff
@@ -107,14 +111,16 @@ def _open_w(self, filename, encoding, override=False):
         """
         if filename in self._overridden_files:
             if override:
-                raise RuntimeError(f"File {filename} is set to be overridden twice")
+                raise FileOverwriteFailedError(
+                    f"File {filename} is set to be overridden twice"
+                )
```
justinmayer, line 117: I have never fully understood the language in this error output. I know you weren't the person who authored this language, but could you take a moment and rewrite this error message such that it can be better understood by end users? (Including using "overwritten" instead of what I believe to be an incorrect usage here of "overridden".)

pelican/writers.py
```diff
@@ -107,14 +111,16 @@ def _open_w(self, filename, encoding, override=False):
...
             logger.info("Skipping %s", filename)
             filename = os.devnull
         elif filename in self._written_files:
             if override:
                 logger.info("Overwriting %s", filename)
             else:
-                raise RuntimeError(f"File {filename} is to be overwritten")
+                raise FileOverwriteFailedError(f"File {filename} is to be overwritten")
```
justinmayer: Same idea here: Could you take a moment and rewrite this error message such that it can be better understood by end users?

justinmayer (commented):
Hey @minchinweb. At the risk of prolonging this further, while you are here and editing these particular lines, could I possibly trouble you to rewrite two of the error messages, with full awareness that you were not the person who originally wrote them? I'd really appreciate it! 😊

pelican/writers.py
minchinweb: I'll take a crack at this in the morning!

minchinweb:
@justinmayer Error messages rewritten! (Overwritten?)

justinmayer (approved):
Many thanks for all your work on this, @minchinweb! 🌟

PR #3276, opened 2024-01-14
Better error logging if a plugin refuses to load

MinchinWeb:
Further to https://github.com/getpelican/pelican/pull/3257, this will print the proper line number if a plugin fails to load, and will print the stacktrace, if *pelican* is run with `--debug`.

Useful when you're first trying to get a plugin to load, or have issue when Pelican first imports/loads the plugin.

### DIFF
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -80,8 +80,14 @@ class Pelican:
                 plugin.register()
                 self.plugins.append(plugin)
             except Exception as e:
-                logger.error("Cannot register plugin `%s`\n%s", name, e, stacklevel=3)
-                print(e.stacktrace)
+                logger.error(
+                    "Cannot register plugin `%s`\n%s",
+                    name,
+                    e,
+                    stacklevel=2,
+                )
+                if self.settings.get("DEBUG", False):
+                    console.print_exception()
 
         self.settings["PLUGINS"] = [get_plugin_name(p) for p in self.plugins]
 
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -80,7 +80,8 @@ class Pelican:
                 plugin.register()
                 self.plugins.append(plugin)
             except Exception as e:
-                logger.error("Cannot register plugin `%s`\n%s", name, e)
+                logger.error("Cannot register plugin `%s`\n%s", name, e, stacklevel=3)
+                print(e.stacktrace)
 
         self.settings["PLUGINS"] = [get_plugin_name(p) for p in self.plugins]
 
--- a/pelican/log.py
+++ b/pelican/log.py
@@ -88,12 +88,16 @@ class FatalLogger(LimitLogger):
     # adding `stacklevel=2` means that the displayed filename and line number
     # will match the "original" calling location, rather than the wrapper here
     def warning(self, *args, **kwargs):
-        super().warning(*args, stacklevel=2, **kwargs)
+        if "stacklevel" not in kwargs.keys():
+            kwargs["stacklevel"] = 2
+        super().warning(*args, **kwargs)
         if FatalLogger.warnings_fatal:
             raise RuntimeError("Warning encountered")
 
     def error(self, *args, **kwargs):
-        super().error(*args, stacklevel=2, **kwargs)
+        if "stacklevel" not in kwargs.keys():
+            kwargs["stacklevel"] = 2
+        super().error(*args, **kwargs)
         if FatalLogger.errors_fatal:
             raise RuntimeError("Error encountered")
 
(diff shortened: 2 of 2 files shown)


pelican/__init__.py
```diff
@@ -80,7 +80,8 @@ def init_plugins(self):
                 plugin.register()
                 self.plugins.append(plugin)
             except Exception as e:
-                logger.error("Cannot register plugin `%s`\n%s", name, e)
+                logger.error("Cannot register plugin `%s`\n%s", name, e, stacklevel=3)
+                print(e.stacktrace)
```
avaris, line 84: No to `print`. Pass `exc_info` to `logger.error` ([example](https://github.com/getpelican/pelican/blob/5d3e87b50be0c84d272e7fb2f9fb794af7887b50/pelican/generators.py#L873C27-L878))

pelican/log.py
```diff
@@ -88,12 +88,16 @@ class FatalLogger(LimitLogger):
     # adding `stacklevel=2` means that the displayed filename and line number
     # will match the "original" calling location, rather than the wrapper here
     def warning(self, *args, **kwargs):
-        super().warning(*args, stacklevel=2, **kwargs)
+        if "stacklevel" not in kwargs.keys():
+            kwargs["stacklevel"] = 2
+        super().warning(*args, **kwargs)
```
avaris, line 93: We can make this "transparent", by effectively doing `stacklevel + 1`, right? `+1` to pass through the `super().warning(...)` part? Otherwise everyone needs to account for it, like you do above with `stacklevel = 3`. And if this is refactored or no longer overridden, everything needs to change.

I feel like there is a better way to do this. Something like:

```python
def warning(self, *args, stacklevel=1, **kwargs):
    kwargs['stacklevel'] = stacklevel + 1
    super().warning(*args, **kwargs)
```

should work. But I didn't test it and it's late so I might not be fully awake :).

MinchinWeb: Done!

pelican/__init__.py
MinchinWeb: So I dropped `print()`, and instead am conditionally using Rich's `console.print_exception()`. This provides a fair bit more context to what is happening. Is this an okay way to do it?

Using `exc_info`, you get something like this:

```
           ERROR    Cannot register plugin `minchin.pelican.readers.commonmark`                           __init__.py:66
                    module 'pelican.plugins.signals' has no attribute 'make_something_up'
                      File "D:\Code\pelican\pelican\__init__.py", line 80, in init_plugins
                        plugin.register()
                      File
                    "D:\Code\minchin.pelican.readers.commonmark\minchin\pelican\readers\commonmark\__init
                    __.py", line 19, in register
                        signals.make_something_up.connect(check_settings())
                        ^^^^^^^^^^^^^^^^^^^^^^^^^
                    AttributeError: module 'pelican.plugins.signals' has no attribute 'make_something_up'
```

With `console.print_exception()`, you get something like this:

```
           ERROR    Cannot register plugin `minchin.pelican.readers.commonmark`                           __init__.py:66
                    module 'pelican.plugins.signals' has no attribute 'make_something_up'
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ D:\Code\pelican\pelican\__init__.py:80 in init_plugins                                           │
│                                                                                                  │
│    77 │   │   │   name = get_plugin_name(plugin)                                                 │
│    78 │   │   │   logger.debug("Registering plugin `%s`", name)                                  │
│    79 │   │   │   try:                                                                           │
│ ❱  80 │   │   │   │   plugin.register()                                                          │
│    81 │   │   │   │   self.plugins.append(plugin)                                                │
│    82 │   │   │   except Exception as e:                                                         │
│    83 │   │   │   │   logger.error(                                                              │
│                                                                                                  │
│ D:\Code\minchin.pelican.readers.commonmark\minchin\pelican\readers\commonmark\__init__.py:19 in  │
│ register                                                                                         │
│                                                                                                  │
│   16 │   signals.initialized.connect(check_settings)                                             │
│   17 │   signals.initialized.connect(commonmark_version)                                         │
│   18 │   signals.readers_init.connect(add_commonmark_reader)                                     │
│ ❱ 19 │   signals.make_something_up.connect(check_settings())                                     │
│   20                                                                                             │
│   21                                                                                             │
│   22 from .constants import LOG_PREFIX                                                           │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
AttributeError: module 'pelican.plugins.signals' has no attribute 'make_something_up'
```

(both in colour)

avaris: > Is this an okay way to do it?

Yes-ish... Technically, both are more or less the same thing, but now it'll bug me that this is done differently in different places :). No big deal though.

avaris (approved):
Thanks!

justinmayer (approved):
Many thanks to @MinchinWeb for the enhancement and to @avaris for reviewing 💯

PR #3278, opened 2024-01-18
Add additional type hints

bjoernricks:
# Pull Request Checklist

Adding the types spotted issues with the `Contents.source_path` attribute if it's `None`. For example `relative_dir` property will crash in that case.

### DIFF
--- a/pelican/contents.py
+++ b/pelican/contents.py
@@ -6,7 +6,8 @@ import os
 import re
 from datetime import timezone
 from html import unescape
-from urllib.parse import unquote, urljoin, urlparse, urlunparse
+from typing import Any, Dict, Optional, Set, Tuple
+from urllib.parse import ParseResult, unquote, urljoin, urlparse, urlunparse
 
 try:
     from zoneinfo import ZoneInfo
@@ -15,7 +16,7 @@ except ModuleNotFoundError:
 
 
 from pelican.plugins import signals
-from pelican.settings import DEFAULT_CONFIG
+from pelican.settings import DEFAULT_CONFIG, Settings
 from pelican.utils import (
     deprecated_attribute,
     memoized,
@@ -44,12 +45,20 @@ class Content:
 
     """
 
+    default_template: Optional[str] = None
+    mandatory_properties: Tuple[str, ...] = ()
+
     @deprecated_attribute(old="filename", new="source_path", since=(3, 2, 0))
     def filename():
         return None
 
     def __init__(
-        self, content, metadata=None, settings=None, source_path=None, context=None
+        self,
+        content: str,
+        metadata: Optional[Dict[str, Any]] = None,
+        settings: Optional[Settings] = None,
+        source_path: Optional[str] = None,
+        context: Optional[Dict[Any, Any]] = None,
     ):
         if metadata is None:
             metadata = {}
@@ -156,10 +165,10 @@ class Content:
 
         signals.content_object_init.send(self)
 
-    def __str__(self):
+    def __str__(self) -> str:
         return self.source_path or repr(self)
 
-    def _has_valid_mandatory_properties(self):
+    def _has_valid_mandatory_properties(self) -> bool:
         """Test mandatory properties are set."""
         for prop in self.mandatory_properties:
             if not hasattr(self, prop):
@@ -169,7 +178,7 @@ class Content:
                 return False
         return True
 
-    def _has_valid_save_as(self):
+    def _has_valid_save_as(self) -> bool:
         """Return true if save_as doesn't write outside output path, false
         otherwise."""
         try:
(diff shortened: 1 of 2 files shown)


avaris:
Can I request that you combine these into a single PR? Commit for each file is fine but PR per file is a bit much :).

bjoernricks:
> Can I request that you combine these into a single PR? Commit for each file is fine but PR per file is a bit much :).

Of course. Just let me know how it work best for you. Most projects prefer smaller PRs because they are easier and faster to merge.

pelican/utils.py
```diff
@@ -600,7 +635,9 @@ def truncate_html_words(s, num, end_text="…"):
     return out
 
 
-def process_translations(content_list, translation_id=None):
+def process_translations(
+    content_list: List["Content"], translation_id: Union[str, Set[str], None] = None
+) -> Tuple[List["Content"], List["Content"]]:
```
avaris, line 640: Do these (`"Content"`) need to be strings? If I remember correctly, this is not needed anymore after python 3.7?

avaris (changes requested):

pelican/utils.py
bjoernricks: > Do these (`"Content"`) need to be strings? If I remember correctly, this is not needed anymore after python 3.7?

As far as I know this is implemented with [PEP 563: Postponed Evaluation of Annotations](https://docs.python.org/3/whatsnew/3.7.html#pep-563-postponed-evaluation-of-annotations) and still requires `from __future__ import annotations`.

bjoernricks: Indeed it requires `from __future__ import annotations`. Otherwise a `NameError` will be raised. And because `pelican.contents` imports `pelican.utils` we can't remove the `if TYPE_CHECKING` gate here. Otherwise we create a circular import.

pelican/contents.py
```diff
@@ -454,25 +463,25 @@ def _get_summary(self):
...
-    def url(self):
+    def url(self) -> str:
         return self.get_url_setting("url")
 
     @property
-    def save_as(self):
+    def save_as(self) -> str:
         return self.get_url_setting("save_as")
 
     def _get_template(self):
```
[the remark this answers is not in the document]
bjoernricks: Oh yes, just forgot to add the annotation. I wanted to check if this is a string or something like a jinja template first :)

pelican/contents.py
```diff
@@ -357,9 +366,9 @@ def _find_path(path):
         parts[2] = origin
         origin = urlunparse(parts)
 
-        return "".join((m.group("markup"), m.group("quote"), origin, m.group("quote")))
+        return "".join((m.group("markup"), m.group("quote"), origin, m.group("quote")))  # type: ignore
```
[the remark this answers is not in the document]
bjoernricks: Ah sorry, this was just for getting pyright quiet. It seem to have issues with the return type of `m.group`. Since neither mypy nor pyright are run on the CI I'll remove that statement.

pelican/utils.py
bjoernricks: I've accidentally pushed changes without the quotes. See https://github.com/getpelican/pelican/actions/runs/7623051907/job/20762296119 for the outcome

avaris: Ah yes, removing quotes would require `__future__` import. Personally, I prefer `__future__` and non-quoted version, for simply maintainability reasons. If that becomes unnecessary later, it's easier to remove one line of import, instead of individually removing quotes from multiple places.

For `Union[..., None]` and `Optional[...]`, I know they are same, but you used `Optional[...]` everywhere else so I suggested only to continue that for consistency :). I gave up on making type hints readable. I think they make code equally unreadable (or readable depending on how you look), beyond anything simple. So I don't have a strong preference either way, other than keeping the same style throughout the project.

bjoernricks: I've changed it to `translation_id: Optional[Union[str, Collection[str]]]`. The following code uses `len` on `translation_id`. Therefore it must be a child of `Sized` and `Iterable` doesn't fit. Normally I use `Sequence` in this case but I noticed that `set` isn't a `Sequence` and therefore re-assigning a set on line 664 wouldn't work. Instead a `Collection` works.

Additionally I've re-read [PEP 563](https://peps.python.org/pep-0563/) and [PEP 649](https://peps.python.org/pep-0649/) and using `from __future__ import annotations` should be fine here as dataclasses or pydantic aren't in use here. With Python >= 3.13 the import can be safely removed again.

avaris (approved):
Perfect, thank you. For the patience as well :).

justinmayer (approved):
Many thanks to @bjoernricks for the enhancement and to @avaris for reviewing.

PR #3279, opened 2024-01-21
Feeds - Update generators.py to fix a bug with slugs

nam-ori:
Update generators.py to fix a bug with slugs that stay unformatted in {category,tag,author,lang} feeds link

In the feeds XML output, specifically for category, tag, author and lang feeds, part "{slug}" of the URL was not formatted in feeds link (line: <link href="...{slug}..." rel="self"/>) as soon as a _URL parameter (like CATEGORY_FEED_ATOM_URL) was given in the settings.

This small commit will fix that, applying the format(slug) in all cases, whether _URL parameter is given or not.

No expected side effects, no problem with tests.

### DIFF
--- a/pelican/generators.py
+++ b/pelican/generators.py
@@ -382,9 +382,9 @@ class ArticlesGenerator(CachingGenerator):
                     arts,
                     self.context,
                     str(self.settings["CATEGORY_FEED_ATOM"]).format(slug=cat.slug),
-                    str(self.settings.get(
+                    self.settings.get(
                         "CATEGORY_FEED_ATOM_URL",
-                        self.settings["CATEGORY_FEED_ATOM"])
+                        str(self.settings["CATEGORY_FEED_ATOM"]),
                     ).format(slug=cat.slug),
                     feed_title=cat.name,
                 )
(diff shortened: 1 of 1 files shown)


pelican/generators.py
```diff
@@ -382,10 +382,10 @@ def generate_feeds(self, writer):
                     arts,
                     self.context,
                     str(self.settings["CATEGORY_FEED_ATOM"]).format(slug=cat.slug),
-                    self.settings.get(
+                    str(self.settings.get(
                         "CATEGORY_FEED_ATOM_URL",
-                        str(self.settings["CATEGORY_FEED_ATOM"]).format(slug=cat.slug),
-                    ),
+                        self.settings["CATEGORY_FEED_ATOM"])
+                    ).format(slug=cat.slug),
```
avaris, line 388: Nice catch. I agree `format` should be done at the end but technically `str(...)` is only needed for `CATEGORY_FEED_ATOM`, because it can be `pathlib.Path`:

```python
                    self.settings.get(
                        "CATEGORY_FEED_ATOM_URL",
                        str(self.settings["CATEGORY_FEED_ATOM"])
                    ).format(slug=cat.slug),
```

Similarly for the others.

avaris (changes requested):
Some minor changes. Also, it would be nice if there was a test added for it, but not a big deal for me.

justinmayer (approved):
Many thanks to @nam-ori for the fix and to @avaris for reviewing. 🎖️

Issue #3265, opened 2023-12-09
Internal URLs not parsed in summaries


arturlangner:
Internal links (ie. `{filename}/stuff/more_stuff.md`) are not parsed in a summary and the output HTML URL is broken. It contains the original verbatim text.

I experienced it when using https://github.com/MinchinWeb/minchin.pelican.plugins.summary but the bug itself is in Pelican codebase.

The problematic [code](https://github.com/getpelican/pelican/blob/master/pelican/contents.py#L432) in `get_summary()` is:
```
if "summary" in self.metadata:
            return self.metadata["summary"]
```

Fix that works for me is:
```
if "summary" in self.metadata:
            return self._update_content(self.metadata["summary"], siteurl)
```

I would be happy if this fix or something similar goes upstream. 🙂

avaris:
Your suggestion looks correct to me. PR would be welcome :).

justinmayer:
@arturlangner: Any follow-up comments about this issue that you submitted?

arturlangner:
I uninstalled and reinstalled Pelican 4.9.1 with `pip` in a virtualenv and I still have the issue. Output HTML in the **summaries** contains links with "raw" URLs like: `<a href="{filename}/articles/2021/stuff.md">stuff</a>`.

I checked my contents.py:527 and it does contain the code @avaris has mentioned above.

The only somewhat special thing about my setup is that I also use `minchin.pelican.plugins.summary` (1.2.1, also via `pip`).

justinmayer:
@avaris already said that he can't reproduce the issue you describe. Do you not think it would make more sense to try _without_ the `minchin.pelican.plugins.summary` plugin installed and see if the issue persists? Don't that seem like a troubleshooting step that should have been performed from the very beginning?

arturlangner:
I retested with the plugin disabled and the output is okay. The plugin only [extracts](https://github.com/MinchinWeb/minchin.pelican.plugins.summary/blob/128b9a93b8dddc06889bc3752af05b912316e7f4/minchin/pelican/plugins/summary/summary.py#L108) the summary and does not do any processing.

justinmayer:
So… If the problem goes away when the plugin is disabled, doesn't this seem like something to ask @MinchinWeb about since it appears related to the plugin instead of Pelican itself?

MinchinWeb:
@arturlangner Is your site setup somewhere online where I can look at it? In particular, it makes a difference what version of Pelican you're trying to run this with...

[...]

PR #3280, opened 2024-01-22
Resolve intrasite links in summaries.

MinchinWeb:
Closes [#3265](https://github.com/getpelican/pelican/issues/3265)
Closes https://github.com/MinchinWeb/minchin.pelican.plugins.summary/issues/5

Changes the order of when intrasite links are resolved, so it is after summary plugins are called.

Also allows `Articles` not to have a `_summary` attribute.

Related history includes:

- https://github.com/getpelican/pelican-plugins/issues/314
- https://github.com/getpelican/pelican/pull/1616
- https://github.com/getpelican/pelican-plugins/pull/410
- https://github.com/getpelican/pelican/pull/2226
- https://github.com/getpelican/pelican/pull/2288
- https://github.com/getpelican/pelican/issues/3265
- https://github.com/MinchinWeb/minchin.pelican.plugins.summary/issues/5

### DIFF
--- a/pelican/contents.py
+++ b/pelican/contents.py
@@ -520,12 +520,17 @@ class Content:
 
         # _summary is an internal variable that some plugins may be writing to,
         # so ensure changes to it are picked up
-        if (
-            "summary" in self.settings["FORMATTED_FIELDS"]
-            and "summary" in self.metadata
-        ):
-            self._summary = self._update_content(self._summary, self.get_siteurl())
-            self.metadata["summary"] = self._summary
+        if "summary" in self.settings["FORMATTED_FIELDS"]:
+            if hasattr(self, "_summary"):
+                self.metadata["summary"] = self._summary
+
+            if "summary" in self.metadata:
+                self.metadata["summary"] = self._update_content(
+                    self.metadata["summary"], self.get_siteurl()
+                )
+
+            if hasattr(self, "_summary") and "summary" in self.metadata:
+                self._summary = self.metadata["summary"]
 
 
 class Page(Content):
--- a/pelican/contents.py
+++ b/pelican/contents.py
@@ -519,7 +519,7 @@ class Content:
                 setattr(self, key.lower(), value)
 
         # _summary is an internal variable that some plugins may be writing to,
-        # so ensure changes to it are picked up
+        # so ensure changes to it are picked up, and write summary back to it
         if "summary" in self.settings["FORMATTED_FIELDS"]:
             if hasattr(self, "_summary"):
                 self.metadata["summary"] = self._summary
@@ -528,8 +528,6 @@ class Content:
                 self.metadata["summary"] = self._update_content(
                     self.metadata["summary"], self.get_siteurl()
                 )
-
-            if hasattr(self, "_summary") and "summary" in self.metadata:
                 self._summary = self.metadata["summary"]
 
 
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -120,11 +120,14 @@ class Pelican:
             if hasattr(p, "generate_context"):
                 p.generate_context()
 
+        # for plugins that create/edit the summary
+        logger.debug("Signal all_generators_finalized.send(<generators>)")
+        signals.all_generators_finalized.send(generators)
+
+        # update links in the summary, etc
         for p in generators:
             if hasattr(p, "refresh_metadata_intersite_links"):
                 p.refresh_metadata_intersite_links()
-
-        signals.all_generators_finalized.send(generators)
 
         writer = self._get_writer()
 


avaris:
> Also allows Articles not to have a _summary attribute.

Why? If `metadata["summary"]` exists, what's the benefit of potentially not having `_summary`?

MinchinWeb:
> > Also allows Articles not to have a _summary attribute.
> 
> Why? If `metadata["summary"]` exists, what's the benefit of potentially not having `_summary`?

Backward compatibility, and it stayed because it was late last night and I didn't want to have to work through the implications of removing it. And, for now, the [docs](https://docs.getpelican.com/en/latest/plugins.html) of this warning:

> Avoid content_object_init signal if you intend to read summary or content properties of the content object. That combination can result in unresolved links when [Linking to internal content](https://docs.getpelican.com/en/latest/content.html#ref-linking-to-internal-content) (see [pelican-plugins bug #314](https://github.com/getpelican/pelican-plugins/issues/314)). Use _summary and _content properties instead, or, alternatively, run your plugin at a later stage (e.g. all_generators_finalized).

But if you're okay with it, I would be happy remove any reference to it, update the documentation, make sure my plugin is up to date, and try and stick a big enough warning in the release notes.

---

> PS: I really dislike the way .summary/._summary or .content/._content is handled in general, especially with plugin interactions. It is clunky, way too open to gotchas and generally makes things more tricky than it is supposed to be. And most of it is for backwards compatibility. I want to rework that at some point, but making it "clean" would need to break a bunch plugins related to that. Maybe for 5.0... /on-topic-off-pr-rant

I've been bitten by this too! If seems rather esoteric and non-obvious, and really only something that's discovered through trial and error. The related part of that is that if you call `.content`, it caches the result, which means that you (or any other plugin) can't actually update it... If you were to change how it all works, I would be happy to work through updating any of my plugins that touch it!

avaris:
> Backward compatibility, and it stayed because it was late last night and I didn't want to have to (i.e. too tired to) work through the implications of removing it. 

But you *did remove* it. In `refresh_metadata_intersite_links`, previously `._summary` was always set as long as `metadata['summary']` was there. Now you are only setting `._summary` *if and only if* there was a `._summary` before. I'm asking why you removed it? :). Or rather you opened up to the possibility to have this odd edge case where you can have `metadata['summary']` but not `._summary`

pelican/contents.py
```diff
@@ -520,12 +520,17 @@ def refresh_metadata_intersite_links(self):
...
+            if hasattr(self, "_summary"):
+                self.metadata["summary"] = self._summary
+
+            if "summary" in self.metadata:
+                self.metadata["summary"] = self._update_content(
+                    self.metadata["summary"], self.get_siteurl()
+                )
+
+            if hasattr(self, "_summary") and "summary" in self.metadata:
+                self._summary = self.metadata["summary"]
```
avaris, line 533: To be precise, I'm talking about the last `if` condition here, specifically `hasattr(self, "_summary")`. If you compare to the previous version, this `if` should not be here and their contents should be combined to ensure that `._summary` and `metadata['summary']` are synced:

```python
            if "summary" in self.metadata:
                self.metadata["summary"] = self._update_content(
                    self.metadata["summary"], self.get_siteurl()
                )
                self._summary = self.metadata["summary"]

**Comment by MinchinWeb** [comment]
> previously `._summary` was always set as long as `metadata['summary']` was there.

erh, not quite. The line before was this:

```python
self._summary = self._update_content(self._summary, self.get_siteurl())
```

so if you got into the function without `._summary`, it would just error out with an `AttributeError`, and `metadata['summary']` wouldn't get set because you had error-ed out (and probably crashed Pelican in the process). Previously, there was not guard to confirm that there was a `._summary` attribute.

An implicit assumption has become explicit.

---

Is the (preferred) solution to remove the last if condition and just always set `._summary`?

avaris:
Sorry, bad choice of words on my part. Previous version had its issues. What I was trying to say is that if you update `metadata['summary']`, update `._summary` as well. Having different conditions to both can lead to those being out of sync. i.e. If the original content did not have summary (no `summary` metadata), but afterwards some plugin set `metadata['summary']` after the end of this block, `metadata['summary']` will be correct but there will not be any `._summary` attribute.

> Is the (preferred) solution to remove the last if condition and just always set ._summary?

Yes, As I wrote above, just update both together:

```python
            if "summary" in self.metadata:
                self.metadata["summary"] = self._update_content(
                    self.metadata["summary"], self.get_siteurl()
                )
                self._summary = self.metadata["summary"]
```

MinchinWeb:
> update `metadata['summary']`, update `._summary` as well.

Done!

Thank you for your patience as I get this figured out, and thanks for all your work with Pelican!

avaris (approved):
Thanks!

justinmayer (approved):
Many thanks to @MinchinWeb for the fix and for the excellent attention to detail, made all the more robust thanks to the review by @avaris 🤗

Issue #3292, opened 2024-03-12
Option to disable the RichHandler log handler?


cpitclaudel:
- [x] I have searched the [issues](https://github.com/getpelican/pelican/issues?q=is%3Aissue) (including closed ones) and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://docs.getpelican.com/) and believe that my question is not covered.
- [x] I am willing to lend a hand to help implement this feature. 

## Feature Request

The default logging format introduces aggressive line wraps.  For example:

```
[11:03:07] ERROR    Could not process                          generators.py:871
                    pages/xxxxx/xxxxxxxxxxx/xxxxxxxxxxx.rst                     
                    /home/cpitclaudel/git/xxxxxxx/xxx/xxx/page                  
                    s/xxxxx/xxxxxxxxxxx/xxxxxxxxxxx.rst:74:                     
                    (ERROR/2) …
```

Unfortunately, my editor is not smart enough to undo the line breaks, so it offers to open `s/xxxxx/xxxxxxxxxxx/xxxxxxxxxxx.rst:74` instead.

I would love an option to just print the error using a less-fancy log handler.  The default format is pretty OK:

```
Could not process pages/xxxxx/xxxxxxxxxxx/xxxxxxxxxxx.rst
/home/cpitclaudel/git/xxxxxxx/xxx/xxx/pages/pages/xxxxx/xxxxxxxxxxx/xxxxxxxxxxx.rst:74: (WARNING/2) …
```

egberts:
What if we were to detect a Unix pipe condition and then omit the RichHandler?

Or a CLI option --no-color to disable RichHandler outright?

PR #3293, opened 2024-03-12
Choose logging handler via `--log-handler` CLI option

cpitclaudel:
* `pelican/__init__.py` (parse_arguments): Declare new `--log-handler argument`. 
  Uses a string instead of a `type=` argument to get better error messages when passed an incorrect choice.
  (main): Pass `args.logs_handler` to `init_logging`
* pelican/log.py: Expose default log handler as `DEFAULT_LOG_HANDLER`.
  (init): Pass handler to `logging.basicConfig`.

# Pull Request Checklist

Resolves: GH-3292

  **Question:** I think there's no tests for the logger?
  **Question:** I think there's no separate documentation for the CLI, I just added an argparse `help`

### DIFF
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -19,7 +19,7 @@ __path__ = extend_path(__path__, __name__)
 
 # pelican.log has to be the first pelican module to be loaded
 # because logging.setLoggerClass has to be called before logging.getLogger
-from pelican.log import console
+from pelican.log import console, DEFAULT_LOG_HANDLER
 from pelican.log import init as init_logging
 from pelican.generators import (
     ArticlesGenerator,  # noqa: I100
@@ -455,6 +455,17 @@ def parse_arguments(argv=None):
         ),
     )
 
+    LOGS_HANDLERS = {"plain": None, "rich": DEFAULT_LOG_HANDLER}
+    parser.add_argument(
+        "--logs-handler",
+        default="rich",
+        choices=LOGS_HANDLERS,
+        help=(
+            "Which handler to to format log messages. "
+            "The `rich` handler prints output in columns."
+        ),
+    )
+
     parser.add_argument(
         "--logs-dedup-min-level",
         default="WARNING",
@@ -508,6 +519,8 @@ def parse_arguments(argv=None):
         logger.warning("--port without --listen has no effect")
     if args.bind is not None and not args.listen:
         logger.warning("--bind without --listen has no effect")
+
+    args.logs_handler = LOGS_HANDLERS[args.logs_handler]
 
     return args
 
@@ -631,6 +644,7 @@ def main(argv=None):
         level=args.verbosity,
         fatal=args.fatal,
         name=__name__,
+        handler=args.logs_handler,
         logs_dedup_min_level=logs_dedup_min_level,
     )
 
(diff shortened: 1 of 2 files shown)


pelican/__init__.py
```diff
@@ -455,6 +455,17 @@ def parse_arguments(argv=None):
         ),
     )
 
+    LOGS_HANDLERS = {"plain": None, "rich": DEFAULT_LOG_HANDLER}
+    parser.add_argument(
+        "--logs-handler",
+        default="rich",
+        choices=LOGS_HANDLERS,
+        help=(
+            "Which handler to to format log messages. "
```
justinmayer, line 464: I think the word "use" is missing here? I assume this was intended to say:

> "Which handler to use to format log messages. "

cpitclaudel: Woops.  Thanks, fixed!

justinmayer (approved):
I saw that but decided that particularly discrepancy is okay for now. We shall strive for eventual consistency! 😉 

@avaris: Any thoughts on this PR?

offbyone:
I'm curious what use this flag would be put to. It would be constructive, I think, to include documentation for how to use it, how to define a new handler, and an example custom handler.

cpitclaudel:
> I'm curious what use this flag would be put to.

I use it in makefiles, because the forced line wrapping that the rich handler does breaks hyperlinks:

```
           ERROR    Skipping                                     contents.py:175
                    /home/xxx/xxx/xxxxxxx/xxxxx/xxxxx/xxxxxxxxxx                
                    /xxxxxxxxxxxxx/xxxxxxxx/xxxxxxxxx/xxxxxx.htm                
                    l: could not find information about 'date'           
```

With this format my terminal and my IDE don't let me click on the file in question to navigate to it.

> It would be constructive, I think, to include documentation for how to use it, how to define a new handler, and an example custom handler.

Defining a custom handler is something the Pelican devs would do, not users.  These are standard Python log handlers, so the appropriate documentation would be https://docs.python.org/3/library/logging.handlers.html

GiovanH:
I was just about to request this. Are there any outstanding issues or is this fully approved?

justinmayer:
Unless there are further comments from others, I intend to merge the Ruff-related PR (3239) and then this one.

justinmayer:
Many thanks for your enhancement, Clément! 🌟

egberts:
WOW!  Now if we can re-add the `name` back into `LOG_FORMAT` for log-handler=plain option, then the `--log-handler=plain` output would look like the following:

```console
$ python -m pelican --log-handler=plain -s /tmp/pelicanconf-invalid-PATH-1
load_source: AttributeError: 'NoneType' object has no attribute 'loader'
```
See the word `load_source` (that is in `settings.py` source file, but we now know what to look for).

Instead of the currently HEAD commit shows the following:
```console
$ python -m pelican --log-handler=plain -s /tmp/pelicanconf-invalid-PATH-1
AttributeError: 'NoneType' object has no attribute 'loader'
```

Currently, we have no idea which method/procedures (that `%(name)s` in LOG_FORMAT) provides in when an error occurs (not to mention the missing source:line-no column.  Furthermore, the patch applied has rendered the inability to use such `%(name)s` format (and possibly ignores `LOG_FORMAT`).

cpitclaudel:
@egberts I'm sorry, I don't really understand what you're saying

PR #3305, opened 2024-05-19
Add disabled status for #3304 (POC)

### LINKED ISSUE #3304: Article status to gracefully ignore the file

## Feature Request

Currently, there are only three valid statuses: 'published', 'hidden', and 'draft'. If an article has a status that isn't in that list, it is ignored completely and an error is reported. Otherwise, it is processed in some way and a file of some variety is output.

I would like to add a status, titled something like "ignored", that mimics the behavior of an invalid status (not publishing the file at all) but does not throw an error at runtime. 

There are some files that I want to work on in my drafts, but are sensitive enough that I don't want the files published at all, even unlisted, so I currently use "invalid statuses" to prevent this, but it's a hacky solution. Looking at the documentation, it seems the only metadata configuration that results in the "skip" behavior I want is an invalid status.

GiovanH:
# Pull Request Checklist

Resolves: #issue-number-here 

Proof of concept, let me know if you like this direction and I'll write test cases. 

Closes #3304


GiovanH:
Because I have many drafts, this also has the side effect of speeding up my publish builds by 80%. That's nice.

justinmayer (commented):
Nice work, @GiovanH. I am not sure what the best name for this status should be, but I think I would prefer `skip` over `disabled`. (I also considered `ignore`, but I think `skip` is clearer.)

Also, since there are a number of new commits to `master`, this PR should be rebased to pick up those changes.

I realize this is still a draft PoC, but in the interest of clear communication, I think we should also add tests for this new code before merging. After rebasing on `master`, you should be able to run `invoke coverage` on `master` and then again on this PR branch in order to compare the output and see which _new_ lines need test coverage.

justinmayer:
@GiovanH: Two quick things…

1. You'll need to update your local clone to reflect the default branch name change to `main`.
2. When do you think this PR might be ready for review? I ask because I would like to release a new version of Pelican soon, and I think this would be a useful feature to include.

GiovanH:
I just wanted to get a thumbs up from you on the feature first. I'll go ahead and write the tests/docs now.

justinmayer:
Ciao Giovanni. Just checking in… Any chance you could get this PR ready this weekend?

GiovanH:
Thanks for the reminder, I'll see what I can do

GiovanH:
I am continuing to have incredible difficulties with the build environment (as I always seem to with this project) so it might take a bit longer. 

Renamed and rebased, looking at tests, coverage...

GiovanH:
Coverage looks good. 

Because the nature of this feature is that articles marked "skip" are to be ignored entirely, I think the test updates may consist of just adding a skipped article and expecting tests to remain the same, like so:

GiovanH:
I think I might actually not like `skip` as the status for this, because it's not parallel with the other statuses. Articles *are* drafts, articles are hidden, articles are published, but an article can't be "skip". "Disabled" would work, or "skipp**ed**" would force skip to be parallel with the existing enums.

justinmayer:
Nice work, @GiovanH. Really appreciate your efforts.

> I think I might actually not like `skip` as the status for this, because it's not parallel with the other statuses. Articles _are_ drafts, articles are hidden, articles are published, but an article can't be "skip". "Disabled" would work, or "skipp**ed**" would force skip to be parallel with the existing enums.

Once we started overloading beyond `Status: draft`, we soon arrived at a point where none of these really make sense. "Disabled", "ignored", "skipped"… None of these have anything to do with the _status_ of the article, but are instead what we are telling Pelican to _do to_ the article. "Draft" works in this context, "hidden" _kind of_ works, and… the rest IMO really don't.

Since "draft" is already taken, I say we come up with another status that implies _unfinished_, which Pelican then handles by ignoring it. Some potential candidates could be:

- `idea`
- `plan`
- `stub`
- `sketch`
- `outline`
- `design`

Personally, I like the first three options more than I like the last three.

justinmayer:
@GiovanH: On a side note, you previously said:

> I am continuing to have incredible difficulties with the build environment (as I always seem to with this project) […]

Which specific difficulties have you had?

Which specific area(s) of the documentation about [setting up the development environment](https://docs.getpelican.com/en/latest/contribute.html#setting-up-the-development-environment) are in need of improvement?

GiovanH:
> Once we started overloading beyond `Status: draft`, we soon arrived at a point where none of these really make sense. "Disabled", "ignored", "skipped"… None of these have anything to do with the _status_ of the article, but are instead what we are telling Pelican to _do to_ the article. "Draft" works in this context, "hidden" _kind of_ works, and… the rest IMO really don't.
> 
> Since "draft" is already taken, I say we come up with another status that implies _unfinished_, which Pelican then handles by ignoring it. Some potential candidates could be:
> 
> * `idea`
> * `plan`
> * `stub`
> * `sketch`
> * `outline`
> * `design`
> 
> Personally, I like the first three options more than I like the last three.

Of these, I think only `stub` works. The others aren't traditional statuses and don't clearly communicate that the document to be skipped. 

In defense of skipped and disabled, I'd point to "hidden", which is also past-tense descriptive.

Pick whichever one you'd prefer out of stub, skip, skipped, or disabled. 

> @GiovanH: On a side note, you previously said:
> 
> > I am continuing to have incredible difficulties with the build environment (as I always seem to with this project) […]
> 
> Which specific difficulties have you had?
> 
> Which specific area(s) of the documentation about [setting up the development environment](https://docs.getpelican.com/en/latest/contribute.html#setting-up-the-development-environment) are in need of improvement?

I think the documentation is fine, it's just that `venv` and `invoke` needed significant prodding to work in a cygwin environment. Invoke absolutely refuses to run outside a venv, and the script it installs didn't handle backslash escaping properly, so I had to run it with `python3 -m invoke setup` from within a venv. Most scripts that write files also use the "system default", and I have to mess with autocrlf or dos2unix to get the expected output.

justinmayer:
Another status option I think fits this use case: `incomplete`

GiovanH:
> Another status option I think fits this use case: `incomplete`

I don't *hate* it, but I think it has the same problem of not clearly communicating the behavior. Drafts are incomplete articles too, but the point of the new status is to behave differently. 

It's up to you though.

GiovanH:
?

justinmayer:
Hey @GiovanH. Thanks for your patience. I had intended on shipping a Pelican release last month, but my schedule took a hectic turn. My hope is to be able to dedicate some time to this in mid-August.

I am still conflicted about what status moniker to use for this, but given the discussion up to this point, I’m inclined to go with: `skip`

GiovanH:
@justinmayer Should be ready.

pelican/contents.py
```diff
@@ -547,9 +547,29 @@ def refresh_metadata_intersite_links(self) -> None:
                 self._summary = self.metadata["summary"]
 
 
+class SkipStub(Content):
+    # Stub class representing content that should not be processed in any way
```
pauloxnet, line 551:
```suggestion
    """Stub class representing content that should not be processed in any way."""
```
It's a perfect docstring

pauloxnet (approved):

justinmayer (approved):
Many thanks to @GiovanH for the excellent work on (and patience with!) the implementation of this feature. Thanks also to @pauloxnet for reviewing! ✨

PR #3311, opened 2024-05-30
Change `pre-commit` to run `ruff` and `ruff-format` with fixes

boxydog:
# Pull Request Checklist

Resolves: 3216

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,7 +17,7 @@ repos:
     rev: v0.1.15
     hooks:
       - id: ruff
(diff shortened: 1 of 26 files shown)


boxydog:
FYI:

This changes 26 files, while #3239 changes 27. The difference is I didn't port over the `log.py` change:

```
__all__ = ["init", "console"]
```

That change is valid IMO, but isn't part of the ruff fixing. I had just made that change because PyCharm pointed out that the `console` symbol was being used but not exported.

justinmayer (approved):
Many thanks for all your work on this, @boxydog! 🌟

PR #3312, opened 2024-05-30
More ruff fixes

boxydog:
# Pull Request Checklist

Resolves: 3216

Doesn't actually resolve all the ruff ignores, but chips away at it.


boxydog:
I'll look at the test failure, it's likely real.

justinmayer:
Does it make sense to fix more things without first upgrading to Ruff 0.4.6? It would seem to me that upgrading might be the most sensible next step. What do you think?

boxydog:
Ah good point! I'll put up another PR to upgrade first, then get back to this.

boxydog:
Put up #3313

boxydog:
Rebased to ruff v0.4.6. Nothing changed.

samples/pelican.conf.py
justinmayer, line 1: I imagine this permission change wasn't intended?

boxydog: This change was intended.

It is one of the ruff checks, EXE002. And I believe it's correct. This config file does not look to be something to execute on its own.

justinmayer: Quite right. Thanks for clarifying.

justinmayer (approved):
Thanks, @boxydog!

PR #3315, opened 2024-05-30
Jinja template, css, js precommit filters

boxydog:
# Pull Request Checklist

This is an automated replacement for #2379.

There are no manual fixes.

If this is acceptable, we can merge it and close #2379.

If this change gets conflicted, we can remake the PR, drop in the .pre-commit-config.yaml changes, and apply the rest automatically. (To apply to pelican/tests/output, temporarily comment out that line in the config, run `pre-commit run --all djhtml` and djcss and djjs, then uncomment that line.)

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -13,3 +13,5 @@ db241feaa445375dc05e189e69287000ffe5fa8e
 9d30c5608a58d202b1c02d55651e6ac746bfb173
 # Apply yet more Ruff checks to code
 7577dd7603f7cb3a09922d1edb65b6eafb6e2ac7
+# Indent Jinja templates, HTML, CSS, and JS via DjHTML
+4af40e80772a58eac8969360e5caeb99e3e26e78
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,6 +13,7 @@ repos:
       - id: end-of-file-fixer
       - id: forbid-new-submodules
       - id: trailing-whitespace
+
   - repo: https://github.com/astral-sh/ruff-pre-commit
     # ruff version should match the one in pyproject.toml
     rev: v0.4.6
@@ -20,3 +21,10 @@ repos:
       - id: ruff
         args: [--fix, --exit-non-zero-on-fix]
       - id: ruff-format
+
+  - repo: https://github.com/rtts/djhtml
+    rev: '3.0.6'
+    hooks:
+      - id: djhtml
+      - id: djcss
+      - id: djjs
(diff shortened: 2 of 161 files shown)


boxydog:
I'm wrong. #2379 changes 124 files, and this only changes 39, so this must be doing less?

boxydog:
Ah, pelican/tests/output was excluded. I ran the same pre-commit filters (djhtml, djcss, djjs) on that directory as well, and now this change is 160 files (compared to 124 in #2379).

boxydog:
One subtlety: pelican/tests/output is still excluded from the pre-commit filter, so for that directory this is a one-time change. Perhaps we should change pre-commit to include those dirs?

boxydog:
FYI, I looked at four jinja template formatting projects:

* curlylint, mentioned in #2379. Last commit two years ago.
* jinjalint. Last commit six years ago.
* djLint. Last commit four months ago.
* djhtml. Last commit three weeks ago.

It looked to me like djhtml was clean, active, maintained. djLint might also be a reasonable choice (not sure), but I went with djhtml.

justinmayer:
I imagine you're right that `pre-commit` should apply the same HTML/CS/JSS formatting rules to `pelican/tests/output`.

Regarding template formatting project selection, I don't have a strong opinion on the matter. I think the key difference is that [DjHTML](https://github.com/rtts/djhtml) only focuses on _indentation_, while, say, [djLint](https://github.com/djlint/djlint) includes formatting in its scope, with for example [configuration options](https://djlint.com/docs/configuration/) for inserting newlines before/after tags, closing void tags, etc. Like [DjHTML](https://github.com/rtts/djhtml), it seems [djLint](https://github.com/djlint/djlint) also has configuration options for formatting CSS & JS, for what that's worth.

That all said, I think the fact that [DjHTML](https://github.com/rtts/djhtml) only focuses on indentation is perhaps a feature rather than a shortcoming, and it's probably the one I would be inclined to select as well.

boxydog:
What are next steps here?

I filed PR #3316 to apply pre-commit filters to all directories. Could merge that, then redo this, then merge this?

boxydog:
Sounds good, I'll unconflict this right now.

boxydog:
To unconflict, I did

* `git reset HEAD~2` (to roll back to just .pre-commit-config.yaml changes)
* `git pull --rebase origin main` (and resolved conflicts), to set up .pre-commit-config.yaml properly and get the whitespace changes from master
* `pre-commit run --all` to re-apply changes to templates, css, js

So, no manual fixes.

justinmayer (approved):
Thanks, @boxydog!

PR #3173, opened 2023-08-11
Terminate unsuccessfully on error as default for --fatal.  Fixes #3172.

aknrdureegaesr:
# Pull Request Checklist

Resolves: #3172 

- [ ] Ensured **tests pass** and (if applicable) updated functional test output
- [ ] Conformed to **code style guidelines** by running appropriate linting tools
- [ ] Added **tests** for changed code
- [ ] Updated **documentation** for changed code

PR #3317, opened 2024-05-31
Change default of --fatal command line arg to terminate on error

boxydog:
I noticed #3173 is now conflicted and also in the "pelican 5.0" milestone.

So, this is a re-implementation, along with a simple unit test.

I don't know how to test it well, so it's worth comparing carefully to #3173 during review.

Resolves: 3172

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+Release type: major
+
+Change default of `--fatal` command line switch to terminate on error.
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -438,12 +438,12 @@ def parse_arguments(argv=None):
 
     parser.add_argument(
         "--fatal",
-        metavar="errors|warnings",
-        choices=("errors", "warnings"),
-        default="",
+        metavar="errors|warnings|ignore",
+        choices=("errors", "warnings", "ignore"),
+        default="errors",
         help=(
             "Exit the program with non-zero status if any "
-            "errors/warnings encountered."
+            "errors/warnings encountered, or ignore any errors."
         ),
     )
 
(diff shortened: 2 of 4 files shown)


boxydog:
Honestly, I don't quite understand why "ignore" is treated differently in one spot (made into ""), and "errors" is passed into a different spot. I was just recreating the functionality in the other PR.

boxydog:
Ugh, this breaks a lot of unit tests.

boxydog:
The original PR used "ignore" to ignore, but did not change `init` to respect that, so also used "". I changed `init` to respect "" or "ignore".

justinmayer:
I haven't had time to dig into this topic very deeply, but I believe this change could affect people who rely on the current default. As such, I'm going to mark this as a WIP but only to prevent accidental merge until we are ready, particularly given the presence of a `RELEASE.md` file here that will trigger the major release automatically.

Before that major release, I would prefer to publish a new release sometime soon with what is currently in `master`, plus any other non-breaking PRs that make it in.

@avaris: Given your discussion in the previous PR thread, what are your thoughts about the current state of things here?

egberts:
Yes, this breaks a lot of unit tests, BUT `--fatal` (and CRITICAL) should be the DEFAULT sys.exit() condition.  (on an unrelated note, I too think that ERROR should selectively do `sys.exit()` like most UNIX programs do, but I will not pursue that now or anytime soon.

There should be no need for `--fatal` option.   However, we can progress toward the `sys.exit()` piecemeal as needed.

I do believe we should slowly start identifying all those fatals as really "a hill to die on".

boxydog:
Unfortunately, I don't understand the desired behavior well enough to represent this PR. I just submitted something with some fixes to try to move forward #3173.

Can someone describe next steps here?

PR #3321, opened 2024-05-31
Log warnings about files that would have been processed by disabled readers

boxydog:
Resolves: 1868

This is a proof of concept of producing legible warning messages if markdown is not installed.

Example output:

> WARNING  ./sample.md: Could not import markdown.Markdown.  Have you installed the markdown package?                                 readers.py:688
Done: Processed 0 articles, 0 drafts, 0 hidden articles, 0 pages, 0 hidden pages and 0 draft pages in 0.07 seconds.

If this PR looks good, I can try to write at least one unit test of the new functionality.

Some downsides:
1. this is not super simple
2. it will produce one warning per file

I don't think 2 is a big deal. I think 1 is a judgment call, I'm interested in opinions.

It would probably be easier to just require the markdown package?

### DIFF
--- a/pelican/readers.py
+++ b/pelican/readers.py
@@ -17,7 +17,7 @@ from pelican import rstdirectives  # NOQA
 from pelican.cache import FileStampDataCacher
 from pelican.contents import Author, Category, Page, Tag
 from pelican.plugins import signals
-from pelican.utils import get_date, pelican_open, posixize_path
+from pelican.utils import file_suffix, get_date, pelican_open, posixize_path
 
 try:
     from markdown import Markdown
@@ -124,6 +124,10 @@ class BaseReader:
         content = None
         metadata = {}
         return content, metadata
+
+    def disabled_message(self) -> str:
+        """Message about why this plugin was disabled."""
+        return ""
 
 
 class _FieldBodyTranslator(HTMLTranslator):
@@ -347,6 +351,12 @@ class MarkdownReader(BaseReader):
             metadata = {}
         return content, metadata
 
+    def disabled_message(self) -> str:
+        return (
+            "Could not import markdown.Markdown.  "
+            "Have you installed the markdown package?"
+        )
+
 
 class HTMLReader(BaseReader):
     """Parses HTML files as input, looking for meta, title, and body tags"""
@@ -508,17 +518,23 @@ class Readers(FileStampDataCacher):
     def __init__(self, settings=None, cache_name=""):
         self.settings = settings or {}
         self.readers = {}
+        self.disabled_readers = {}
+        # extension => reader for readers that are enabled
         self.reader_classes = {}
+        # extension => reader for readers that are not enabled
+        disabled_reader_classes = {}
 
         for cls in [BaseReader] + BaseReader.__subclasses__():
             if not cls.enabled:
                 logger.debug(
                     "Missing dependencies for %s", ", ".join(cls.file_extensions)
                 )
-                continue
 
             for ext in cls.file_extensions:
-                self.reader_classes[ext] = cls
+                if cls.enabled:
+                    self.reader_classes[ext] = cls
+                else:
+                    disabled_reader_classes[ext] = cls
 
         if self.settings["READERS"]:
             self.reader_classes.update(self.settings["READERS"])
@@ -531,6 +547,9 @@ class Readers(FileStampDataCacher):
 
             self.readers[fmt] = reader_class(self.settings)
 
+        for fmt, reader_class in disabled_reader_classes.items():
+            self.disabled_readers[fmt] = reader_class(self.settings)
+
         # set up caching
         cache_this_level = (
             cache_name != "" and self.settings["CONTENT_CACHING_LAYER"] == "reader"
@@ -541,7 +560,12 @@ class Readers(FileStampDataCacher):
 
     @property
     def extensions(self):
+        """File extensions that will be processed by a reader."""
         return self.readers.keys()
+
+    @property
+    def disabled_extensions(self):
+        return self.disabled_readers.keys()
 
     def read_file(
         self,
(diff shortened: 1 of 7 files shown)


boxydog:
Looking at the failing tests..

pelican/readers.py
```diff
@@ -347,6 +351,12 @@ def read(self, source_path):
             metadata = {}
         return content, metadata
 
+    def disabled_message(self) -> str:
+        return (
+            "Could not import markdown.Markdown.  "
+            "Have you installed the markdown package?"
```
justinmayer, line 357: Total nitpick, but since the package name is actually upper-cased — and because capitalized proper nouns make sentences more readable — I suggest: `markdown package` ➡️ `Markdown package`

pyproject.toml
```diff
@@ -113,6 +113,10 @@ source-includes = [
 requires = ["pdm-backend"]
 build-backend = "pdm.backend"
 
+[tool.ruff]
+
+# Support python 3.8
+target-version = "py38"
```
justinmayer, line 119: Did you run into a problem without this directive? I only ask because my understanding is that this [isn't necessary](https://docs.astral.sh/ruff/settings/#target-version) when `project.requires-python` is defined in [`pyproject.toml`](https://github.com/getpelican/pelican/blob/8fd94ea025cc0c2f29bf11dd63f02fd04e5add5c/pyproject.toml#L30).

justinmayer (commented):
Nice work, @boxydog! I made a couple of minor comments.

pelican/readers.py
boxydog: I'm actually confused on this point.

https://github.com/Python-Markdown/markdown says `pip install markdown`, i.e. lowercase. To me, the package name is the thing after "pip install". The string in pyproject.toml is also lowercase.

However, pip freeze shows uppercase:

```
$ pip freeze | grep -i mark
Markdown==3.5.1
markdown-it-py==3.0.0
MarkupSafe==2.1.3
```

I'll do whatever you want. Just let me know.

pyproject.toml
boxydog: This was just me thrashing around when I had troubles with python 3.8 type hints. I thought it might help, but I think it didn't.

Removed.

pelican/readers.py
offbyone: I'm with boxydog on this one; I prefer the string to match the one you pip install.

pelican/readers.py
```diff
@@ -508,17 +518,23 @@ class Readers(FileStampDataCacher):
...
+        # extension => reader for readers that are enabled
         self.reader_classes = {}
+        # extension => reader for readers that are not enabled
+        disabled_reader_classes = {}
 
         for cls in [BaseReader] + BaseReader.__subclasses__():
             if not cls.enabled:
                 logger.debug(
                     "Missing dependencies for %s", ", ".join(cls.file_extensions)
                 )
```
offbyone, line 531: It looks like this piece of code is already doing some of the work for you; why not hang the warning off of here?

boxydog: It's easy to warn if disabled, but if I have no relevant files, I don't want the warning. 

So I need somewhere that knows there are relevant files, which mostly in the other place. 

If you think this can look at file suffixes, please say more.

offbyone: Ah... I see what you're getting at. 

I think I'd prefer a single warning rather than one per file, but I'm not dead set on it either way, and I'd rather have lots of warnings than none, so ... +1 from me.

boxydog: I agree one would be better. I could track whether a particular suffix has been warned about and suppress the rest, but I'd rather have fewer lines of code devoted to this minor feature.

pelican/readers.py
justinmayer: Not a big deal either way — we can leave it as-is.

Some info re: my reasoning… I personally prefer uppercase proper nouns in almost all cases, as I find lowercase proper nouns to be harder to visually parse in the context of a sentence. In a richer-text context, that can be ameliorated by using `code formatting` or [links](https://github.com/getpelican/pelican/pull/3321), but that doesn't work in plain-text contexts such as this one.

justinmayer (approved):
Looks good to me! 👍  Any further comments from @getpelican/reviewers before merging?

justinmayer:
@boxydog said:

> If this PR looks good, I can try to write at least one unit test of the new functionality.

Do you think it would be worthwhile to add additional unit test coverage to this PR before it is merged?

boxydog:
> @boxydog said:
> 
> > If this PR looks good, I can try to write at least one unit test of the new functionality.
> 
> Do you think it would be worthwhile to add additional unit test coverage to this PR before it is merged?

Yes, I think that's a good idea, otherwise I'll forget. I'll work on it. 

Fyi, my personal philosophy is "epsilon testing", my term for, "write some tests, more than zero", meaning I don't aim for 100% coverage (as that can feel like more work than it's worth to me).

boxydog:
Coverage on master:

```
TOTAL                                  3918    984    75%

Required test coverage of 74% reached. Total coverage: 74.89%
```

Coverage on this branch with added tests:

```
TOTAL                                  3950    982    75%

Required test coverage of 74% reached. Total coverage: 75.14%
```

I'm happy for you to merge this if you like it.

However, FYI, if I were emperor of pelican, I'd just require the markdown library. Less code, simpler to explain. We want to support markdown anyway, so we'll want to support that "optional" dependency.

boxydog:
Ugh, tests failed! I'll look..

boxydog:
Ugh, tests will be more complicated than I had hoped. I changed Markdown.enabled to False, but presumably if some other test is running at the same time it gets disturbed. So probably I have to mock a Markdown class that has a different enabled.

boxydog:
> However, FYI, if I were emperor of pelican, I'd just require the markdown library. Less code, simpler to explain. We want to support markdown anyway, so we'll want to support that "optional" dependency.

.. and then I'd close this PR instead of merge it.

Again, whatever you want, I just want to move important things forward.

justinmayer:
Given the continued interest in alternative Markdown parsers (e.g., #3344), there may be value in leaving [Python-Markdown](https://github.com/Python-Markdown/markdown) as an optional dependency. I am therefore inclined to merge this.

Many thanks for this enhancement, @boxydog! 🏅

PR #3322, opened 2024-06-01
Ruff UP031: Use F-string format specifiers instead of percent format

boxydog:
Related to: 3216

Note: these fixes are automatically applied by using ruff with `--unsafe-fixes`. I figured automatic is better.

From https://docs.astral.sh/ruff/rules/printf-string-formatting/:

"In cases where the format string contains a single generic format specifier (e.g. %s), and the right-hand side is an ambiguous expression, we cannot offer a safe fix."

Hopefully unit tests flush out errors, but there is risk here.


justinmayer (approved):
Thank you, @boxydog! 👍

PR #3325, opened 2024-06-05
Add Dependabot checking

boxydog:
I think I also had to enable dependabot in the github interface. See [here](https://docs.github.com/en/code-security/getting-started/dependabot-quickstart-guide#enabling-dependabot-for-your-repository).

I wasn't sure if dependabot was even running, because everything is silent, but [this run](https://github.com/boxydog/pelican/actions/runs/9352026556/job/25739167659) shows dependabot checking. It is not clear what file it is checking, but for example I see "Checking if feedgenerator  needs updating", which is only in pyproject.toml, so it looks to me like it is doing the right thing.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+# See https://docs.github.com/en/free-pro-team@latest/
+#  github/administering-a-repository/enabling-and-disabling-version-updates
+
+version: 2
+updates:
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    open-pull-requests-limit: 10


justinmayer (approved):
Dependabot appears to currently have five settings at `/settings/security_analysis`, the first two of which were already enabled:

- Dependabot alerts
- Dependabot security updates
- Grouped security updates
- Dependabot version updates
- Dependabot on Actions runners

I just enabled "Grouped security updates" and "Dependabot on Actions runners". The only one left, "Dependabot version updates", should be enabled when this PR is merged. Let's proceed and see how we fare.

Many thanks for this, @boxydog! ✨

PR #3330, opened 2024-06-14
Add theme, Python version, and feed domain to GitHub Pages deployment workflow

seanh:
Add theme, Python version and feed domain support to the reusable GitHub Actions workflow for deploying a Pelican site to GitHub Pages:

1. Add a new `theme` option to the workflow that callers can use to specify an external theme to be checked out and used

2. Add a new `python` option to the workflow that callers can use to specify the Python version, in case they need to build their site with a particular version of Python

3. Pass `--extra-settings FEED_DOMAIN='"${{ steps.pages.outputs.origin }}"'` to the `pelican` command to set the value of Pelican's `FEED_DOMAIN` setting for feed URLs.

### DIFF
--- a/.github/workflows/github_pages.yml
+++ b/.github/workflows/github_pages.yml
@@ -17,6 +17,26 @@ on:
         default: "output/"
         description: "Where to output the generated files (`pelican`'s `--output` option)"
         type: string
+      theme:
+        required: false
+        default: ""
+        description: "The GitHub repo URL of a custom theme to use, for example: 'https://github.com/seanh/sidecar.git'"
+        type: string
+      python:
+        required: false
+        default: "3.12"
+        description: "The version of Python to use, for example: '3.12' (to use the most recent version of Python 3.12, this is faster) or '3.12.1' (to use an exact version, slower)"
+        type: string
+      siteurl:
+        required: false
+        default: ""
+        description: "The base URL of your web site (Pelican's SITEURL setting). If not passed this will default to the URL of your GitHub Pages site, which is correct in most cases."
+        type: string
+      feed_domain:
+        required: false
+        default: ""
+        description: "The domain to be prepended to feed URLs (Pelican's FEED_DOMAIN setting). If not passed this will default to the URL of your GitHub Pages site, which is correct in most cases."
+        type: string
 permissions:
   contents: read
   pages: write
@@ -33,18 +53,31 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
-          python-version: "3.11"
+          python-version: ${{ inputs.python }}
+      - name: Checkout theme
+        if: ${{ inputs.theme }}
+        run: git clone '${{ inputs.theme }}' .theme
       - name: Configure GitHub Pages
         id: pages
         uses: actions/configure-pages@v3
       - name: Install requirements
         run: pip install ${{ inputs.requirements }}
       - name: Build Pelican site
+        shell: python
         run: |
-          pelican \
-            --settings "${{ inputs.settings }}" \
-            --extra-settings SITEURL='"${{ steps.pages.outputs.base_url }}"' \
-            --output "${{ inputs.output-path }}"
+          import subprocess
+
+          cmd = "pelican"
+          cmd += " --settings ${{ inputs.settings }}"
+          cmd += " --extra-settings"
+          cmd += """ SITEURL='"${{ inputs.siteurl || steps.pages.outputs.base_url }}"'"""
+          cmd += """ FEED_DOMAIN='"${{ inputs.feed_domain || steps.pages.outputs.base_url }}"'"""
+          cmd += " --output ${{ inputs.output-path }}"
+
+          if "${{ inputs.theme }}":
+              cmd += " --theme-path .theme"
+
+          subprocess.run(cmd, shell=True, check=True)
       - name: Fix permissions
         run: |
           chmod -c -R +rX "${{ inputs.output-path }}" | while read line; do
--- a/docs/tips.rst
+++ b/docs/tips.rst
@@ -163,8 +163,8 @@ Pelican-powered sites can be published to GitHub Pages via a `custom workflow
 
 Notes:
 
-* You don't need to set ``SITEURL`` in your Pelican settings: the workflow will
-  set it for you
+* You don't need to set ``SITEURL`` or ``FEED_DOMAIN`` in your Pelican
+  settings: the workflow will set them correctly for you
 
 * You don't need to commit your ``--output`` / ``OUTPUT_PATH`` directory
   (``output/``) to git: the workflow will run ``pelican`` to build the output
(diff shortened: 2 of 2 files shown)


.github/workflows/github_pages.yml
```diff
@@ -33,18 +43,31 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
-          python-version: "3.11"
+          python-version: ${{ inputs.python }}
+      - name: Checkout theme
+        if: ${{ inputs.theme }}
+        run: git clone '${{ inputs.theme }}' .theme
```
seanh, line 49: This should work for any Git repo URL, not just GitHub ones

.github/workflows/github_pages.yml
```diff
@@ -33,18 +43,31 @@ jobs:
...
+          cmd += " --settings ${{ inputs.settings }}"
+          cmd += " --extra-settings"
+          cmd += """ SITEURL='"${{ steps.pages.outputs.base_url }}"'"""
+          cmd += """ FEED_DOMAIN='"${{ steps.pages.outputs.origin }}"'"""
+          cmd += " --output ${{ inputs.output-path }}"
+
+          if "${{ inputs.theme }}":
+              cmd += " --theme-path .theme"
+
+          subprocess.run(cmd, shell=True, check=True)
```
seanh, line 70: This needs to pass the `--theme-path` to `pelican` only if the caller workflow passed the `theme` option. I'm dropping into Python as the shell here, this simple script is just complex enough that writing it in Bash would be a pain.

.github/workflows/github_pages.yml
```diff
@@ -33,18 +43,31 @@ jobs:
...
-            --settings "${{ inputs.settings }}" \
-            --extra-settings SITEURL='"${{ steps.pages.outputs.base_url }}"' \
-            --output "${{ inputs.output-path }}"
+          import subprocess
+
+          cmd = "pelican"
+          cmd += " --settings ${{ inputs.settings }}"
+          cmd += " --extra-settings"
+          cmd += """ SITEURL='"${{ steps.pages.outputs.base_url }}"'"""
+          cmd += """ FEED_DOMAIN='"${{ steps.pages.outputs.origin }}"'"""
```
seanh, line 64: Adding the `FEED_DOMAIN` setting, along with the `SITEURL` setting that was already here.

GitHub Actions knows the URL of the GitHub Pages site that is being deployed to, so there's no need for the user to supply the `SITEURL` and `FEED_DOMAIN` settings, the workflow can inject them automatically.

`${{ steps.pages.outputs.base_url }}` is the base URL that the site is being deployed to, appropriate for the `SITEURL` setting. For example `"https://octocat.github.io"` (for a user or organization site) or `"https://octocat.github.io/my-repo"` for a project site. If the user has set up a custom domain for GitHub Pages it'll use that, e.g. `"https://www.example.com"` or `"https://www.example.com/my-repo"`.

`${{ steps.pages.outputs.origin }}` is the origin part of the site URL, for example `"https://octocat.github.io"` or `"https://www.example.com/my-repo"` (no `/my-repo` at the end for project sites). I believe this is correct for `FEED_DOMAIN`?

(These variables come from the output's of GitHub's `configure-pages` action, they're documented here: <https://github.com/actions/configure-pages/blob/e1bedb377708461b49c882870df192bdec04b996/action.yml#L22-L30>.)

justinmayer: I added the `FEED_DOMAIN` setting for the use case in which the domain used for the web site (e.g., `example.com`) is different than the domain used for serving the feed(s) (e.g., `feeds.example.com`). If, on the other hand, both the web site and the feeds are served from the same domain, then the appropriate setting should be: `FEED_DOMAIN = SITEURL`

Does that help clarify?

seanh: Got it, thanks.

Ok, I've changed it to set `FEED_DOMAIN` to the same URL of the GitHub Pages site as it sets `SITEURL` to. I've tested this on two of my own Pelican sites, one a user site and one a project site (project sites are deployed to sub-folders of the user's domain) and it seems to work correctly.

I also added an optional `feed_domain` workflow input so that users can override the default `FEED_DOMAIN` in case they're somehow deploying their feeds to a different domain.

Also added a `siteurl` input to override the default `SITEURL`. I'm not sure whether there's a use-case for setting `SITEURL` manually with this workflow but I put it in just for good measure.

I also added a note to the docs about GitHub Pages generating `http://` URLs for `https://` sites and how to fix it.

docs/tips.rst
```diff
@@ -152,6 +152,25 @@ Pelican-powered sites can be published to GitHub Pages via a `custom workflow
...
+
+       # .github/dependabot.yml
+       version: 2
+       updates:
+         - package-ecosystem: "github-actions"
+           directory: "/"
+           schedule:
+             interval: "monthly"
+
+   See `GitHub's docs about using Dependabot to keep your actions up to date <https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot>`_.
```
seanh, line 172: Also added this note about pinning the workflow version and then using Dependabot to keep it up to date.

justinmayer (approved):
Many thanks for this enhancement, Sean! 😁

seanh:
@justinmayer is the readthedocs integration broken? I noticed that the documentation changes in this PR haven't appeared over in [the docs](https://docs.getpelican.com/en/latest/tips.html#publishing-to-github-pages-using-a-custom-github-actions-workflow) and it looks like the docs [haven't had a build in several months](https://readthedocs.org/projects/pelican/builds/)

seanh:
(I was assuming RTD builds should be triggered automatically when a commit is merged into `master`? But maybe they're triggered manually?)

justinmayer:
@seanh: ReadTheDocs is _supposed_ to automatically trigger a build when a commit lands in the primary branch, but the web-hook constantly gets out-of-sync for who-knows-what-reason, which silently breaks those alleged automatic builds.

And then, when I tried a manual build just now, I see this inscrutable error: https://readthedocs.org/projects/pelican/builds/24724000/

Any idea how to resolve that error?

(I find ReadTheDocs to be very frustrating.)

seanh:
> (I find ReadTheDocs to be very frustrating.)

Yeah readthedocs/Sphinx can be a bit fragile and complicated. Sphinx has a lot of powerful features that're useful for maintaining large documents and that other documentation systems lack. But I think it's a bit heavy for small projects. Readthedocs does have [support for MkDocs](https://docs.readthedocs.io/en/stable/intro/getting-started-with-mkdocs.html) which might be simpler and more robust.

> Any idea how to resolve that error?

I may be able to look into it at some point in the next several days...

The error appears to be coming from [`sphinx.ext.ifconfig`](https://www.sphinx-doc.org/en/master/usage/extensions/ifconfig.html), an extension that's only used in one place in the Sphinx docs, so removing that extension might be an option. Apparently readthedocs has a [version warning feature](https://docs.readthedocs.io/en/stable/versions.html#version-warning) that looks like it might be a substitute

justinmayer:
@seanh: Thank you for taking a quick look. Via the PR linked above, I removed the `ifconfig` extension, and now the ReadTheDocs build completes successfully. Moreover, the web-hook now appears to be working properly after the re-sync I did yesterday, and I even managed to get ReadTheDocs to build the documentation when pull requests are submitted so we can hopefully catch these and other documentation-related problems in the future. Last but not least, the built-in version warning banner appears when viewing the `latest` documentation, so the functionality originally added in #815 seems to no longer be necessary. ✨

PR #3359, opened 2024-07-05
Add docstring for each procedure and method for pelican/__init__.py

egberts:
# Pull Request Checklist

Add missing docstrings to `pelican/__init__.py`.  

Zero code change.

Just a rough cut to see if the Pelican community are up for that.  ( I know some folks like comment-less code so treading lightly here).

### DIFF
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -1,3 +1,42 @@
+"""
+Pelican - Extensible Static Site Generator, with plugin support.
+
+Pelican is a static site generator, written in Python, that allows
+you to create websites by composing text files in formats such as
+Markdown, reStructuredText, and HTML.
+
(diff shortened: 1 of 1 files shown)


boxydog:
I am only one voice, and I don't have much influence over whether this gets merged, but ..

I'd be concerned that the long comments would get out of date. Anything that is not unit tested or with eyes on it eventually breaks.  For example, the very long comments (e.g., at the top, in __init__, etc.) seem more likely to be out of date to me. So I'd want to hear more about what sort of comments might be likely to remain correct over the long term.

Also, I'd be more interested in type hints, for example. Instead of ":returns None", how about "-> None" on the function signature. Those are also not currently checked, but at least I'd see if they were wrong in my IDE.

PR #3384, opened 2024-07-28
Handle internal links without extensions

dltj:
# Pull Request Checklist

# Reason for change

Jekyll—or at least Jekyll as I've configured it—allows for internal links to files regardless of their extension. Say, for example, you have two posts: `post1.md` and `post2.html`

```markdown
Title: Post 1

This file has an internal link to post 2: [Link to Post 2]({filename}/post2)
```

I could convert-and-rename `post2.html` to `post2.md` and the internal link in `post1.md` would not need to change. (I did this a lot in my Jekyll days because the Jekyll blog had a lot of content exported from WordPerfect as HTML files.)

# Proposed implementation

In `_link_replacer()`, create a loop of Reader filename extensions to append to the end of the file path and calls `_get_linked_content()` with that extension. The first time through the loop the extension is `None` so the function behaves as it currently does. If that fails to find a path in the context, the loop tries adding extensions to the path.

# Open questions

1. How does one get a list of extensions? I could import them through, say, `YAMLMetadataReader.file_extensions`, but that seems like an unwise coupling.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+Release type: minor
+
+Allow extensionless internal links
--- a/pelican/contents.py
+++ b/pelican/contents.py
@@ -14,7 +14,6 @@ try:
 except ModuleNotFoundError:
     from backports.zoneinfo import ZoneInfo
 
-
 from pelican.plugins import signals
 from pelican.settings import DEFAULT_CONFIG, Settings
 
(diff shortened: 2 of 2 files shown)


justinmayer:
Hi Peter. Apologies for the delay in responding to your question in this draft pull request.

> Where do we find a comprehensive list of Reader extensions?

As you may have already noticed, if you look through `pelican/readers.py`, you should see that each reader class has its own `file_extensions` attribute. For example: [the list of Markdown extensions](https://github.com/getpelican/pelican/blob/882cd16e114cac40f0bc39e0b317981939b598d7/pelican/readers.py#L293).

Further down, you should see the [`Readers`](https://github.com/getpelican/pelican/blob/882cd16e114cac40f0bc39e0b317981939b598d7/pelican/readers.py#L508) class, which has an [`extensions` property](https://github.com/getpelican/pelican/blob/882cd16e114cac40f0bc39e0b317981939b598d7/pelican/readers.py#L561-L564).

Does that help at all?

PR #3387, opened 2024-08-04
Optimize imports in `tasks.py.jinja2`

gagath:
When opening the generated tasks.py file with a Python linter enabled, some imports will be marked as unecessary depending on the options passed to pelican-quickstart.

Add if statements around the optional imports so they are used only when necessary.

# Pull Request Checklist


justinmayer (approved):
Thank you for the enhancement, Agathe! 🏅

PR #3388, opened 2024-08-06
Change order of SUFFIXES so index.html is served first

RichGibson:
A file named .html should not exist in the output directory. 

Somehow I ended up with one, and it was difficult to debug. Changing the order of the SUFFIXES allows the index.html file to be served before checking for .html.

### DIFF
--- a/pelican/server.py
+++ b/pelican/server.py
@@ -49,7 +49,7 @@ def parse_arguments():
 
 
 class ComplexHTTPRequestHandler(server.SimpleHTTPRequestHandler):
-    SUFFIXES = [".html", "/index.html", "/", ""]
+    SUFFIXES = ["/index.html",".html", "/", ""]
 
     extensions_map = {
         **server.SimpleHTTPRequestHandler.extensions_map,


pauloxnet (approved):
There's some test to update to merge this fix?

RichGibson:
I am not sure what the test would look like, I'm sorry.

boxydog:
I don't understand why this pull request didn't run tests. When it runs test_server.py I believe it will fail due to [these lines](https://github.com/getpelican/pelican/blob/0da2530d9b0de8c8a3c9757f8e9d6d520d837fc2/pelican/tests/test_server.py#L46):

```
            # foo.html has precedence over foo/index.html
            path = handler.get_path_that_exists("foo" + suffix)
            self.assertEqual(path, "foo.html")
```

It failed when I ran it locally.

To prevent your issue, you might have to do something more complicated, such as "if foo is a directory, then look for index.html first"? I don't understand get_path_that_exists deeply.

boxydog:
FYI the test failure on my box:

```
        for suffix in ["", "/"]:
            # foo.html has precedence over foo/index.html
            path = handler.get_path_that_exists("foo" + suffix)
>           self.assertEqual(path, "foo.html")
E           AssertionError: 'foo/index.html' != 'foo.html'
E           - foo/index.html
E           ?    ------
E           + foo.html

pelican/tests/test_server.py:48: AssertionError
=========================== short test summary info ============================
FAILED pelican/tests/test_server.py::TestServer::test_get_path_that_exists - ...
```

PR #3392, opened 2024-09-13
Fix symbolic link handling in `pelican-themes -s/-c`

frederik-elwert:
# Pull Request Checklist

# Description

Symlinks with relative paths were not handled properly by `pelican-themes`.

Example:

The pelican theme to be installed is in a sibling directory to the project directory. Running

```bash
pelican-themes -s ../my-theme
```

creates a symlink `../my-theme` in pelican’s theme folder. This link is broken, since the relative path in the symlink will now be resolved from pelican’s theme folder.

When running `pelican-themes -c` from the project folder, it will not detect the broken symlink, since the relative path will actually resolve from the current directory (just not from pelican’s theme folder).

This pull request fixes this by running `os.path.realpath` on the passed path both when creating symlinks and when checking symlinks. This will make sure that always the absolute paths are linked/checked.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,3 +1,0 @@
-Release type: minor
-
-Fix symlink handling in `pelican-themes`.
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+Release type: minor
+
+Fix symlink handling in `pelican-themes`.
--- a/pelican/tools/pelican_themes.py
+++ b/pelican/tools/pelican_themes.py
@@ -249,6 +249,7 @@ def install(path, v=False, u=False):
 
 def symlink(path, v=False):
     """Symbolically link a theme"""
+    path = os.path.realpath(path)
     if not os.path.exists(path):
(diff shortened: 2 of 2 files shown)


justinmayer:
@frederik-elwert: Thank you for the submission!

Any thoughts from @getpelican/reviewers regarding this pull request?

avaris (approved):

frederik-elwert:
> Thanks for the contribution.
> 
> It would be nice to have a unit test. It might be pretty easy. Whatever broke before.

Yes, I agree, it would always be good to have tests. I'm not super experienced with writing tests, and pelican-themes requires file system level interactions with the current Python environment, so I was a bit out of my depth.

justinmayer (approved):
Many thanks for the fix, Frederik. I'm going to merge this now so that we can release a fix for this and thereby eliminate the problem for end users as soon as possible. If you or @boxydog can think of one or more unit tests that might prevent future regressions, submitting them in a follow-up pull request would be most appreciated. Thanks again!

Issue #3394, opened 2024-09-19
Running `pelican -r -l` produces Pickle-related error


olivierverdier:
- [x] I have read the [Filing Issues](https://docs.getpelican.com/en/latest/contribute.html#filing-issues) and subsequent “How to Get Help” sections of the documentation.
- [x] I have searched the [issues](https://github.com/getpelican/pelican/issues?q=is%3Aissue) (including closed ones) and believe that this is not a duplicate.

- **OS version and name**: MacOS 14.6.1
- **Python version**: 3.12.6
- **Pelican version**: 4.10.0
- **Link to theme**: None
- **Links to plugins**: None
- **Link to your site**: None
- **Link to your source**: 
- **Link to a [Gist](https://gist.github.com/) with the contents of your settings file**: Default settings

#### Preliminary steps
With the very last pelican version (coming from github), create a new blog with `pelican-quick-start`. Go in the newly created site and run ~`make devserver`~ `pelican -lr`.

#### Outcome
Error coming from a pickle error, but which looks like:
```bash
[11:02:20] CRITICAL AttributeError: 'NoneType' object has no attribute        __init__.py:683
                    'terminate'                                                              
```

### Explanation

The problem is that the process `p1` created in `__init__.main` contains an object of type `RichHandler` which can not get pickled.

### ~Solution~ Workarounds

Use `pelican -lr --log-handler=plain` until the problem is fixed.

If you call `pelican -lr` from `make devserver`, modify the `Makefile` by specifying a plain log hander, for instance by modifying the line containing `PELICANOPTS`:
```Makefile
PELICANOPTS=--log-handler plain
```

justinmayer:
Upon further review, to be fair to @ben-n93, issue #3397 refers to running `pelican -r -l`, which is where the problem actually originates. The Makefile changes suggested in this issue only resolve the problem at the `make devserver` wrapper level — not at the level of the underlying `pelican -r -l` command that said wrapper invokes. So the Makefile change is really only a useful workaround for folks who use the Make automation, and the problem should really instead be fixed where it is actually occurring.

olivierverdier:
Sure, maybe the title should be changed to "Pickling error with `pelican -lr`" instead.

The workaround I suggest is, generally, to change the log handler in order to solve the problem, and that is by adding the extra option `--log-handler=plain`.

So, `pelican -lr --log-handler=plain` is another way to solve the problem.

justinmayer:
[...]

PR #3398, opened 2024-09-21
Don't store a RichHandler object in the `argparse` namespace (#3394)

cpitclaudel:
Resolves: #3394 

- [ ] ~Added **tests** for changed code~
- [ ] ~Updated **documentation** for changed code~

Unfortunately I can't test locally, but this patch should be harmless, at least.

### DIFF
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -310,6 +310,9 @@ class ParseOverrides(argparse.Action):
         setattr(namespace, self.dest, overrides)
 
 
+LOG_HANDLERS = {"plain": None, "rich": DEFAULT_LOG_HANDLER}
+
+
 def parse_arguments(argv=None):
     parser = argparse.ArgumentParser(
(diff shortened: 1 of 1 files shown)


snh:
I too was getting the error described in #3394 and can confirm this patch fixes the issue for me.

justinmayer:
Thanks, Clément! Do you think it's worth adding a test that fails for MacOS images in CI when this change is not present?

cpitclaudel:
A regression test would be nice, yes.  I suppose the best way to go about it would be to start the file watcher and then stop it a few seconds later somehow, asserting that it didn't crash in the meantime?

justinmayer (approved):
Many thanks for the fix, Clément. I'm going to merge this now so that we can release a fix for this and thereby eliminate the problem for end users as soon as possible. If you think of one or more unit tests that might prevent future regressions, submitting them in a follow-up pull request would be most appreciated. Thanks again!

Issue #3403, opened 2024-10-07
Decouple Deployment to Github Pages from Published Action


noelmiller:
- [x] I have searched the [issues](https://github.com/getpelican/pelican/issues?q=is%3Aissue) (including closed ones) and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://docs.getpelican.com/) and believe that my question is not covered.
- [x] I am willing to lend a hand to help implement this feature. 

## Feature Request

Hi There! 

When building my pelican site in a PR, I noticed that the action the pelican team publishes will push the deployment to github pages no matter what. My suggestion would be to decouple the deployment of the site and the building of the site from each other. This way, you can skip the publish job when running the build workflow from a PR and can ensure your build is successful in a PR before merging your change into your main branch.

Here is an example of what I did in my pelican.yml: https://github.com/noelmiller/noelmiller.github.io/blob/main/.github/workflows/pelican.yml

Line 15 is the important part: https://github.com/noelmiller/noelmiller.github.io/blob/e8322ab315dae4a488ba403fd2b78ddfacedee41/.github/workflows/pelican.yml#L15

Here is what I did in my fork: https://github.com/getpelican/pelican/compare/main...noelmiller:pelican:main

I'm filing an issue because I'm not sure if this is something you guys would want? It would just require removing that section and including documentation on what a user is supposed to do. 

Thoughts? 

Thanks, 

Noel Miller

justinmayer:
@seanh: Any thoughts on this topic?

seanh:
@justinmayer I think it's a great idea to let people test if the build succeeds on their PR before merging and trying to deploy it!

I would probably implement it slightly differently than what @noelmiller has done. Here's what I'd try:

1. Have a `.github/workflows/build.yml` workflow that just contains the `build` job from the current `github_pages.yml` workflow.

   Users who want to can then have their own `build.yml` caller workflow in their website repos that calls Pelican's `build.yml` workflow, and this caller `build.yml` workflow can be configured to run on PRs and also to be runnable manually (with `workflow_dispatch`).

2. Have a separate `.github/workflows/github_pages.yml` workflow that has a job that calls the `build.yml` workflow and then has the `deploy` job from the current `github_pages.yml` workflow.

   In their website repos users would continue to call this `github_pages.yml` workflow in the same way that they do currently, with a caller workflow that runs on push to the `main` branch or manually.

[...]

PR #3404, opened 2024-10-08
De-couple build workflow from GitHub Pages publication

noelmiller:
# Pull Request Checklist

Resolves: #3403  


seanh (changes requested):
Hi @noelmiller,

I've given this a test and it works but I think there are two problems with my idea to split out a separate `build.yml` file, so I'm very sorry to have wasted your time by getting you to do that:

1. As you've pointed out you have to duplicate all of the inputs and their descriptions, types, etc in both files.
2. The reference to `@main` in `uses: "getpelican/pelican/.github/workflows/build.yml@main"` in the `github_pages.yml` workflow makes development and testing awkward. For example let's say you want to send a pull request that contains changes to `build.yml` and that also requires changes to `github_pages.yml` at the same time. The problem is that the `github_pages.yml` on your PR branch will still contain the reference to `build.yml` `@main` so if you try to test the branch's version of `github_pages.yml` it'll call the version of `build.yml` from `main` not from the PR branch and will fail. This could be worked around but it's awkward.

So I think we should go back to having both jobs in the one `github_pages.yml` file as you had it originally, sorry! I've sent you a PR to do this: https://github.com/noelmiller/pelican/pull/1. Take a look at my PR and if you agree, please merge it into this PR and re-request a review from me.

Aside from that, the only other issue I see is adding docs explaining how to test your build without deploying it, but I think that can be done in a separate PR later.

### Future thoughts

In the future if we ever want to add support for more deployment platforms (not just GitHub Pages) we can add those as additional optional jobs in the same workflow file.

The naming of the workflow as `github_pages.yml` is a bit unfortunate now because the workflow doesn't always deploy to GitHub Pages, if you're only calling it to test the build or if (in future) you're calling it to deploy to a different platform then your workflow run doesn't actually have anything to do with GitHub Pages. A different name would probably be better but renaming the file would break any existing calling workflows that're using the current name. I think we'd have to put a `github_pages.yml` workflow that just prints out an error message telling people to update their caller workflows (or perhaps it could delegate to the new workflow name, passing through all the arguments). **I don't think we should rename the workflow now**, let's just leave it as-is for now, just a thought for the future.

noelmiller:
@seanh 

I like this approach a lot better. Where would the best place be to update the documentation regarding this? Or do you want to do that as a separate PR?

seanh:
> Where would the best place be to update the documentation regarding this? Or do you want to do that as a separate PR?

I'm actually a new contributor to Pelican myself, the GitHub Actions workflow is the only thing I've contributed.

I think you could either add the documentation changes to this PR or do them as a separate PR, either way seems fine to me.

The <https://docs.getpelican.com/en/latest/tips.html> page in the docs comes from the <https://github.com/getpelican/pelican/blob/main/docs/tips.rst> file in this repo, so that's the file you want to change. There's already docs in that file about how to use the workflow to deploy to GitHub Pages, so you just need to add a section explaining how to use the workflow to test the build only. Using the workflow just to test the build doesn't have anything in particular to do with GitHub Pages so I wouldn't put the new documentation in the GitHub Pages section. I'd maybe add a new top-level section to the file about using GitHub Actions to test site generation.

seanh:
Oh, I forgot to say if you would like to preview your changes to the docs locally you can run:

```console
tox -e docs
```

...to build the docs and then run:

```console
python3 -m http.server -d docs/_build/html/
```

...to serve the docs, and open <http://localhost:8000/>

justinmayer:
If you have `invoke` installed, there is a built-in task for building, watching, and serving:

	invoke docserve

noelmiller:
Added a new commit to update the docs. Any feedback?

> If you have `invoke` installed, there is a built-in task for building, watching, and serving:
>
> ```
> invoke docserve
> ```

**Side Tangent:**
So I went down quite a rabbit hole trying to figure out how to set up my python environment (I am more of a sysadmin than a dev). I'm not sure if this Containerfile would be useful to folks? I work off a Fedora Atomic desktop, so installing dependencies on my local machine is not usually a thing I like to do.

```
FROM cgr.dev/chainguard/python:latest-dev

WORKDIR /app

ADD ./requirements /app/
COPY tasks.py /app/

VOLUME /app

USER root

RUN apk update \
    && apk upgrade

RUN python -m venv /root/virtualenvs/pelican

ENV PATH="/root/virtualenvs/pelican/bin:${PATH}"

RUN pip install -Ur developer.pip \
    && pip install -Ur noel.pip # Includes invoke we could obviously include invoke in the developer.pip

RUN invoke tools

ENTRYPOINT ["invoke"]

```

This is what I ended up writing. Not sure how useful it would be? Works great for docs.

To build: `podman build . -t pelican-source`
To run: `podman run -p 8000:8000 -v ./:/app:Z pelican-source:latest docserve`

Equivalent docker commands would be:
`docker build . -f ./Containerfile -t pelican-source`
`docker run -p 8000:8000 -v ./:/app pelican-source:latest docserve`

docs/tips.rst
```diff
@@ -246,7 +246,52 @@ Here's the complete list of workflow inputs:
...
 |                  |          | in most cases.                             |        |               |
 +------------------+----------+--------------------------------------------+--------+---------------+
+| ``deploy``       | No       | This is used to determine whether you will | bool   | ``true``      |
+|                  |          | deploy the site or not to GitHub Pages.    |        |               |
+|                  |          | This is most useful if you want to test a  |        |               |
+|                  |          | change to your website in a pull request   |        |               |
+|                  |          | before deploying those change.             |        |               |
++------------------+----------+--------------------------------------------+--------+---------------+
+
+Testing your build in a GitHub Pull Request
```
justinmayer, line 256: I would capitalize "Your" and "Build", to be consistent with (most) of the headers in this file.

docs/tips.rst
```diff
@@ -246,7 +246,52 @@ Here's the complete list of workflow inputs:
...
+      push:
+        branches: ["main"]
+      pull_request:
+        branches: ["main"]
+      workflow_dispatch:
+        inputs:
+          deploy:
+            required: false
+            default: true
+            description: "Whether to deploy the site. If checked then build the site and deploy it. If not checked then just test that  the site builds successfully but don't deploy anything."
```
justinmayer, line 274: `If checked then […]` 👉 `If checked, then […]`

`If not checked then […]` 👉 `If not checked, then […]`

`just test that  the` 👈  there is an extra space between "that" and "the"

docs/tips.rst
```diff
@@ -246,7 +246,52 @@ Here's the complete list of workflow inputs:
...
+        with:
+          settings: "publishconf.py"
+          requirements: "-r requirements.txt"
+          deploy: ${{ (github.event_name == 'workflow_dispatch' && inputs.deploy == true) || (github.event_name == 'push' && github.ref_type == 'branch' && github.ref_name == github.event.repository.default_branch) }}
+
+The ``on`` section of the workflow defines the events that will trigger the workflow. In this example, the workflow will run on pushes to the main branch, pull requests to the main branch, and manual runs of the workflow.
+
+``workflow_dispatch`` defines the deploy boolean to be true by default. This means that if you run the workflow manually, it will deploy the site.
+
+The ``deploy`` input for the job is using a set of standard github workflow variables to control when deploy will either be true or false (you can customize this to your needs).
```
justinmayer, line 292: `github` 👉 `GitHub`

deploy 👉 \`\`deploy\`\`

docs/tips.rst
```diff
@@ -246,7 +246,52 @@ Here's the complete list of workflow inputs:
...
+
+In this example, the deploy will be true if the event is a push to the main branch (or merging into main from a PR) or a manual run of the workflow. If the event is a pull request, the deploy will be false and it will only build an artifact for the site.
```
justinmayer, line 294: the deploy 👉 \`\`deploy\`\`

justinmayer (commented):
Thank you for the documentation enhancement. Seems good to me, aside from a few minor suggested changes 👍

justinmayer:
@seanh: Any thoughts on the current state of this pull request?

seanh (approved):

noelmiller:
I agree with squashing the commits. We went through 2 different refactors, so commit history is a bit messy.

justinmayer (approved):
Many thanks to @noelmiller for the enhancement and to @seanh for reviewing 💫

noelmiller:
Awesome! Thank you for the feedback and @seanh for pushing it over the finish line :)

seanh:
:partying_face:

PR #3405, opened 2024-10-10
Add more blocks to Simple theme's base template

frederik-elwert:
This change makes it easier to create new themes by inheriting from the simple theme. It allows customization of the whole body (while still making use of the theme’s head), or individual parts of the body like header, menu, or footer.

# Pull Request Checklist

### DIFF
--- a/pelican/tests/test_theme.py
+++ b/pelican/tests/test_theme.py
@@ -1,0 +1,138 @@
+import os
+import unittest
+from shutil import rmtree
+from tempfile import mkdtemp
+
+from pelican import Pelican
+from pelican.settings import read_settings
+from pelican.tests.support import LoggedTestCase, mute
+
+CURRENT_DIR = os.path.dirname(os.path.abspath(__file__))
+CONTENT_DIR = os.path.join(CURRENT_DIR, "simple_content")
+
+
+class TestTemplateInheritance(LoggedTestCase):
+    def setUp(self):
+        super().setUp()
+        self.temp_output = mkdtemp(prefix="pelican_test_output.")
+        self.temp_theme = mkdtemp(prefix="pelican_test_theme.")
+        self.temp_cache = mkdtemp(prefix="pelican_test_cache.")
+
+        # Create test theme directory structure
+        os.makedirs(os.path.join(self.temp_theme, "templates"), exist_ok=True)
+
+        # Create base.html template that inherits from simple theme
+        template_content = """{% extends "!simple/base.html" %}
+
+{% block head %}
+{{ super() }}
+    <link rel="stylesheet" type="text/css" href="{{ SITEURL }}/theme/css/style.css" />
+{% endblock %}
+
+{% block footer %}
+    <p>New footer</p>
+{% endblock %}
+"""
+
+        with open(os.path.join(self.temp_theme, "templates", "base.html"), "w") as f:
+            f.write(template_content)
+
+    def tearDown(self):
+        """Clean up temporary directories and files"""
+        for path in [self.temp_output, self.temp_theme, self.temp_cache]:
+            if os.path.exists(path):
+                rmtree(path)
+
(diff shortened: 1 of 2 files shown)


justinmayer:
Any feedback on this PR from @getpelican/reviewers?

boxydog:
It is traditional to have blocks for header and footer, so that seems like a decent idea. It might also be traditional to separate them out into footer.html and header.html, though I wouldn't insist.

I don't see customization of the menu, although that is mentioned in the description of this PR.

My bigger questions:

1. Is this actually useful for someone? What is the use case?
2. If yes, then is it possible to write a test?

frederik-elwert:
Sorry for not getting back to this earlier. Thanks for the comments.

> It is traditional to have blocks for header and footer, so that seems like a decent idea. It might also be traditional to separate them out into footer.html and header.html, though I wouldn't insist.

Yes, separating them out is also a good idea, and it would make it easy to override them without even touching the base template. It would require a bit more understanding of the file layout, though, so for beginners having a single file to override might even be simpler. But I’m open to both suggestions.

A slightly related question: At the moment, the blocks are defined _within_ the corresponding elements, i.e., there’s a `<footer>`, and inside that a `{% block footer %}`. I followed the model for `head` here. But the consequence is that even when overriding the footer block, one is still bound to having that footer tag around it. (It would still be possible to override the complete body block, though.) Would it make sense to switch that and have the elements within the block? (The same would apply to includes if one were to factor them out into separate files.)

> I don't see customization of the menu, although that is mentioned in the description of this PR.

That’s the block `nav` that I was referring to.

> My bigger questions:
> 
>     1. Is this actually useful for someone? What is the use case?

I’m teaching SSGs with Pelican to university students, and one issue is customizing the appearance without having to write a full theme. My understanding for the existence of the simple theme was that it would be a starting point for simple customizations, which is what the docs suggest. The reason for the PR is that customization works well for the use case in the docs (mainly providing a custom CSS file), but going beyond that requires a customized copy of the base template. This PR would offer a middle way of reusing the simple theme, but allowing more customization.

And I think it would even be useful for theme authors. (I’m currently writing a simple custom theme that basically uses this mechanism.)

>     2. If yes, then is it possible to write a test?

I’m happy to provide a test, but I’m not really sure what the test would cover?

boxydog:
> I’m teaching SSGs with Pelican to university students

Sounds good. What's an SSG?

> I’m happy to provide a test, but I’m not really sure what the test would cover?

Literally any line of code that you changed. For example, override the footer and test that the new footer shows up.

justinmayer:
SSG is short for [static site generator](https://en.wikipedia.org/wiki/Static_site_generator).

pauloxnet:
> Any feedback on this PR from @getpelican/reviewers?

Please add explicit name in endblocks

justinmayer (approved):
Many thanks for this enhancement, Frederik! Looks good to me.

Any last comments from @getpelican/reviewers? If possible, I would like to merge this today and ship a new release with the other accumulated improvements.

frederik-elwert:
Thanks for the feedback.

* I added explicit endblock names.
* I don’t think there already were tests in place for the simple theme and inheritance. I’m not very familiar with code testing, and in this case it also requires some i/o, but with the help of Claude AI (but manually checked and updated by me), I created a basic test for the simple theme and theme inheritance based on the "simple_content" test data. I hope this goes into the right direction.
* I just rebase the branch onto main and squashed a few extra commits.

pauloxnet (approved):
Thanks for adding endblock names.

justinmayer:
Thanks again to @frederik-elwert for the contribution and to @boxydog and @pauloxnet for reviewing ✨

boxydog:
> I created a basic test for the simple theme and theme inheritance

Fantastic, thank you.

justinmayer:

Issue #3419, opened 2024-11-03
Only one occurrence of {static} is replaced inside a tag


mart-e:
- [x] I have read the [Filing Issues](https://docs.getpelican.com/en/latest/contribute.html#filing-issues) and subsequent “How to Get Help” sections of the documentation.
- [x] I can reproduce this problem with stock/default settings file, theme, and sample content (as described in above “How to Get Help” sections of the documentation).
- [x] I have searched the [issues](https://github.com/getpelican/pelican/issues?q=is%3Aissue) (including closed ones) and believe that this is not a duplicate.

- **OS version and name**: Archlinux
- **Python version**:Python 3.11
- **Pelican version**: 4.10

## Issue
Create a page on a new site that contains a tag with two `{static}` instructions. For instance an image with lazyloading:
```
Title: Test Article
Date: 2024-11-02 21:30
Author: me
Category: Uncategorized
Tags: 
Slug: test-article
Status: draft

<img
  src="{static}/images/placeholder.png"
  data-src="{static}/images/my_image.png"
  loading="lazy"/>

Bye
```

Expected output:

```html
<p><img
  src="/images/placeholder.png"
  data-src="/images/my_image.png"
  loading="lazy"/></p>
<p>Bye</p>
```
Found output:

```html
<p><img
  src="{static}/images/placeholder.png"
  data-src="/images/my_image.png"
  loading="lazy"/></p>
<p>Bye</p>
```

The issue comes from [the regex](https://github.com/getpelican/pelican/blob/0da2530d9b0de8c8a3c9757f8e9d6d520d837fc2/pelican/contents.py#L374-L380)
```py
(?P<markup><[^\>]+  # match tag with all url-value attributes
(?:href|src|poster|data|cite|formaction|action|content)\s*=\s*)
(?P<quote>["\'])      # require value to be quoted
(?P<path>[{|](?P<what>.*?)[|}](?P<value>.*?))  # the url value
(?P=quote)"""
```
that will match only the last attribute (greedy). Changing it to `(?P<markup><[^\>]+?` makes it lazy and matches the first one only so doesn’t really solves the issue either.

Still trying to find a way to solve this without using 2 regex call but I am no expert in regex.

GiovanH:
Might relate to https://github.com/getpelican/pelican/issues/3409?

PR #3420, opened 2024-11-03
[FIX] match all occurrences of static/attach

mart-e:
If a markup element contains more than one {static} or {attach} value, only the last one was replaced.

Simplify the regex to match every occurrence.
This regex no longer checks for `<` at the start which may imply false positive (to test if acceptable or not).

Detail of the issue: #3419


pauloxnet:
Can you add a test?

justinmayer:
@mart-e: If the test failure you are seeing above is due to the updated regex now actually correctly replacing the second `{static}` reference (and please make sure that is actually the case — I have not looked at length), then you can [update the functional test output as described in the docs](https://docs.getpelican.com/en/latest/contribute.html#running-the-test-suite).

pauloxnet (approved):

mart-e:
@justinmayer no that test failure is the test I am adding, looks like my patch is actually not working/not enough. I have some troubles to invoke the test suit on my machine (probably some setup issues).
Is there a way to use a debugger like pudb?

justinmayer:
Let's start with getting the tests running on your local workstation… What have you done so far, and what seems to be the problem with running the tests?

justinmayer:
Looks like your test now passes. Bravo, Martin! Well done 👏

mart-e:
Ok, I have figured out what was the issue, missing file declaration in my new test.

> What have you done so far, and what seems to be the problem with running the tests?

I used `invoke setup` to be sure my env is up to date and then `invoke test` on the main branch. The `test_pelican.py` is failing. The output is very long ([output](https://github.com/user-attachments/files/17611759/output.txt)), my terminal doesn’t have enough scroll back to reach the beginning but the interesting part seems to be that for every file, I get
```
E           diff --git a/home/mat/other/pelican/pelican/tests/output/basic/feeds/all.atom.xml.orig b/home/mat/other/pelican/pelican/tests/output/basic/feeds/all.atom.xml.orig
E           new file mode 100644
E           index 00000000..5552b0ea
E           --- /dev/null
E           +++ b/home/mat/other/pelican/pelican/tests/output/basic/feeds/all.atom.xml.orig
```

justinmayer:
I do not believe there should be any `[…].orig` files in the Pelican repository, so those are likely leftover detritus of some Git operation. I suggest using something like [`fd`](https://github.com/sharkdp/fd), particularly with its [command execution option](https://github.com/sharkdp/fd#command-execution), to remove those files. Perhaps something like:

```shell
cd ~/projects/pelican
fd --hidden --glob '*.orig' --exec trash
```

mart-e:
Oh you are right, much better when purging those files, I can run the tests now. Thanks a lot for your help :+1: 

As a sidenote, I am no longer the initial example with two attributes containing `{static}` for lazyloading images on my blog (as it required javascript to work).
I still think this change is valid but in case you have a doubts, feel free to reject this PR as the usecase is probably rare.

PR #3424, opened 2024-11-10
feat: Change `IGNORE_FILES` setting default to all hidden files

projectgus:
This is to avoid subtle behaviour that contributed to root cause of https://github.com/pelican-plugins/sitemap/issues/36

Specifically: if installing Pelican into a local virtualenv with the pdm or uv default name ".venv", then subdirectories of .venv will contain all of the test .rst and .md files.

If you then run Pelican in that same root directory using the default PATH value (".") then it will add those content files to the site.

# Pull Request Checklist

Resolves: #issue-number-here 

(I haven't added a new test for the default setting, but I can if needed.)

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -150,12 +150,12 @@ Basic settings
 
       READERS = {'foo': FooReader}
 
-.. data:: IGNORE_FILES = ['.#*']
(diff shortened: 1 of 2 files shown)


justinmayer (approved):
Thank you, Angus, for this improved default 👍

PR #3430, opened 2024-11-30
Make build reproducible by reading envvar `SOURCE_DATE_EPOCH` if set

Vincent-C:
As part of the ongoing [Reproducible Builds](https://reproducible-builds.org/) project, packages in Debian/Ubuntu are routinely scanned and tested for build reproducibility. [A bug](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1056571) was filed against the `pelican` package in Debian, so I'm just forwarding the patch that we've applied there to make it build reproducibly, i.e. by attempting to get the current datetime from a standard envvar, [SOURCE_DATE_EPOCH](https://reproducible-builds.org/docs/source-date-epoch/), if it's available. Thanks for considering this change!

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -31,8 +31,8 @@ extensions = [
 source_suffix = ".rst"
 master_doc = "index"
 project = project_data.get("name").upper()
-year = datetime.datetime.utcfromtimestamp(
-    int(os.environ.get("SOURCE_DATE_EPOCH", time.time()))
+year = datetime.datetime.fromtimestamp(
+    int(os.environ.get("SOURCE_DATE_EPOCH", time.time())), datetime.timezone.utc
 ).year
 copyright = f"2010–{year}"  # noqa: RUF001
 exclude_patterns = ["_build"]
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,7 +1,7 @@
 import datetime
 import os
-import time
 import sys
+import time
 
 if sys.version_info >= (3, 11):
     import tomllib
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -30,7 +31,9 @@
 source_suffix = ".rst"
 master_doc = "index"
 project = project_data.get("name").upper()
-year = datetime.datetime.now().date().year
+year = datetime.datetime.utcfromtimestamp(
+    int(os.environ.get("SOURCE_DATE_EPOCH", time.time()))
+).year
```
justinmayer, line 36: This change yields the following deprecation warning:

> DeprecationWarning: `datetime.datetime.utcfromtimestamp()` is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: `datetime.datetime.fromtimestamp(timestamp, datetime.UTC)`

Perhaps the following instead?

```python
year = datetime.datetime.fromtimestamp(
    int(os.environ.get("SOURCE_DATE_EPOCH", time.time())), datetime.timezone.utc
).year
```

See also:
- https://docs.python.org/3/library/datetime.html#datetime.datetime.utcfromtimestamp
- https://til.simonwillison.net/python/utc-warning-fix

justinmayer (changes requested):
Hi Vincent! Thank you for the submission. As it turns out, `datetime.datetime.utcfromtimestamp()` is deprecated, so I made a suggested change. Would you please take a look and let me know your thoughts?

justinmayer (approved):
I went ahead and made my suggested change so I could include this PR in an imminent new release. Thanks again, Vincent.

justinmayer:

Vincent-C:
Thank you!

PR #3432, opened 2024-12-02
Use html5 writer from docutils

### LINKED ISSUE #3004: Switch to Docutils HTML5 writer instead of the HTML4 one

## Feature Request

Pelican has used the Docutils HTML4 writer for about 11 years at this point, and I think it's time to put it to rest. Since 0.13, Docutils includes an HTML5 writer called `html5_polyglot`, which generates (X)HTML5, with correct semantic tags for article sections, etc.

As far as I can tell at a glance, all that is needed is just swapping out the import. I have done so in a custom plugin, by subclassing `pelican.readers.RstReader` and changing the `writer_class` (along with the translator) and it appears to work just fine.

Note that this is a breaking change in some aspects. For example, themes that add styles for `div.section` will break as now those are replaced with actual `section` tags (they still have their ids though).

Also worth of note is that while issue #959 is similar, it's not related as at that time Docutils did not know about HTML5.

nkr0:
# Pull Request Checklist

Resolves: #3004


pauloxnet (approved):
LGTM

justinmayer:
As I mentioned in the linked issue, I suspect this change might be best introduced in the context of a major version, which is why I have targeted it for an upcoming 5.0 release. I hope to organize a sprint sometime soon so we can make that happen 😅

nkr0:
> As I mentioned in the linked issue, I suspect this change might be best introduced in the context of a major version, which is why I have targeted it for an upcoming 5.0 release. I hope to organize a sprint sometime soon so we can make that happen 😅

Yes, I remember. I did a rebase yesterday to fix a build failure after merging `main`

pauloxnet:
After the release of Pelican 4.12 I think this PR is unblocked after a final rebase.

PR #3439, opened 2024-12-27
feat: Add setting to omit selected Typogrify filters. Fixes #3436

### LINKED ISSUE #3436: Using a subset of Typogrify filters

## Feature Request

I appreciate using Typogrify with Pelican, but would like to disable some of the filters it provides. 

I'd especially like to disable the `widont` filter. That filter can sometimes cause text in narrow containers to overflow when the last two words are long, whereas the same text would wrap nicely without the non-breaking space introduced by `widont`. But the nail on `widont`'s coffin, I think, is the CSS `text-wrap` property which can do a much better job (with `balance` or `pretty`). 

One filter I'd definitely want to continue using is `smartypants`.

I suspect that allowing to select a list of filters might require changes to Typogrify. Any ideas on the best way to solve this? Thanks.

davidlesieur:
Remarks:

- The **tests will fail with Typogrify 2.0.7** but work with justinmayer/typogrify#1.
- A test message and the documentation provided here will need to be updated should the next version of Typogrify be other than 2.0.8.

# Pull Request Checklist

Resolves: #3436

- [x] Ensured **tests pass** and (if applicable) updated functional test output *__⚠️ See condition above!__*


docs/settings.rst
```diff
@@ -288,7 +288,14 @@ Basic settings
...
    A list of tags for Typogrify to ignore. By default Typogrify will ignore
    ``pre`` and ``code`` tags. This requires that Typogrify version 2.0.4 or
-   later is installed
+   later is installed.
+
+.. data:: TYPOGRIFY_OMIT_FILTERS = []
+
+   A list of Typogrify filters to skip. Allowed values are: ``'amp'``,
+   ``'smartypants'``, ``'caps'``, ``'initial_quotes'``, ``'widont'``. By
+   default, no filter is omitted (in other words, all filters get applied). This
```
justinmayer, line 297: I wonder whether it would be better for this new setting to define which filters *should* be applied, with the default being the same: all filters are applied. For example, perhaps a default setting that looks something like:

```python
TYPOGRIFY_FILTERS = ["amp", "caps", "initial_quotes", "smartypants", "widont"]
```

I am not saying I am certain that the above approach is superior, but I figured now would be the time to consider it. What do you think?

docs/settings.rst
```diff
@@ -288,7 +288,14 @@ Basic settings
...
+   setting requires that Typogrify version 2.0.8 or later is installed.
```
justinmayer, line 298: Given the additional to-be-released functionality, in keeping with [semantic versioning](https://semver.org/) I expect the next release to be `2.1.0`.

pelican/tests/test_readers.py
```diff
@@ -409,6 +409,90 @@ def test_typogrify_summary(self):
...
+            page = self.read_file(
+                path="article.rst",
+                TYPOGRIFY=True,
+                TYPOGRIFY_OMIT_FILTERS=["this-filter-does-not-exists"],
+            )
+            self.assertRaises(TypeError)
+        except ImportError:
+            return unittest.skip("need the typogrify distribution")
+        except TypeError:
+            return unittest.skip("need typogrify version 2.0.8 or later")
```
justinmayer, line 494: As noted above, I expect the next release to be `2.1.0`.

justinmayer (commented):
Nice work, David! I made a few comments and look forward to your thoughts.

docs/settings.rst
davidlesieur: In general I favor "positive" settings like you suggest. For this particular one, I was on the fence. I thought that it perhaps made sense that all filters be applied by default without Pelican having to know what they are, or users having to know the full list when all they want is maybe remove one or two.

If in the future Typogrify was to get new filters, what behavior would be expected? With the "negative" setting, Pelican users automatically get the new filters as soon as they upgrade Typogrify, whether or not they have disabled some filters. With the "positive" setting, users who have kept the default will automatically get the new filters as soon as they upgrade both Typogrify and Pelican, but users who had specified a list of filters will need to add the new ones to their list (which is not necessary bad; those users might prefer to remain in control and not get unexpected new features).

`TYPOGRIFY_FILTERS` would seem more intuitive to use than `TYPOGRIFY_OMIT_FILTERS`, but I'm still on the fence. I'll go with the one you prefer. Just let me know!

docs/settings.rst
davidlesieur: Got it. I'll make the change accordingly.

docs/settings.rst
justinmayer: You raise some good points. After thinking about them, perhaps the current `TYPOGRIFY_OMIT_FILTERS` setting might be appropriate for most users. 👍

justinmayer:
@davidlesieur: I just released [Typogrify 2.1](https://github.com/justinmayer/typogrify/releases/tag/2.1.0). Could you double-check and ensure tests will pass successfully with the new release? (They failed when I tested locally).

justinmayer (approved):
Many thanks for all your work on this feature, David! 🥇

justinmayer:

PR #3438, opened 2024-12-27
test:  Fix `test_typogrify_ignore_tags` due to Typogrify 2.1 changes

davidlesieur:
Typogrify's 'widont' filter used to ignore the list of 'ignore_tags', but this will be fixed in an upcoming version of Typogrify.

The **tests will fail with Typogrify 2.0.7** but work with justinmayer/typogrify#1.

# Pull Request Checklist

- [x] Ensured **tests pass** and (if applicable) updated functional test output *__⚠️ See condition above!__*
- [x] Conformed to **code style guidelines** by running appropriate linting tools
- [ ] Added **tests** for changed code
- [ ] Updated **documentation** for changed code

PR #3446, opened 2025-01-14
Pin Pygments due to indentation changes in 2.19

justinmayer:
Another day, another wrench thrown into our functional test output because of what appear to be spurious indentation changes introduced by [Pygments 2.19](https://github.com/pygments/pygments/releases/tag/2.19.0). (Pygments 2.19.1 does not appear to address this problem.)

Rather than update our functional test output with what appear to be sub-optimal indentation changes, for now I would rather pin Pygments (yet again) until we decide how to sort things out in a more long-term manner.

The failing test in this PR is related to Typogrify 2.1 and can be ignored, as it will be fixed when #3438 is merged.


justinmayer:
Thank you, Paolo, for taking a look 😊

PR #3452, opened 2025-03-04
GitHub workflow stork install option

joelsvensson:
# Summary
The current github workflow for building and deploying to github pages will fail if using pelican-search with stork
https://github.com/pelican-plugins/search

This PR will add an extra option for it to be installed on the github runner.

I choose to install via cargo as that is available on runner  
https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#installed-software  
and is listed here https://stork-search.net/docs/install

However it takes about 1min for the runner to run this step. I guess the other ways are faster but this should be safe over time.

See this repo for tests with and without the var set  
https://github.com/joelsvensson/joelsvensson.github.io/actions/workflows/build_and_deploy.yml

# Pull Request Checklist

### DIFF
--- a/.github/workflows/github_pages.yml
+++ b/.github/workflows/github_pages.yml
@@ -42,6 +42,11 @@ on:
         default: true
         description: "Whether to deploy the site. If true then build the site and deploy it. If false then just test that the site builds successfully but don't deploy anything."
         type: boolean
+      stork:
+        required: false
+        default: false
+        description: "Whether to add stork. If true it will be installed on runner"
+        type: boolean
 permissions:
   contents: read
   pages: write
@@ -62,6 +67,9 @@ jobs:
       - name: Configure GitHub Pages
         id: pages
         uses: actions/configure-pages@v5
+      - name: Install stork
+        if: ${{ inputs.stork }}
+        run: cargo install stork-search --locked
       - name: Install requirements
         run: pip install ${{ inputs.requirements }}
       - name: Build Pelican site
--- a/.github/workflows/github_pages.yml
+++ b/.github/workflows/github_pages.yml
@@ -82,7 +82,7 @@ jobs:
           subprocess.run(cmd, shell=True, check=True)
       - name: Fix permissions
         run: |
-          chmod -c -R +rX "${{ inputs.output-path }}" | while read line; do
+          chmod -c -R +rX "${{ inputs.output-path }}" | while read -r line; do
             echo "::warning title=Invalid file permissions automatically fixed::$line"
           done
       - name: Upload artifact
--- a/docs/tips.rst
+++ b/docs/tips.rst
@@ -252,6 +252,10 @@ Here's the complete list of workflow inputs:
 |                  |          | change to your website in a pull request   |        |               |
 |                  |          | before deploying those change.             |        |               |
 +------------------+----------+--------------------------------------------+--------+---------------+
+| ``stork``        | No       | This is used to determine whether stork    | bool   | ``false``     |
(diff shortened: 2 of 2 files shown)


.github/workflows/github_pages.yml
```diff
@@ -42,6 +42,11 @@ on:
         default: true
         description: "Whether to deploy the site. If true then build the site and deploy it. If false then just test that the site builds successfully but don't deploy anything."
         type: boolean
+      stork:
+        required: false
+        default: false
+        description: "Whether to add stork. If true it will be installed on runner"
```
justinmayer, line 48: Minor suggested change to description: `"Whether to add Stork search tool. If true, it will be installed on runner."`

I would also suggest capitalizing Stork in all places other than those that are strictly code/CLI-related.

.github/workflows/github_pages.yml
```diff
@@ -62,6 +67,9 @@ jobs:
       - name: Configure GitHub Pages
         id: pages
         uses: actions/configure-pages@v5
+      - name: Install stork
```
justinmayer, line 70: `- name: Install Stork`

docs/tips.rst
```diff
@@ -252,6 +252,10 @@ Here's the complete list of workflow inputs:
 |                  |          | change to your website in a pull request   |        |               |
 |                  |          | before deploying those change.             |        |               |
 +------------------+----------+--------------------------------------------+--------+---------------+
+| ``stork``        | No       | This is used to determine whether stork    | bool   | ``false``     |
+|                  |          | will be installed on the runner to be able |        |               |
+|                  |          | to build a site with stork search enabled  |        |               |
```
justinmayer, line 257: I suggest capitalizing Stork in the two places in this description.

justinmayer (commented):
Other than a few minor comments mostly related to capitalizing "Stork", this seems good to me.

@seanh: Any thoughts about this?

.github/workflows/github_pages.yml
joelsvensson: Fixed in new commit

docs/tips.rst
joelsvensson: Fixed in new commit

.github/workflows/github_pages.yml
joelsvensson: Fixed in new commit

justinmayer:
I resolved the unrelated test failures in a separate PR, so those can be ignored.

@seanh: Any thoughts on this?

justinmayer:
Hi @seanh. Just checking in. Any thoughts on this pull request?

seanh:
@justinmayer Apologies, I'm just seeing this now. I hope to be able to take a look in the next few weeks

seanh (commented):
This looks fine to me, apologies for the slow response @joelsvensson .

This looks simple enough and looks like it won't make any difference to users that aren't using Stork, which is the default, so I don't see any reason not to merge this :+1:

It's a shame that for Stork users installing it is apparently slow. I don't use Stork and don't know anything about Rust/Cargo, so can't comment about that.

Rather than adding individual options for every possible dependency, I wonder if there's a generic solution that would allow users to install any dependency they want? For example perhaps users could provide a script that would be executed, so they could provide the `cargo install stork-search --locked` command themselves. Maybe there's a security issue with allowing callers to provide executable code? But I don't think so: we've nothing to secure here, the workflow runs in the caller's context.

justinmayer (approved):
Many thanks to @joelsvensson for the enhancement and to @seanh for reviewing 🏅

PR #3462, opened 2025-04-15
docs: Use raw string for the `PATH_METADATA` setting

Kristinita:
### 1. Summary

Currently, the Pelican documentation 2 times proposes this syntax:

```python
PATH_METADATA = "(?P<path_no_ext>.*)\..*"
```

It’s not a valid Python syntax. I fixed typos.

### 2. Style

I prefer raw strings to escaping backslashes due to the [**rule WPS342**](https://wemake-python-styleguide.readthedocs.io/en/latest/pages/usage/violations/consistency.html#wemake_python_styleguide.violations.consistency.ImplicitRawStringViolation) of wemake-python-styleguide:

> It is hard to read escape sequences inside regular strings, because they use `\\` double backslash for a single character escape.
>
> Use raw strings `r''` to rewrite the escape sequence with a `\` single backslash.

### 3. USE_FOLDER_AS_CATEGORY

I don’t understand why [**Pelican FAQ also recommends using `USE_FOLDER_AS_CATEGORY = False`**](https://docs.getpelican.com/en/latest/faq.html#how-do-i-make-my-output-folder-structure-identical-to-my-content-hierarchy) with the option `PATH_METADATA`.

If `USE_FOLDER_AS_CATEGORY = True`, Pelican still save the folder structure identical to the content hierarchy. `USE_FOLDER_AS_CATEGORY` is a stylistic (and in my opinion a useful) option. I propose removing `USE_FOLDER_AS_CATEGORY = False` from FAQ or explain in FAQ why this option is required.

Thanks.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -118,7 +118,7 @@ How do I make my output folder structure identical to my content hierarchy?
 Try these settings::
 
     USE_FOLDER_AS_CATEGORY = False
-    PATH_METADATA = "(?P<path_no_ext>.*)\..*"
+    PATH_METADATA = r"(?P<path_no_ext>.*)\..*"
     ARTICLE_URL = ARTICLE_SAVE_AS = PAGE_URL = PAGE_SAVE_AS = "{path_no_ext}.html"
 
 How do I assign custom templates on a per-page basis?
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -458,7 +458,7 @@ If you don't want that flexibility and instead prefer that your generated
 output paths mirror your source content's filesystem path hierarchy, try the
 following settings::
 
-    PATH_METADATA = '(?P<path_no_ext>.*)\..*'
+    PATH_METADATA = r'(?P<path_no_ext>.*)\..*'
     ARTICLE_URL = ARTICLE_SAVE_AS = PAGE_URL = PAGE_SAVE_AS = '{path_no_ext}.html'
 
 Otherwise, you can use a variety of file metadata attributes within URL-related
--- a/pelican/settings.py
+++ b/pelican/settings.py
@@ -144,7 +144,7 @@ DEFAULT_CONFIG = {
     "DEFAULT_ORPHANS": 0,
     "DEFAULT_METADATA": {},
     "FILENAME_METADATA": r"(?P<date>\d{4}-\d{2}-\d{2}).*",
-    "PATH_METADATA": "",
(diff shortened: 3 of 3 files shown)


justinmayer:
Hi @Kristinita. Thank you for the pull request.

Perhaps the following instance should also be changed for clarity?

[`"PATH_METADATA": "",`](https://github.com/getpelican/pelican/blob/9c7749a29887754d0ada49338503f0545870f8d2/pelican/settings.py#L147) 👉 `"PATH_METADATA": r"",`

Also, I renamed the title of this PR. Could you perhaps change the commit message to match it? That is:

> docs: Use raw string for the `PATH_METADATA` setting

Regarding the unrelated question in (3) above, I do not believe the current FAQ entry implies that `USE_FOLDER_AS_CATEGORY = False` setting is required — ergo the word “try”. I imagine the reason it is suggested is that the “bargain” with the user is being changed: normally the content folder structure has no bearing on the *output* folder structure, but the suggested changes are of course intended to instead mirror the content and output folder structures. Whether someone feels the `USE_FOLDER_AS_CATEGORY` setting is also appropriate in the latter scenario is indeed probably subjective, and PRs to clarify that are welcome.

Kristinita:
**Type: Replies** 💬

1. > Regarding the unrelated question in (3) above

	In my opinion my question is unrelated to the same extent as `USE_FOLDER_AS_CATEGORY = False` unrelated with the FAQ question “How do I make my output folder structure identical to my content hierarchy?”

1. > Perhaps the following instance should also be changed for clarity?

	**Done** :heavy_check_mark:
	
	Maybe another stylistic changes be a nice idea too. Examples of inconsistency in Pelican:
	
	1. The file `docs/settings.rst` uses single quotes, another files include `pelican/settings.py` use double quotes.
	1. Pelican uses the word `behavior` in [**13 files**](https://github.com/search?q=repo%3Agetpelican%2Fpelican%20behavior&type=code) and the word `behaviour` [**in 4 files**](https://github.com/search?q=repo%3Agetpelican%2Fpelican+behaviour&type=code).

1. > Also, I renamed the title of this PR. Could you perhaps change the commit message to match it?

	**Done** :heavy_check_mark:

Thanks.

justinmayer:
I agree that it would be better to address the two style inconsistencies you mentioned above 👍

justinmayer (approved):
Many thanks for this improvement, @Kristinita 👍

PR #3474, opened 2025-05-31
Change tuples to lists for SOCIAL and LINKS

clockback:
In the documentation, both `SOCIAL` and `LINKS` are described as a list of tuples, but all examples show them as tuples of tuples. It makes the most sense for them to be lists, so the examples should match the docs going forward.

Similarly, `LOCALE` is described as a list, but the examples show it as a tuple. Likewise, the examples should be lists, in line with the description.

# Pull Request Checklist

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -819,9 +819,9 @@ Time and Date
 
    .. parsed-literal::
 
-      LOCALE = ('usa', 'jpn',      # On Windows
+      LOCALE = ['usa', 'jpn',      # On Windows
                 'en_US', 'ja_JP'   # On Unix/Linux
-      )
+      ]
 
    For a list of available locales refer to `locales on Windows`_  or on
    Unix/Linux, use the ``locale -a`` command; see manpage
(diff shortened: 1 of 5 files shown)


clockback:
I notice that [LOCALE](https://docs.getpelican.com/en/latest/settings.html#LOCALE) is similar. It is described as a list, but the example is of a tuple. I can modify this PR or open a new one if you prefer.

avaris:
So... the title should be `tuples to lists`, right? Because the actual changes suggest that. On the other hand, you also say "It makes the most sense for them to be tuples". I'm a bit confused here.

clockback:
> So... the title should be `tuples to lists`, right? Because the actual changes suggest that. On the other hand, you also say "It makes the most sense for them to be tuples". I'm a bit confused here.

Thanks for pointing this out. Clumsiness on my part. I've corrected the title, description, and git commit.

avaris (approved):
Thanks, LGTM.

justinmayer (approved):
Thanks to @clockback for the fix and to @avaris for reviewing 👍

PR #3475, opened 2025-06-01
Enable `flake8-builtins` Ruff rule

clockback:
# Pull Request Checklist

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -56,7 +56,7 @@ def coverage(c):
 
 
 @task
-def format(c, check=False, diff=False):
+def formatcode(c, check=False, diff=False):
     """Run Ruff's auto-formatter, optionally with --check or --diff"""
     check_flag, diff_flag = "", ""
     if check:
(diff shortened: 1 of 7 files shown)


tasks.py
```diff
@@ -56,7 +56,7 @@ def coverage(c):
 
 
 @task
-def format(c, check=False, diff=False):
+def ruff_format(c, check=False, diff=False):
```
justinmayer, line 59: Because we could eventually use other formatters, perhaps we should either (1) add inline exceptions for the two `format` instances in this file, or (2) name this `formatcode` instead?

justinmayer (commented):
Nice work! I made one minor suggestion. Let me know your thoughts?

tasks.py
clockback: Personally, I prefer to avoid inline suppression where it is easy to do so, so I'll opt for (2).

justinmayer (approved):
Nicely done, Elliot. Many thanks for this enhancement! ✨

PR #3482, opened 2025-07-01
Limit Dependabot scope to GitHub Actions only

justinmayer:
Sometimes we intentionally pin versions for specific packages that can affect Pelican’s functional test output — such as Pygments, Jinja2, and Markdown — and thus automated Dependabot PRs for those packages are spurious and unwanted.

Meanwhile, [Dependabot will happily (and unhelpfully) generate a PR to update the Pygments version](https://github.com/getpelican/pelican/pull/3481), even though Pygments is [clearly and explicitly pinned in `pyproject.toml`](https://github.com/getpelican/pelican/blob/9e3e1325c0cdcabd7d0a99c78b7eadc170656f22/pyproject.toml#L37).

To resolve this, this pull request removes the portion of Dependabot’s configuration that generates PRs for Python package version updates, regulating Dependabot to only updating GitHub Actions version numbers in workflows.

------------------------

*Original commit that added the configuration that is removed herein:* https://github.com/getpelican/pelican/commit/8c7c29e4f2a35124f2bfe7905fd0a5a334f7e056

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,13 +1,8 @@
 # See https://docs.github.com/en/free-pro-team@latest/
 #  github/administering-a-repository/enabling-and-disabling-version-updates
-
+---
 version: 2
 updates:
-  - package-ecosystem: "pip"
-    directory: "/"
-    schedule:
-      interval: "monthly"
-    open-pull-requests-limit: 10
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:


boxydog:
> Sometimes we intentionally pin versions for specific packages that can affect Pelican’s functional test output — such as Pygments, Jinja2, and Markdown

I don't understand this. We pin it so the tests won't change, or we pin it so Pelican output won't change for users?

If it's just tests, I feel like ideally we would not pin things. Instead, if we updated a library, we'd update the tests and move on. But I'm naive.

avaris:
> If it's just tests, I feel like ideally we would not pin things. Instead, if we updated a library, we'd update the tests and move on. But I'm naive.

It's for tests. If not pinned, `pip` will install the latest version and some dependency could release a new version that tweaks their output (looking at you Pygments) causing `pelican` to generate a slightly different output, which triggers a failure in certain tests. All of a sudden, someone submitting/working on a PR will be greeted with a bunch of puzzling test failures that (most likely) have nothing to do with their changes.

boxydog:
I was not being clear at all, sorry.

Yes, we absolutely should pin things, for the reason you described. We have to know how to reproduce our own system.

But we shouldn't disable dependabot from updating the dependencies, should we? (I meant "pin" as in "never change" as in "tell dependabot not to change", which was completely confusing.) If dependabot updates our dependencies in a PR and the tests fail, then we should go fix the tests, no?

avaris:
Not if dependabot thinks this is a "reasonable" change: `<2.19.0 to <2.20.0`. Like, what is this? It's specifically set to not allow any `2.19.*` and dependabot comes with "hey, I see you don't want `2.19.*`, let me send a change to allow precisely that"

avaris:
Based on the way it is setup, I don't think anyone other than maintainers can amend the dependabot PR.

avaris:
If someone (besides a maintainer) wants to work on this to fix it, they can't add commits to that PR and make it pass. They would need to make a new PR anyway. So at best, it is a bug report in the form of a PR.

avaris (approved):

boxydog:
> If someone (besides a maintainer) wants to work on this to fix it, they can't add commits to that PR and make it pass. They would need to make a new PR anyway. So at best, it is a bug report in the form of a PR, which is less than ideal since it's not the first place people would look for issues to work on.

Again, I'm fine if you merge this.

However:

> at best, it is a bug report in the form of a PR

1. Sometimes you can just merge the PR without working on it, so it's not just a bug report.
2. We don't have anything else reporting on a regular basis that we're behind the times, so even an automated bug report has some value.

I predict that if we merge this, we'll be less likely to update dependencies. However, I'd be happy to be wrong.

justinmayer:
I appreciate the thoughtful discussion and input. Really helpful.

I think @avaris encapsulated very well my reservations about Dependabot’s behavior, which I ultimately find to be annoying rather than something that provides utility.

> I predict that if we merge this, we'll be less likely to update dependencies.

If we made a regular habit of adding upper-bound version pins to our dependencies, I might agree with this, but we intentionally avoid doing that whenever possible. The Pygments pin in question was only meant as a temporary, stop-gap measure, and I would say the system is working as intended, as thankfully @nkr0 was kind enough to update the Pygments dependency (in a way that Dependabot could not).

Pygments is currently the only dependency that is pinned, and we could presumably remove that pin if we had any confidence that future versions would be less likely to abruptly cause Pelican’s functional tests to fail.

All the other dependencies only have minimum versions specified, and there is really no reason to increment them unless we are aware of a problem that incrementing would solve.

Given that, I am not much concerned about dependency updates. If and when that actually becomes a problem, I would be more than happy to restore the relevant Dependabot configuration — or perhaps find some even better way to keep up with dependency updates.

Dreamsorcerer:
Just some feedback from my experience of managing dozens of repos, I think pinning everything in the requirements files and letting Dependabot update them is essential for good maintenance.

I have come back to repos many times after months of inactivity and found that if test dependencies are not pinned, then the CI is broken. You fix one issue, and there's another issue, and it takes hours sometimes to get the CI back to a workable state.

The requirements files are there only for the purpose of your test env (user's dependencies are from pyproject.toml), so I think it makes sense to ensure that testing is as reproducible as possible. Therefore, in all our repos today we pin to exact versions and allow Dependabot to update them one at a time. When you come back after months, the CI is probably still working, and you'll see from open Dependabot PRs if any new dependencies are breaking it, and they can be fixed one at a time more easily than trying to fix all the problems at once.

If a breakage is affecting users, they will probably create a PR to fix the issue and manually bump the version (they don't need to edit the Dependabot PR). At some point, Dependabot will recognise that the dependency has been updated and automatically close its PR. (Or they don't bump the version, and Dependabot eventually rebases its PR and it merges).

PR #3490, opened 2025-07-18
Implement dark mode for notmyidea template

TheJackiMonster:
I really liked to use the notmyidea template but it wasn't supporting a proper dark mode. So I added a few adjustments in CSS to make up for that. Also I removed loading a Github logo to get rid of requests to external content from third parties making my life easier for data security details about my page.

If anyone's interested in a preview for the changes, you can see them on my blog [here](https://thejackimonster.de/blog/).


justinmayer (commented):
Thank you for proposing these changes, Tobias. I looked at how your site renders in dark mode, and it is great to see the default theme finally blending in with the other UI elements in dark mode.

The CI process is reporting linter and test failures, so we will need to address those. I believe the whitespace changes are causing the former, so let's first revert those changes, and then we can move on to the test failures.

Have you followed the [guidelines for setting up the development environment](https://docs.getpelican.com/en/latest/contribute.html#setting-up-the-development-environment)? The `invoke setup` step should set everything up so you can run `invoke tests` and `invoke lint` to ensure everything passes before it even hits the GitHub Actions CI workflow. Does that make sense?

TheJackiMonster:
So from my testing with the development environment, the whitespace issue is solved. `invoke lint` works. But the tests fail since I have changed the output from generator by adjusting the template. So I assume I need to run `invoke update-functional-tests` updating those test cases but it seems like changing a lot more files than necessary because the newer generator adds different whitespace to the resulting HTML files.

justinmayer:
It appears your latest changes have addressed the linter issue, as those checks are now passing.

The functional tests are failing for a reason unrelated to your changes (i.e., the FeedGenerator 2.2.0 dependency upgrade), a problem that I believe I fixed in https://github.com/getpelican/pelican/commit/7ef5eae6399641b59987bbb408733fc9f6fe3207. If you rebase this branch onto latest `main`, I expect the tests will pass. Would you please try that and let me know?

justinmayer:
Indeed, you are correct. Would you be so kind as to run `invoke update-functional-tests`, ensure the diff looks correct to you, and commit and push the resulting changes in a separate commit?

justinmayer (approved):
Many thanks for these enhancements, Tobias. I try to avoid adding unintentional whitespace, even for something as trivial as the functional test output, so I went ahead and addressed that in a replacement commit, as well as taking the liberty of squashing the fix-up commits in this pull request to keep things tidy. Since I can’t force-push to your `main` branch, **this will take the form of a manual merge**, so don’t be alarmed when you see that this pull request has been closed — the relevant changes have already been merged and are in Pelican’s `main` branch.

If I may make a few suggestions for future contributions… _and I hope you will indeed continue to contribute!_ … they would be the following:

1. Run linters and tests locally, making sure pre-commit hooks are installed. (We talked about that already — just including it here for completeness.)
2. Always create a new branch for each contribution, as opposed to working directly on `main`. This makes it easier to manage and review changes. Example command: `git switch --create notmyidea-dark-mode`
3. Submit separate pull requests for each change, even when they seem small. In this case, separate pull requests for the dark mode and GitHub link changes would have made things considerably easier on my end 😸 

Thank you again for the excellent improvements to the default theme. Much appreciated! 🌟

TheJackiMonster:
Alright! Thanks for your feedback and thank you for maintaining Pelican. It's great software.

PR #3491, opened 2025-07-22
Enable `flake8-unused-arguments` Ruff rule.

clockback:
# Pull Request Checklist

- [ ] ~Updated **documentation** for changed code~

Some points:
* Where `del` has been run on an argument, I have reason to believe the function/method _could_ be called by keyword (due to polymorphism or duck typing), so I will not rename. Alternatively, I can change this to a Ruff suppression. Considerations:
  1. Using `del` avoids the need for suppressions, thus is a long-term solution.
  2. Using Ruff suppressions (e.g. `# noqa: ARG001`) is arguably uglier and may imply that the problem still needs solving. With that said, in the future, perhaps the rule will be more particular and raise fewer false positives, in which case `RUF100` will catch the suppression and allow us to remove it easily.
* `get_summary` - I'm not sure where this is actually used, but it doesn't seem to override anything, nor does it seem to be overriden by anything, so I figured (given there is some funky caching going on with the `memoize` decorator) I would prefix the argument with `_`.
* All unused lambda arguments now have `_` prefixes. Anyone calling keyword arguments for lambda functions is already doing something wrong.
* `receiver_test_function`, `get_posts`, and `makefile` are defined within a testing context, and keyword arguments are not used on them, so safe to rename their arguments.
* `fields2pelican` genuinely doesn't seem to use it's argument `filename_template`, and it doesn't seem like Pelican users (as far as I can tell) use it either, so I've removed it.
* `decorator` does not use its argument, but it is always passed in as a positional argument, so prefixing with `_`.
* `pelican_open` does not use `strip_crs` as far as I can tell, so I've removed it.

### DIFF
--- a/pelican/__init__.py
+++ b/pelican/__init__.py
@@ -252,6 +252,7 @@ class Pelican:
 
 class PrintSettings(argparse.Action):
     def __call__(self, parser, namespace, values, option_string):
+        del option_string  # Unused argument
         init_logging(name=__name__)
 
         try:
@@ -287,6 +288,7 @@ class PrintSettings(argparse.Action):
 
 class ParseOverrides(argparse.Action):
     def __call__(self, parser, namespace, values, option_string=None):
+        del parser, option_string  # Unused arguments
         overrides = {}
         for item in values:
             try:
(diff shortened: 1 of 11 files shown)


boxydog (approved):
I think your choices are reasonable: _ where you can, otherwise del with a comment.

A visual scan looks reasonable. Tests pass.

There is some risk that this could break something, but I would just blaze ahead. Velocity is better than fear.

PR #3502, opened 2025-10-04
Avoid `upload-pages-artifact` action in GitHub Pages workflow

seanh:
In upload-pages-artifact@v4 the maintainers have made the curious decision to omit hidden files (files whose names begin with ".") from the site that gets deployed to GitHub Pages and to provide no way to include a hidden file (at least "for now"), even though static sites may of course want to include files whose names begin with "." https://github.com/actions/upload-pages-artifact/pull/102

Work around this in by creating the "artifact" (tar file) ourselves and uploading it using the upload-artifact action, as suggested in the upload-pages-artifact PR linked above.

## Context

GitHub's upload-pages-artifact action is really just a lightweight wrapper for their upload-artifact action. The upload-pages-artifact action does two things:

1. It runs a `tar` command to create a tar file of your directory (in our case: Pelican's output directory)
2. It calls upload-artifact to upload the tar file to GitHub Pages

In upload-pages-artifact@v4 they've added a `--exclude` option to the `tar` command that causes any hidden files in Pelican's output dir to be omitted from the tar file and therefore ultimately omitted from the GitHub Pages site that gets deployed.

In this PR I've removed upload-pages-artifact from our workflow in favour of just running our own `tar` command without the new `--exclude` and then calling upload-artifact ourselves.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,9 +21,9 @@ jobs:
             python: "3.9"
 
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
       - name: Set up Python ${{ matrix.python }}
-        uses: actions/setup-python@v5
+        uses: actions/setup-python@v6
         with:
           python-version: ${{ matrix.python }}
           cache: "pip"
@@ -50,7 +50,7 @@ jobs:
     name: Lint
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
       - uses: pdm-project/setup-pdm@v4
         with:
           python-version: "3.11"
@@ -68,7 +68,7 @@ jobs:
     name: Test build
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
       - uses: pdm-project/setup-pdm@v4
         with:
           python-version: "3.11"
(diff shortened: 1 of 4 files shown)


.github/workflows/github_pages.yml
```diff
@@ -85,10 +85,28 @@ jobs:
...
+            --dereference \
+            --hard-dereference \
+            --directory "$OUTPUT_PATH" \
+            -cvf "$RUNNER_TEMP/artifact.tar" \
+            --exclude=.git \
+            --exclude=.github \
+            .
+          echo "::endgroup::"
+        env:
+          OUTPUT_PATH: ${{ inputs.output-path }}
```
seanh, line 102: This is exactly the same as the ["Archive artifact" step from upload-pages-artifact@v3.0.1](https://github.com/actions/upload-pages-artifact/blob/56afc609e74202658d3ffba0e8f6dda462b719fa/action.yml#L24-L38) other than `INPUT_PATH` being renamed to `OUTPUT_PATH` to match the name used elsewhere in our file.

In v4.0.0 upload-pages-artifact just added `--exclude=".[^/]*"` to [the `tar` command](https://github.com/actions/upload-pages-artifact/blob/7b1f4a764d45c48632c6b24a0339c27f5614fb0b/action.yml#L24-L39) to exclude hidden file.

justinmayer (approved):
Many thanks for enhancement and the detailed explanation! ✨

PR #3539, opened 2025-12-16
Add `HEADING_METADATA` feature for Markdown title extraction

mcepl:
This feature addresses the GitHub discussion issue https://github.com/getpelican/pelican/discussions/3290#discussioncomment-15270486

Previously, Markdown files without explicit 'Title:' metadata would fail to get titles extracted from their first heading, unlike RST files which support both metadata and heading-based title extraction.

**New Configuration Options:**
- HEADING_METADATA: Enable/disable heading metadata extraction
- HEADING_METADATA_MAP: Map heading levels to metadata fields
- HEADING_METADATA_PATTERNS: Custom regex patterns for extraction

**Key Features:**
- Extract metadata from markdown headings (# ## ### etc.)
- Support custom patterns (e.g., "### Author: Name")
- Metadata processors integration
- Regular metadata takes precedence over heading metadata
- Unicode and special character support
- Graceful error handling

**Implementation Details:**
- Added _extract_heading_metadata() method to MarkdownReader
- Modified read() method to extract headings before markdown conversion
- Backward compatible - no existing functionality changed
- Performance optimized with regex-based parsing

**Test Coverage:**
- Basic title/subtitle/summary extraction
- Unicode character support (Czech, accented)
- Metadata priority override behavior
- Custom pattern matching
- Edge cases and error handling
- Performance benchmarking (936KB/s throughput)
- Real-world scenario validation

Resolves issue where markdown files like:
    # My Article Title
    Date: 2023-12-01
    Category: tech

Can now have titles extracted without manual Title: metadata.

Files modified:
- pelican/settings.py: Add new configuration options
- pelican/readers.py: Implement heading extraction functionality

# Pull Request Checklist

Resolves: #issue-number-here 

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,4 @@
+Release type: minor
+
+Added HEADING_METADATA feature for extracting metadata from Markdown
+headings using configurable patterns and level mappings.
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -24,6 +24,7 @@ Release history
 4.11.0 - 2025-01-15
 ===================
 
+- Add HEADING_METADATA feature for Markdown title extraction, allowing metadata to be extracted from headings based on configurable patterns and level mappings
 - Add setting to selectively omit Typogrify filters `(#3439) <https://github.com/getpelican/pelican/pull/3439>`_
 - Add more blocks to the Simple theme’s base template, making it easier to create new themes by inheriting from the Simple theme `(#3405) <https://github.com/getpelican/pelican/pull/3405>`_
 - Fix auto-reload behavior upon changes to the theme, content or settings. Make default ``IGNORE_FILES`` recursively ignore all hidden files as well as the `default filters <https://watchfiles.helpmanual.io/api/filters/#watchfiles.DefaultFilter.ignore_dirs>`_ from ``watchfiles.DefaultFilter``. `(#3441) <https://github.com/getpelican/pelican/pull/3441>`_
--- a/docs/content.rst
+++ b/docs/content.rst
@@ -175,7 +175,38 @@ example, if you would like to extract both the date and the slug, you could set
 something like: ``'(?P<date>\d{4}-\d{2}-\d{2})_(?P<slug>.*)'``
 
 Please note that the metadata available inside your files takes precedence over
-the metadata extracted from the filename.
+metadata extracted from the filename.
+
+Metadata extraction from Markdown headings
+==========================================
+
+For Markdown files, you can also extract metadata from headings using the
+``HEADING_METADATA`` setting. This is particularly useful for articles where you
+want to use the first heading as the title without explicitly declaring it in
+the metadata section.
+
+When ``HEADING_METADATA`` is enabled, Pelican can extract metadata from headings
+using either level mappings or custom regex patterns. This allows Markdown files
+like::
+
+    Date: 2023-12-01
+    Category: tech
+
+    # My Article Title
+
+    Content goes here...
+
+This allows their titles to be extracted automatically without requiring
+explicit ``Title:`` metadata. Note that Markdown files must begin with the
+metadata block, and headings should follow it.
+
+The headings used for metadata extraction are automatically removed from the
+content to avoid duplication in the output.
+
+See the :ref:`HEADING_METADATA <settings/HEADING_METADATA>`,
+:ref:`HEADING_METADATA_MAP <settings/HEADING_METADATA_MAP>`, and
+:ref:`HEADING_METADATA_PATTERNS <settings/HEADING_METADATA_PATTERNS>` settings
+in the documentation for configuration options.
 
 Pages
 =====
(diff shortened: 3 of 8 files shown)


boxydog:
Below is Claude's review. I have looked it over, and I do tend to agree with it.

At a high level:
- It's a real deficiency that markdown doesn't have the feature rST has
- This PR got bigger than the rST functionality in several ways (more fancy, but also bundling unrelated changes)
- There's a bug: the summary pattern will always be matched before author or date, since they are iterated in order
- There aren't tests for the new functionality

I wonder if there is a smaller version of this PR we could start with? Or two: 1) bugfixes (pyproject.toml, Exception => ValueError, ..), 2) smaller heading feature?

Claude's detailed review is below.

================================

**Summary**: The PR adds a way to extract metadata (title, subtitle, etc.) from Markdown headings when explicit frontmatter metadata is absent. The motivation is valid — RST files already use document structure for title extraction, so Markdown should too. However, there are several issues.

---

### Bugs

**1. Default pattern ordering makes `author`/`date` patterns unreachable** (`settings.py:185–192`)

```python
HEADING_METADATA_PATTERNS = {
    'title': r'^#\s+(.+)$',
    'subtitle': r'^##\s+(.+)$',
    'summary': r'^###\s+(.+)$',      # ← matches any ### line
    'author': r'^###\s+Author[:\s]+(.+)$',   # ← NEVER reached
    'date': r'^###\s+Date[:\s]+(.+)$',       # ← NEVER reached
}
```

The generic `summary` pattern (`^###\s+(.+)$`) matches any `###` line, so `### Author: John` gets captured as `summary = "Author: John"` — never reaching the `author` pattern. More-specific patterns must come before generic ones.

**2. Dead "fallback" code in `_extract_heading_metadata`** (`readers.py:365–366`)

```python
if not heading_map and not heading_patterns:
    heading_patterns = {"title": r"^#\s+(.+)$"}
```

Since both `HEADING_METADATA_MAP` and `HEADING_METADATA_PATTERNS` are now set in `DEFAULT_CONFIG`, this branch is unreachable. The comment and code are misleading.

**3. Redundant/conflicting defaults** (`settings.py:179–192`)

Both `HEADING_METADATA_MAP` and `HEADING_METADATA_PATTERNS` define mappings for `#`/`##`/`###`. Since patterns are checked first (`custom_matched` flag), the entire level-mapping default is dead when both are set. Users will be confused about which one takes effect.

---

### Missing tests

The test changes in `test_readers.py` are only linter cleanups (renaming `content` → `_content`). There are **no new tests** for the actual feature: no tests for `HEADING_METADATA=True`, no tests for pattern matching, no tests for the precedence rule (regular metadata overrides heading metadata), no tests for line removal from content. This is a significant gap for a feature this complex.

---

### Design concerns

**Over-engineered for the stated use case.** The referenced discussion (#3290) asks for a simple thing: "use the first `#` heading as the title when no `Title:` frontmatter is present." The PR introduces 3 new settings, level maps, and custom regex patterns. Compare to RST behavior — RST just uses the document title, it doesn't extract `## subtitle` or `### summary` as metadata fields. The scope has grown considerably beyond RST parity.

A simpler design: `HEADING_METADATA = False` (boolean), and when `True`, extract only the first `#` heading as `title` if no `title` metadata is already present. This handles 99% of use cases without the complexity.

**Unrelated changes bundled in.** The PR also includes several good but unrelated fixes:
- `Exception` → `ValueError` throughout `settings.py`
- `pyproject.toml` typo fix (`*.py}"` → `*.py"`)
- Variable rename cleanups in tests
- `test_generators.py` assertRaises improvement

These should ideally be separate PRs so they can be merged without depending on the contested feature.

---

### Minor issues

- `docs/settings.rst`: The `HEADING_METADATA_PATTERNS` example ends with `"This allows Markdown files like: ..."` followed by the same example shown in the `HEADING_METADATA` section — awkward repetition and the sentence structure is broken ("This allows their titles..." doesn't follow the code block it's attached to).
- `RELEASE.md` — this file format may not be standard for this project; worth checking.
- `docs/changelog.rst` — entry is inserted at the top of the 4.11.0 section but the PR targets `main` (unreleased). Verify the section heading is correct for when this would ship.

---

### What's good

- The motivation is real and the general approach (pre-process text before `_md.convert()`, remove matched lines) is sound.
- `Exception` → `ValueError` changes are correct and welcome.
- The `pyproject.toml` glob typo fix is a genuine bug fix.
- Metadata precedence (regular beats heading) is implemented correctly: `heading_metadata.update(metadata); metadata = heading_metadata`.

ingwinlu (commented):
I always thought pelican has too many configuration options and it is overwhelming. I tend to NOT want to add to them by providing a workaround to a deficiency in the files we hare handling.

Could this not be done in a plugin instead for users who explicitly want / need it?

mcepl:
`pyproject.toml` fix is now https://github.com/getpelican/pelican/pull/3592

mcepl:
Exceptions cleanups are now https://github.com/getpelican/pelican/pull/3593.

PR #3544, opened 2025-12-24
fix: Don't log feed warning if `FEED_DOMAIN` is set

Dreamsorcerer:
`FEED_DOMAIN` appears to overrule `SITEURL`, so this warning should not be emitted if `FEED_DOMAIN` is set.


Dreamsorcerer:
Added a regression test. This one should be trivial to review.

PR #3547, opened 2025-12-26
Fix raising exceptions on filtered logs

### LINKED ISSUE #3546: Fatal option flag exits with no error or warning logged
In many cases (but not all) I am seeing fatal mode exit with nothing useful logged. The only output is:
```
[00:26:47] CRITICAL RuntimeError: Warning encountered            __init__.py:683
```
https://github.com/aio-libs/aio-libs.github.io/actions/runs/20513012381/job/58936364201?pr=1#step:5:12

I then need to remove `--fatal` to see the warning. However, in this example linked, there is no warning or anything logged when `--fatal` is removed (https://github.com/aio-libs/aio-libs.github.io/actions/runs/20512755690/job/58935737306#step:5:12), so I'm still completely lost as to what the issue is (but, I did have several cases earlier, where the warning did appear after removing `--fatal`).

Dreamsorcerer:
Fixes #3546.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu, macos, windows]
-        python: ["3.10", "3.11", "3.12", "3.13", "3.14"]
+        python: ["3.11", "3.12", "3.13", "3.14"]
 
     steps:
       - uses: actions/checkout@v6
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -9,7 +9,7 @@ version: 2
 build:
   os: ubuntu-22.04
   tools:
-    python: "3.10"
+    python: "3.11"
 
 # Build HTML & PDF formats
 formats:
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -2,12 +2,7 @@ import datetime
 import os
 import sys
 import time
-
-if sys.version_info >= (3, 11):
-    import tomllib
-else:
-    import tomli as tomllib
-
+import tomllib
 
 sys.path.append(os.path.abspath(os.pardir))
 
@@ -32,7 +27,7 @@ source_suffix = ".rst"
 master_doc = "index"
 project = project_data.get("name").upper()
 year = datetime.datetime.fromtimestamp(
-    int(os.environ.get("SOURCE_DATE_EPOCH", time.time())), datetime.timezone.utc
+    int(os.environ.get("SOURCE_DATE_EPOCH", time.time())), datetime.UTC
 ).year
 project_copyright = f"2010–{year}"  # noqa: RUF001
 exclude_patterns = ["_build"]
--- a/docs/locale/zh_CN/LC_MESSAGES/faq.po
+++ b/docs/locale/zh_CN/LC_MESSAGES/faq.po
@@ -7,8 +7,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PELICAN 4\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2024-11-07 16:25+0800\n"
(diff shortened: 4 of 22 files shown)


Dreamsorcerer:
@justinmayer Any chance we can get this one reviewed and merged? This is the biggest issue we're having currently, with the CI repeatedly passing and deploying despite critical errors being present.

pelican/log.py
```diff
@@ -83,44 +87,27 @@ def enable_filter(self):
...
+
+    def _log(self, level, msg, args, exc_info=None, extra=None, stack_info=False,
+             stacklevel=1):
+        try:
+            super()._log(level, msg, args, exc_info, extra, stack_info, stacklevel + 1)
+        except FilteredMessage:
+            # Avoid raising RuntimeError below if no log was emitted.
+            return
+
+        # __init__.py:main() catches this exception then does it's own critical log.
```
justinmayer: Typo: `it's` → `its`

pelican/log.py
```diff
@@ -83,44 +87,27 @@ def enable_filter(self):
...
+        # We need to avoid throwing the exception a second time here.
+        if level >= FatalLogger.fatal_lvl and level != logging.CRITICAL:
+            raise RuntimeError("Warning or error encountered")
```
justinmayer, line 119: Right now I only have a moment to skim this, so my apologies if what I am about to suggest is totally off-base…

What if we side-stepped the `FilteredMessage` exception class and avoided overriding the private `_log()` by instead overriding `handle()`, a public method and direct caller of `self.filter()`. Something like:

```python
class FatalLogger(LimitLogger):
      fatal_lvl = logging.CRITICAL + 1

      def handle(self, record):
          rv = self.filter(record)
          if rv:
              self.callHandlers(record)
              if record.levelno >= self.fatal_lvl and record.levelno != logging.CRITICAL:
                  raise RuntimeError("Warning or error encountered")
```

What do you think? Would something akin to this approach be feasible?

pelican/log.py
Dreamsorcerer:
```suggestion
        # __init__.py:main() catches this exception then does its own critical log.
```

pelican/log.py
Dreamsorcerer: My concern there is that we've replaced the entire implementation and no longer get upstream changes. A public method means that it's meant to be called by users, not that it's meant to be overriden by users. At the end of the day, the only way to deal with this is by messing with logger internals. The current solution is the least error prone as it just wraps the original implementation using super(). I think I also chose _log() rather than handle() as the latter may be used in other situations and suddenly raising a RuntimeError may break other libraries or something.

As already mentioned, the correct Pythonic solution would be to remove all these logging hacks and use the warnings module. If you'd be happy with that change, I can look at preparing a separate PR for that later. But, that'd be atleast a partial backward-compatibility break and should probably go in for v5.

pelican/log.py
```diff
@@ -83,44 +87,27 @@ def enable_filter(self):
...
-        super().error(*args, stacklevel=stacklevel, **kwargs)
-        if FatalLogger.errors_fatal:
-            raise RuntimeError("Error encountered")
+    fatal_lvl = logging.CRITICAL + 1  # i.e. No levels by default
+
+    def filter(self, record):
+        """A hack to let _log() know whether a message was logged or not."""
+        result = super().filter(record)
+        if not result:
+            raise FilteredMessage()
```
Dreamsorcerer:
```suggestion
            raise FilteredMessage
```

pelican/log.py
Dreamsorcerer: @justinmayer Added some regression tests.

Dreamsorcerer: @justinmayer As it looks like you're planning a v5, would you like me to submit another PR that replaces all these logs with warnings? Would greatly simplify the code and ease maintenance in future.

Issue #3536, opened 2025-12-13
Quickstart guide not working


ComputerCarsten:
- [x] I have read the [Filing Issues](https://docs.getpelican.com/en/latest/contribute.html#filing-issues) and subsequent “How to Get Help” sections of the documentation.
- [x] I can reproduce this problem with stock/default settings file, theme, and sample content (as described in above “How to Get Help” sections of the documentation).
- [x] I have searched the [issues](https://github.com/getpelican/pelican/issues?q=is%3Aissue) (including closed ones) and believe that this is not a duplicate.

- **OS version and name**: Linux Mint 22.1 ( related to Ubuntu )
- **Python version**: Python 3.12.3
- **Pelican version**: N/A
- **Link to theme**: None
- **Links to plugins**: None
- **Link to your site**: None
- **Link to your source**: https://getpelican.com/#quickstart
- **Link to a [Gist](https://gist.github.com/) with the contents of your settings file**: None

## Issue

Hi,

following the quick start tutorial the very first step fails:

`
$ python3 -m pip install "pelican[markdown]"
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.
    
    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.
    
    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.
    
    See /usr/share/doc/python3.12/README.venv for more information.

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
`
I know this is more a Python and distro problem and I have successfully worked around this some time ago. 

This is a non-trivial problem. And it will throw you and new users off at such an early point that they will not even bother trying fixing it. I suggest a portable installation, but that is a different issue.

virtadpt:
Perhaps adding bits about setting up a venv prior to this part of the instructions would be advisable.

justinmayer:
In #3549, I have made some installation documentation changes intended to address this issue. Could folks in this thread (and @getpelican/reviewers) take a moment to review the changes and provide some feedback?

PR #3549, opened 2025-12-29
docs: Add Pipx & `uv` as alternate install methods

justinmayer:
Pip and virtual environments can pose Pelican installation challenges for newcomers, so these changes include recommendations for Pipx and `uv` to facilitate installing Pelican.

Resolves: #3536

See also: #2560

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -1,62 +1,59 @@
 Installing Pelican
 ##################
 
-Pelican currently runs best on |min_python|; earlier versions of Python are not supported.
+Pelican currently runs best on Python |min_python|; earlier versions of Python are not supported.
 
 Once Pelican is installed, you can run ``pelican --help`` to see basic usage
 options. For more detail, refer to the :doc:`Publish<publish>` section.
 
-You can install Pelican via several different methods. The simplest is via
-`pip <http://www.pip-installer.org/>`_::
+You can install Pelican via several different methods.
 
-    python -m pip install pelican
+**Recommended method:** `Pip <https://pip.pypa.io/>`_ User Install
 
-Keep in mind that operating systems will often require you to prefix the above
-command with ``sudo`` in order to install Pelican system-wide. **You should
-not do this** as it may break your operating system. In this case you
-can add the ``--user`` flag or try one of the recommended methods below.
+To install Pelican via Pip::
 
-**Recommended method 1:** `pipx <https://github.com/pipxproject/pipx/>`_
+    python3 -m pip install --user "pelican[markdown]"
 
-pipx lets you execute binaries from Python packages in isolated environments.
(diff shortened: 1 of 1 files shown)


docs/install.rst
```diff
@@ -1,56 +1,85 @@
...
-Pelican currently runs best on |min_python|; earlier versions of Python are not supported.
+Pelican currently runs best on Python |min_python|; earlier versions of Python are not supported.
 
-You can install Pelican via several different methods. The simplest is via Pip_::
+Once Pelican is installed, you can run ``pelican --help`` to see basic usage
+options. For more detail, refer to the :doc:`Publish<publish>` section.
+
+You can install Pelican via several different methods.
+
+**Recommended method:** `Pip <https://pip.pypa.io/>`_ User Install
```
offbyone: I think it might be worth making `uv` the recommended method; it solves _so_ many of the possible permutations (python versions, dependency locks, etc).

If we do that, we can rewrite this with a whole bunch fewer variations:

- Basic install: 
    1. install uv
    2. `uv tool install pelican`

- Optional packages: 
    - `uv tool install 'pelican[markdown]'`

That gets rid of, like, all of this :D

minchinweb:
Thoughts:

- I'm not crazy about have 4 alternative installation methods listed side by side. Ideally, we would stick with one that would cover 95% of usecases.
    - partly this is because my plugins' documentation should support the officially recommended way to installing Pelican, and I'd rather not have to write it 4 different ways
- I think whatever method we pick needs to have a virtual environment, either explicitly or under the hood, as plugins (post v4.5) now get automatically loaded if they are installed.
    - I work on several Pelican sites on my same machine, all with slightly different plugin combinations, but I don't know how common that is
    - adding packages to a pipx install has always been a bit of a pain, but only because I can never remember how to properly add them (I think you can add them to the "spec", which is what you want, or can drop to pip directly)
    - with uv, it appears you can install additional packages into your tool install, but I've never tried it
    - personally, I'd probably use venv, but I've used it for so long I'm not sure if I'm missing how hard (or not) it is to set up for the first time.
- there is a philosophical question of whether to present Pelican a CLI script (and somewhat (or entirely) downplay the Python side) or to present Pelican at a Python library that you set up as a Python project. In the former, uv (or pipx) is a good presentation, but for the later, venv is probably more "proper".
- perhaps alternative installation methods could be moved to an "alternative install" methods page?

I don't feel hugely strongly about which method is presented as "blessed", just that it should be only one.

justinmayer:
Many thanks to everyone who provided feedback on my effort to expand the installation documentation. I feel comfortable enough with these changes to merge and see how they work in practice. If it seems like it’s too much, I’m happy to pare it down a bit in the future.

PR #3551, opened 2025-12-31
Feature: support for MyST markdown input

offbyone:
This adds support for the Markedly Structured Text format, which is a hybrid of Markdown and some of RST's better features.

https://myst-parser.readthedocs.io/

# Pull Request Checklist

Resolves: #issue-number-here 

### DIFF
--- a/MYST_README.myst
+++ b/MYST_README.myst
@@ -1,0 +1,204 @@
+---
+title: MyST Markdown Support in Pelican
+date: 2024-01-15
+---
+
+# MyST Markdown Support in Pelican
+
+## What is MyST?
+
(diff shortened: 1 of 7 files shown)


minchinweb:
Extensions like this are often done via Plugin rather than directly in Pelican core.

Also, there already is a MyST Reader plugin that might do what you're after --> https://pypi.org/project/pelican-myst-reader/

offbyone:
I'm thinking that it might be worthwhile to consider a more capable markdown parser. The extension is great, but I'm more interested in bringing it into core. I'm not attached to this; it's been a fun side experiment, and it's a draft for a reason.

PR #3555, opened 2026-01-04
Support `CSS_FILE` in Simple theme

### LINKED ISSUE #3554: CSS_FILE doesn't work with simple theme
In the docs, CSS_FILE is listed as a general setting that applies to themes. Yet, the setting has no impact on the simple theme.

Dreamsorcerer:
This allows CSS_FILE to work as documented when using the simple theme (or more likely inheriting from it). This also simplifies development of a custom theme, as we now only need to create the CSS file to get a working theme. If we're happy with this, I'll also tweak the theming docs to show that it is no longer necessary to manually add this to the theme.

Fixes #3554.


Dreamsorcerer:
Updated to make the link output conditional on the file existing. Regression tests added too.

justinmayer:
Thank you, Sam. Would you mind resolving the merge conflict and adding the appropriate changes to the docs that you previously suggested?

Dreamsorcerer:
Yeah, forgot about that. Docs updated now.
Do you have someone to update the Chinese translations? If not, I guess it'll just use the English ones for now until someone comes along to update it.

Note I deleted the .zip file as it seems pointless for a single file, plus binary files won't compress well in git and are a known exploit target (e.g. xz-utils).

PR #3583, opened 2026-04-05
fix: Move `livereload` import inside `docserve()` task

cr7yash:
This fixes 'invoke setup' failing in fresh virtual environments. The top-level import of livereload caused tasks.py to fail on import before any task could execute, since livereload is a dev dependency only installed by pdm during the setup process.

# Pull Request Checklist

Resolves: #issue-number-here 


minchinweb:
Ran into this the other day myself; glad to see the fix!

PR #3584, opened 2026-04-06
Doc update 2026

minchinweb:
A collection of updates to the docs, that I hope are helpful:

- document another (plugin) signal
- more information on how to extra metadata from filenames (using regexes)
- note that Plugins and Themes can be found on PyPI as well
- updated code formatting, to bring it more inline with current black/ruff formatting, particularly with the use of double (over single) quotes


justinmayer (approved):
These documentation enhancements are very helpful. Thank you, @MinchinWeb! 💫

PR #3589, opened 2026-04-13
Fix disabled category output

### LINKED ISSUE #3571: Can't disable categories
There seems to be no way to properly disable categories.

We have set CATEGORY_URL/CATEGORY_SAVE_AS/CATEGORIES_SAVE_AS all to None.

In the simple theme it outputs the category on each article with an `if article.category` check. But, there seems to be no way to make that check return false to avoid the output. The category link is obviously broken and leads to a 404.

If I disable USE_FOLDER_AS_CATEGORY, then it just changes the category to 'misc'.
If I also change DEFAULT_CATEGORY to None, then I get an AttributeError, or if I change it to an empty string then I get `could not find information about 'category'` and no articles get rendered at all.

Dreamsorcerer:
Fixes #3571.

Categories should not be output with broken links if CATEGORY_SAVE_AS is not set.

### DIFF
--- a/pelican/contents.py
+++ b/pelican/contents.py
@@ -577,7 +577,7 @@ class Page(Content):
 
 
 class Article(Content):
-    mandatory_properties = ("title", "date", "category")
+    mandatory_properties = ("title", "date")
     allowed_statuses = ("published", "hidden", "draft", "skip")
     default_status = "published"
     default_template = "article"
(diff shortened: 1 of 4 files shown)


justinmayer:
@Dreamsorcerer: It seems merging one of your other PRs has resulted in a merge conflict in this one. Would you be so kind as to resolve it?

justinmayer (approved):
Thank you for this enhancement, Sam 👍

PR #3593, opened 2026-04-17
refactor: Use specific exceptions instead of bare `Exception`

mcepl:
There are numerous instances in Pelican using bare Exception even in situations where something more specific exception could (and thus should) be used. This PR tries to eliminate the most egregious ones.


justinmayer (approved):
Thank you for these improvements, Matěj 👍

PR #3594, opened 2026-04-19
fix: Ensure MIME type for JS is JavaScript

### LINKED ISSUE #3588: Wrong MIME type for JavaScript on development server
- [x] I have read the [Filing Issues](https://docs.getpelican.com/en/latest/contribute.html#filing-issues) and subsequent “How to Get Help” sections of the documentation.
- [x] I can reproduce this problem with stock/default settings file, theme, and sample content (as described in above “How to Get Help” sections of the documentation).

- **OS version and name**: Windows 11 Home (25H2) x86_64
- **Python version**: 3.14.4
- **Pelican version**: 4.11.0
- **Link to theme**: N/A
- **Links to plugins**: N/A
- **Link to your site**: N/A
- **Link to your source**: N/A
- **Link to a [Gist](https://gist.github.com/) with the contents of your settings file**: N/A

## Issue

On the development server in Pelican, JavaScript files are incorrectly served as MIME type "text/plain" not "text/javascript".

This was causing problems for me since my browser would not work predictably with JavaScript files served that way.

After poking around the code in Pelican, then Python's http.server module, then Python's mimetypes module, I think I figured out the problem.

I suspect this is a bug/quirk in the mimetypes module when used on Windows (the OS I'm currently using).

For some reason, the .js extention is mapped to "text/plain" in that module. 

I was able to fix this problem on my end by making my own thrown together local server and then adding SimpleHTTPRequestHandler.extensions_map = {".js": "text/javascript"}. (FYI: I also had to reset my browser's cache for it to finally work. Apparently it had cached the originally served files with the wrong MIME type, which was causing me frustration until I thought to flush the cache.)

It looks like this can also be worked around in a similar fashion in Pelican's server.py file by adding ".js": "text/javascript" to the extensions_map dictionary defined in the ComplexHTTPRequestHandler class.

I'm not sure if this is something that you want to fix in Pelican since it is mostly a Python library problem and (I suspect) a Windows only problem. But working around it may help other users of Pelican that end up having the same problems I had. (Also, I'm a hobbyist/amateur coder, so it's possible I may have misinterpreted some of my findings.)

justinmayer:
On systems such as Windows, JavaScript files can be incorrectly assigned the "text/plain" MIME type in Python’s `mimetypes` module. This overrides this behavior when it occurs.

Fixes: #3588


robertdmendoza:
Unfortunately, this does not work for me. The main index.html page is being served as "text/plain" (the /favicon.ico file is also served as "text/plain"). The browser just shows the raw html and loads nothing further.

Looking at the changes, it appears that line 123 of server.py is causing the problem.

justinmayer:
Ugh. That line was merely some debugging code I had temporarily added and forgot to remove. Can you try again?

robertdmendoza:
Everything is working on my end. The JavaScript file MIME type is correct. Thank you for looking into this.

justinmayer:
Thanks for testing it and being so responsive. I’ll include this in tomorrow’s release

PR #3606, opened 2026-06-11
Enable flake8-bandit (Ruff "S") lint rules

rivassec:
## Summary

This PR resolves the `# TODO: "S"` marker in `pyproject.toml` by enabling the `flake8-bandit` security rule set with a clean, low-noise suppression strategy.

* **Test file ignores:** Added a global `per-file-ignores` rule for `pelican/tests/*`. This cleanly suppresses rules that don't make sense for test suites, like `S101` (asserts), `S105`/`S106` (mock credentials), `S108` (`/tmp` paths), and `S311` (non-crypto random numbers).
* **Library inline annotations:** Discovered 15 true findings in the source code. Each has been given a targeted `# noqa` comment with a clear context justification.
* **Deferred changes:** Three findings highlight actual structural issues that require follow-up work. I chose to temporarily suppress these with specific comments so we can keep this PR strictly focused on enabling the linter without introducing stealthy behavior changes.

There are no logic or runtime behavior changes in this commit.

## Why

This check has been sitting on the maintainer wishlist for a while but was blocked by the backlog of existing violations. By unblocking it now, we ensure every future contribution gets automatic static security analysis. The actual code issues we found are now cleanly surfaced and tracked rather than being buried in generic technical debt.

## Test plan

* Run `pdm lint --diff` to confirm the formatting is green.
* Run `pdm run pytest pelican/tests -q` to verify all 312 tests still pass.
* Run `ruff check --select S` to ensure the new ruleset runs cleanly.

## Follow-up work

These items were intentionally left out of this PR to avoid breaking changes, but they should be tackled next:

* Replace the `assert` statements used for control flow in `pelican_import.py` (which get completely stripped out if Python runs with optimizations).
* Migrate `pelican/server.py` away from the deprecated `ssl.wrap_socket` toward `ssl.SSLContext`.
* Audit the `shell=True` subprocess call used during the WordPress pandoc import step.

### DIFF
--- a/pelican/server.py
+++ b/pelican/server.py
@@ -153,7 +153,10 @@ if __name__ == "__main__":
             args.path, (args.server, args.port), ComplexHTTPRequestHandler
         )
         if args.ssl:
-            httpd.socket = ssl.wrap_socket(
+            # ssl.wrap_socket is deprecated since Python 3.12 in favor of
+            # ssl.SSLContext.wrap_socket, which would also let us pass a
+            # tightened protocol.
+            httpd.socket = ssl.wrap_socket(  # noqa: S504
                 httpd.socket, keyfile=args.key, certfile=args.cert, server_side=True
             )
     except ssl.SSLError as e:
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,16 @@
+Release type: patch
+
+Enable the `S` (flake8-bandit) Ruff rule set, addressing the `# TODO: "S"`
+note in `pyproject.toml`. Adds a `pelican/tests/*` ignore for rules that
+are universally appropriate to suppress in tests (asserts, hardcoded test
+values, subprocess fixtures, loopback bindings) and adds inline `# noqa`
+suppressions, each with a justification comment, on the existing library
+findings.
+
+No behavior changes; this commit only enables the lint and documents why
+each existing finding is acceptable. Two `S101` and one `S504`
+suppressions in `pelican/tools/pelican_import.py` and `pelican/server.py`
+flag latent issues that warrant follow-up (asserts used as control flow
+or invariant checks are stripped under `python -O`; `ssl.wrap_socket` is
+deprecated since Python 3.12) and are intentionally deferred to keep
+this commit purely lint-enabling.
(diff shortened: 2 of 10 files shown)


rivassec:
FWIW the main role of this PR is to start rolling out bandid

rivassec:
bump?

rivassec:
Gentle bump - it has been about a month. CI is green and this is a small, self-contained change (enables the flake8-bandit / Ruff `S` security lint rules). Happy to rebase onto current main or split it into smaller chunks if that would make review easier. Thanks!

pelican/server.py
```diff
@@ -153,7 +153,11 @@ def __init__(self, base_path, *args, **kwargs):
             args.path, (args.server, args.port), ComplexHTTPRequestHandler
         )
         if args.ssl:
-            httpd.socket = ssl.wrap_socket(
+            # ssl.wrap_socket is deprecated since Python 3.12 in favor of
+            # ssl.SSLContext.wrap_socket, which would also let us pass a
+            # tightened protocol. Fixing this without breaking the local-dev
```
boxydog: I would delete "Fixing this without breaking the local-dev workflow is tracked separately" .. making promises about things far from the code rots.

boxydog (approved):
works for me.

boxydog:
Thanks for your work. It can be an uphill climb to get an open source project to accept a PR, but this seems like a step forward to me.

rivassec:
I understand, i assume there has been a flood of PRs and issues this year

pelican/server.py
rivassec: Good call - removed. The comment now just states the deprecation fact, no promises about external tracking.

rivassec:
Thanks for the review @boxydog. Addressed your inline comment (dropped the "tracked separately" sentence) and re-pushed - the amend also gives the PR a fresh CI run, since the June checks had aged out of the UI. Do you have merge rights here, or is @justinmayer the right person to ping when he is next around?

justinmayer:
I triggered another CI run and will review as soon as I can.

boxydog:
I do not have merge access. Justin's the one. Best wishes.

rivassec:
Hi @justinmayer, checking in on this one. CI's green and @boxydog approved it a couple weeks back, so it should be good to go whenever you get a chance. No rush at all, and happy to rebase or make changes if you'd like anything adjusted.

PR #3611, opened 2026-07-09
Fix output path boundary validation

ShiroKSH:
﻿## What changed

- Replace prefix-based output-path validation with normalized common-path checks.
- Resolve symlinks before validating the boundary.
- Keep the existing output path format and error behavior.
- Add a regression test for sibling directories sharing the output prefix.
- Add the required patch release metadata.

## Why

A path such as ../output-evil/site.html could pass the old startswith check and be written outside the configured OUTPUT_PATH.

## Validation

- 270 tests passed, 28 skipped in the expanded suite.
- Targeted sanitised_join tests: 4 passed.
- Ruff check and format passed.
- git diff --check passed.

The unfiltered Windows suite has seven pre-existing failures because a tracked symlink fixture is checked out as a text file without symlink privileges.


ShiroKSH:
@justinmayer - could you take a look at the output path boundary fix in this PR when you have a chance, or route it to the right reviewer? I can make any requested changes or add tests. Thanks!

ShiroKSH:
Hi maintainers, this PR has been open for over two weeks. When you have a moment, I would appreciate a review. Thank you.

justinmayer:
Summer has been busy so far, but rest assured, I will turn my attention to this request as soon as I can.

PR #3612, opened 2026-07-20
Clarify documentation for *S_SAVE_AS vs *_SAVE_AS

andymckay:
When generating a simple static site, I wanted to not have the listings page, such as `authors.html` not be generated. I read through the docs and noted that `AUTHOR_SAVE_AS` would not generate an Authors page, but I couldn't stop `authors.html` from being generated. I didn't spot the difference between `AUTHOR_SAVE_AS=""` and `AUTHORS_SAVE_AS=""`.

`AUTHORS_SAVE_AS=""` is needed to stop the `authors.html` page from being generated. I couldn't spot a matching note in the docs about this, so this added in note hopefully clarifies it a bit.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+Release type: patch
+
+- Minor documentation improvement to clarify how to prevent listing pages from being generated.
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -687,6 +687,10 @@ template.
 
    The location to save the list of all articles. The default is ``"index.html"``.
 
+Note
+
+If you do not want one or more of the list pages to be created (e.g., you are the only author on your site and thus do not need a listing of Authors), set the corresponding *_SAVE_AS setting to "" to prevent the relevant page from being generated.
+
 URLs for direct template pages are theme-dependent. Some themes use
 corresponding ``*_URL`` setting as string, while others hard-code them:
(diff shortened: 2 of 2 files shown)


minchinweb:
I'm not sure I understand from your note what you are trying to clarify here. Personally, I like code samples I can copy; consider adding/replacing/extending your notes with some of what you're written here, perhaps like:

> If you want to do <generate a single author page, but not the all author listing page>, set your settings thus:
> 
> ```python
> # pelicanconf.py
>
> AUTHORS_SAVE_AS = ""
> AUTHOR_SAVE_AS = "authors.html"
> ```
>
> This assumes you have a single author, and will break (because Pelican will try and overwrite the page) if you have multiple authors.

Adjust to taste and fix my code if I got it wrong...

andymckay:
I was trying to duplicate the note underneath the `*_SAVE_AS` settings. Which already exists in the docs, you can see that here 👇🏾 

https://pelican--3612.org.readthedocs.build/en/3612/settings.html

Screenshot 2026-07-26 at 12 36 27 PM

So this new note is kinda similar 👇🏾 

Screenshot 2026-07-26 at 12 45 55 PM

I'm not sure a note that combines multiple settings as an example makes sense. I'm a little confused about would break though? I would have to try and test what you might be talking about if you have those settings.

